PR #6, opened 2023-11-14
Ark encoding

altafan:
This fixes the existing method for secret/public key encoding and adds support for:
* encoding/decoding ark addresses
* encoding/decoding ark relays
* encoding/decoding ark url

This also fixes the name of the submodule, called now `github.com/ark-network/ark/common`.

This also adds a replacement in the main go.mod for the import of the submodule for future tasks.

Last but not least, this adds (few) tests for all the encoding methods.

Please @louisinger  review.

### DIFF
--- a/pkg/common/encoding.go
+++ b/pkg/common/encoding.go
@@ -15,7 +15,7 @@ const (
 	RelaySep = "-"
 )
 
-func EncodeSecKey(hrp string, key *secp256k1.PrivateKey) (string, error) {
+func EncodeSecKey(hrp string, key *secp256k1.PrivateKey) (seckey string, err error) {
 	if key == nil {
 		return "", fmt.Errorf("missing secret key")
 	}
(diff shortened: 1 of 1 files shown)


pkg/common/encoding.go
```diff
@@ -2,44 +2,167 @@ package common
...
-	return secp256k1.ParsePubKey(buf)
+func EncodeAddress(key *secp256k1.PublicKey) (string, error) {
+	if key == nil {
+		return "", fmt.Errorf("missing public key")
+	}
+	grp, err := bech32.ConvertBits(key.SerializeCompressed(), 8, 5, true)
+	if err != nil {
+		return "", err
+	}
+	return bech32.Encode(AddressPrefix, grp)
```
louisinger: The HRP could be an argument of that function instead ? it's common to have different prefix for testnet networks for instance

pkg/common/encoding.go
```diff
@@ -2,44 +2,167 @@ package common
...
 	"github.com/btcsuite/btcd/btcutil/bech32"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
 
 const (
-	PubKeyPrefix = "arkpub"
-	SecKeyPrefix = "arksec"
+	PubKeyPrefix  = "arkpub"
+	SecKeyPrefix  = "arksec"
+	AddressPrefix = "arkaddr"
```
tiero: Why not just `ark` without `addr`?

pkg/common/encoding.go
```diff
@@ -2,44 +2,167 @@ package common
...
+	"strings"
 
 	"github.com/btcsuite/btcd/btcutil/bech32"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
 
 const (
-	PubKeyPrefix = "arkpub"
-	SecKeyPrefix = "arksec"
+	PubKeyPrefix  = "arkpub"
```
tiero: What about just truncating to `a` instead of full `ark` for pub and sec? ie. `apub` and `asec`

altafan: Done.

pkg/common/encoding.go
altafan: done

pkg/common/encoding.go
altafan: Right! Shall we add some kind of net params struct in this PR?

louisinger: I think a `string` as argument is fine (at least now for the PoC)

pkg/common/encoding.go
```diff
@@ -2,44 +2,179 @@ package common
...
+	combinedKey := append(aspKey.SerializeCompressed(), userKey.SerializeCompressed()...)
+	grp, err := bech32.ConvertBits(combinedKey, 8, 5, true)
+	if err != nil {
+		return "", err
+	}
+	return bech32.EncodeM(AddressPrefix, grp)
+}
 
-	return secp256k1.ParsePubKey(buf)
+func DecodeAddress(addr string) (*secp256k1.PublicKey, *secp256k1.PublicKey, error) {
```
louisinger: nit: what do u think of having a struct as return type instead of a tuple `key, key, err` ? Saying this because it's hard to discriminate user/asp key looking at the signature as it is

altafan: I'll add a name to the return values

altafan: done.

PR #9, opened 2023-11-17
Ocean connection support

### LINKED ISSUE #8: Add connection to ocean wallet
Let's use [ocean](https://github.com/vulpemventures/ocean) as the underlying wallet for the asp. Requires a config entry to specify the connection URL and and adapter to call its APIs.

altafan:
This adds the basics to support connection to an ocean wallet. Since we don't really need multiple accounts, I decided to hide the account scope in the ports interfaces. The adapter makes use of an hardcoded `ark` account whenever required.

This closes #8.

Please @louisinger review.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/ark-network/ark/common v0.0.0-00010101000000-000000000000
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/viper v1.17.0
-	github.com/vulpemventures/go-elements v0.4.6
+	github.com/vulpemventures/go-elements v0.4.7
 	google.golang.org/grpc v1.59.0
 	google.golang.org/protobuf v1.31.0
 )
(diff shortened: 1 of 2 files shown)


pkg/common/appdata.go
```diff
@@ -0,0 +1,106 @@
...
+	"strings"
+	"unicode"
+)
+
+// appDataDir returns an operating system specific directory to be used for
+// storing application data for an application.  See AppDataDir for more
+// details.  This unexported version takes an operating system argument
+// primarily to enable the testing package to properly test the function by
+// forcing an operating system that is not the currently one.
+func appDataDir(goos, appName string, roaming bool) string {
```
louisinger, line 21: any reason to not use `btcutil.AppDataDir` ? We are used to do so in tdex-daemon isn't it ?

altafan: didn't want to have btcutil as direct dep of the project only for this function

PR #11, opened 2023-11-20
Add domain and app layers

altafan:
This adds the core layers of the PoC:
* **domain** layer- `round` aggregates all info about a round like the current stage, the registered payments, the pool txid, etc. 
* **application** layer - basic service that schedules 3 tasks: start round (start registration stage), end registration/start finalization, end finalization (broadcast pool tx to the mempool)
* **persistency** layer - the repo manager handles an event store and a projection store. The first is used to persist events ONLY, that they can be replayed anytime. Only internal components interact with this store. The second is used to persist projections whenever events occur, ie. the status of an aggregate after a list of events took place. This is useful, internally, to get the current round without replaying events, or to implement public APIs consumed by other service interacting with an ASP. The idea is that the implementation of the repo manager takes care of keeping the projection store up-to-date once events are persisted in order to not have this logic as part of the app layer.

No implementation of the interfaces added to `ports` yet.

Please @louisinger review.


internal/core/domain/events.go
```diff
@@ -0,0 +1,50 @@
...
+func (r PaymentRegistrationStarted) isEvent() {}
+func (r PaymentRegistrationEnded) isEvent()   {}
+func (r PaymentFinalizationStarted) isEvent() {}
+func (r PaymentFinalizationEnded) isEvent()   {}
+func (r RoundFailed) isEvent()                {}
+func (r InputsRegistered) isEvent()           {}
+func (r OutputsRegistered) isEvent()          {}
+
+type PaymentRegistrationStarted struct {
+	Id        string
```
louisinger, line 15: I would wrap `ID` member into a `BaseEvent` struct. Could be useful to gather "common members" at the same place no?

internal/core/domain/payment.go
```diff
@@ -0,0 +1,25 @@
...
+func (p Payment) TotOutputAmount() uint64 {
+	tot := uint64(0)
+	for _, r := range p.Receivers {
+		tot += r.Amount
+	}
+	return tot
+}
+
+type Vtxo struct {
+	Txid string
```
louisinger, line 18:
```suggestion
	TxId string
```

internal/core/domain/events.go
```diff
@@ -0,0 +1,50 @@
...
+	Timestamp int64
+}
+
+type PaymentRegistrationEnded struct {
```
louisinger: Is there only 1 `PaymentRegistrationEnded` event per round ? or each `PaymentRegistrationStarted` should have a "end" event ? The name suggest the second option 

--- 
to make the congestion tree verifiable, the event should include pool transaction (unsigned) ?

internal/core/domain/events.go
```diff
@@ -0,0 +1,50 @@
...
+	Id             string
+	ForfeitTxs     []string
+	CongestionTree []string
+}
+
+type PaymentFinalizationStarted struct {
```
louisinger: From timeline point of view, it looks similar to `PaymentRegistrationEnded`. Do we really need that event ?

internal/core/domain/events.go
```diff
@@ -0,0 +1,50 @@
...
+	Id string
+}
+
+type PaymentFinalizationEnded struct {
```
louisinger: What if we rename that `RoundDone` ?

internal/core/domain/payment.go
```diff
@@ -0,0 +1,25 @@
...
+	VOut uint32
+}
+
+type Receiver struct {
+	Pubkey string
```
louisinger, line 23:
```suggestion
	PubKey string
```

internal/core/domain/round_repo.go
```diff
@@ -0,0 +1,18 @@
+package domain
+
+import "context"
+
+type RoundEventRepository interface {
+	Save(ctx context.Context, events ...RoundEvent) error
+	Load(ctx context.Context, id string) (*Round, error)
+}
+
+type RoundRepository interface {
```
louisinger, line 10:
```suggestion
type RoundProjectionRepository interface {
```

internal/core/domain/round_repo.go
```diff
@@ -0,0 +1,18 @@
+package domain
+
+import "context"
+
+type RoundEventRepository interface {
+	Save(ctx context.Context, events ...RoundEvent) error
+	Load(ctx context.Context, id string) (*Round, error)
```
louisinger, line 7: Should return a `RoundEvent` instead of a `Round` ?

internal/core/domain/round_repo.go
```diff
@@ -0,0 +1,18 @@
+package domain
+
+import "context"
+
+type RoundEventRepository interface {
```
louisinger, line 5: does it make sense to add a `LoadAllEvents` method in case we need to recover the projections DB from the events one ? and maybe a `RestoreFromEvents` in the `RoundRepository` ?

internal/core/domain/round.go
```diff
@@ -0,0 +1,247 @@
...
+	Id                string
+	StartingTimestamp int64
+	EndingTimestamp   int64
+	Stage             Stage
+	Payments          map[string]Payment
+	Txid              string
+	ForfeitTxs        []string
+	CongestionTree    []string
+	Version           uint
+	Changes           []RoundEvent
```
louisinger, line 49:
```suggestion
	Events           []RoundEvent
```

internal/core/domain/round.go
```diff
@@ -0,0 +1,247 @@
...
+
+func NewRound() *Round {
+	return &Round{
+		Id:       uuid.New().String(),
+		Payments: make(map[string]Payment),
+		Changes:  make([]RoundEvent, 0),
+	}
+}
+
+func NewRoundFromEvents(events []RoundEvent) *Round {
```
louisinger, line 60: To make this possible, I think one of the `RoundEvent` must contain the Ark service providers coins selected to make the pool transaction (or the tx itself). Otherwise, this function cannot be deterministic

internal/core/domain/events.go
altafan: I don't think this app will have other domains. I thought about creating the `BaseEvent` struct but in the end it would just increase the complexity.

internal/core/domain/events.go
altafan: The round is divided into 2 phases: registration and finalization.

During the registration, users can send their vtxos to spend, and the receivers for the new vtxos that are going to be created.
When a user registers his "inputs" (vtxos to spend), he receives an id (mocks blinded credential), that he can use to register the "outputs" (<pubkey, amount> of the vtxos to be created).
At the end of this stage, the asp publishes the congestion tree and the forfeit txs (w/ connectors) for the senders.

The second and final stage is the finalization where users send back to the asp the signed forfeit txs. At the end of this stage, the asp publishes the pool tx.

So yes, there's only 1 `PaymentRegistrationStarted` and 1 `PaymentRegistrationEnded` per round. There can be multiple `InputsRegistered` and `OutputsRegistered` events per round instead. The first happens when user registers his "inputs", while the second when a user registers his "outputs".

We may want to find better names for these events, 'cause I guess you got confused by their names.

altafan: > to make the congestion tree verifiable, the event should include pool transaction (unsigned) ?

Not sure.. what means "verifying the congestion tree" to the ASP for the moment? For example, It could mean checking only that the input of the root tx of tree is exactly the shared output, for this the txid is enough.

internal/core/domain/events.go
altafan: what if we rename the events: `RoundStarted`, `RoundFinalizationStarted`, `RoundFinalized`? The second event would be the actual `PaymentRegistrationEnded` + `PaymentFinalizationStarted`

internal/core/domain/round_repo.go
altafan: Nope, when you "load" events, you basically want to replay all the events for an aggregate (domain) to know its status at some point in time.

internal/core/domain/round_repo.go
altafan: Nope it doesn't make any sense.

internal/core/domain/round.go
altafan: When a round is completed, the outcome is a pool tx. The minimal set of data to record this event is the `txid` of the pool tx. You can then use an explorer to retrieve the hex of the pool tx and know the coins selected as inputs.

What's so interesting in the inputs of the pool tx to make you believe it's valuable recording them?

louisinger: > When a round is completed, the outcome is a pool tx. The minimal set of data to record this event is the `txid` of the pool tx. You can then use an explorer to retrieve the hex of the pool tx and know the coins selected as inputs.

It works only if the round is done and the pool tx has been broadcasted.

> What's so interesting in the inputs of the pool tx to make you believe it's valuable recording them?

I'm worrying about the recovery of the **unsigned** pool transaction. In case the round failed before signing step, it seems to me that we cannot recover the pool tx isn't it ?

internal/core/domain/events.go
louisinger: Maybe you are right. I was thinking if later Events structs have other common parameters. It's up to you not important anyway now

internal/core/domain/events.go
louisinger: The txid is not enough in my opinion (at least until the tx isn't broadcasted), I see other possible checks:
- is the shared output script correct ?
- is the pool tx funded correctly (with ASP inputs)
- is change output(s) correct ? 
- later, with collaborative exits, pool tx could also include outputs going to "main-chain" ?

internal/core/domain/events.go
louisinger: Nit: `RoundTransactionCreated` for the second event ?

internal/core/domain/round.go
altafan: > It works only if the round is done and the pool tx has been broadcasted.

Yes, because the round is finalized only if the pool tx is broadcasted, otherwise it fails with some error message

> I'm worrying about the recovery of the unsigned pool transaction. In case the round failed before signing step, it seems to me that we cannot recover the pool tx isn't it ?

We can actually, because when the finalization is started, all the forfeit txs (that include the connectors) and the congestion tree have been created and persisted as data of that event.

PR #12, opened 2023-11-22
Support for making ark payments

altafan:
This adds the core functions that enable ark payments:

1. `SpendVtxos()` allows to register the "inputs" of a payment, ie the vtxos for which the asp creates a corresponding forfeit tx (w/ connector input included)
2. `ClaimVtxos()` allows to register the "receivers" of a payment, ie. <pubkey, amount> for which the asp will create corresponding vtxos later in the round
3. `SignVtxos()` allows to send one or many signed forfeit txs, ie. with vtxo from 1) signed by the users

@louisinger please review.

Note that I made changes to the domain layer, I suggest you to take a look how the following structs changed to see if there are any conflicts with your tasks:
- refactored `Vtxo`, now is a composition of `VtxoKey` + `Receiver`
- renamed events and functions for registering "payment inputs" and "outputs" for the `Round` domain

This also adds the following auxiliary data structures to the app layer:
- `paymentsMap`, a queue-like struct allowing to collect payments registration even when the current round is not in registration phase anymore. Allows to update a registered payment by adding receivers. Pops only payments with inputs and receivers in FIFO order. Keeps FIFO order by using timestamps
- `forfeitTxsMap`, a cache for unsigned txs produced by the asp when starting the finalization stage. Allows to push unsigned txs (indexed by txid). Allows to update a forfeit tx and overwrite the stored base64 with the one of the signed tx. Pops only signed txs.

### DIFF
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -50,7 +50,7 @@ func (s *service) SpendVtxos(ctx context.Context, inputs []domain.VtxoKey) (stri
 		}
 	}
 
-	payment := domain.NewPayment(inputs)
+	payment := domain.NewPayment(vtxos)
 	if err := s.paymentRequests.push(payment); err != nil {
 		return "", err
 	}
@@ -65,7 +65,11 @@ func (s *service) ClaimVtxos(ctx context.Context, creds string, receivers []doma
 	}
 
 	// Check that input and output and output amounts match.
-	vtxos, err := s.repoManager.Vtxos().GetVtxos(ctx, payment.Inputs)
+	ins := make([]domain.VtxoKey, 0, len(payment.Inputs))
+	for _, in := range payment.Inputs {
+		ins = append(ins, in.VtxoKey)
+	}
+	vtxos, err := s.repoManager.Vtxos().GetVtxos(ctx, ins)
 	if err != nil {
 		return err
 	}
--- a/internal/core/domain/payment.go
+++ b/internal/core/domain/payment.go
@@ -8,11 +8,11 @@ import (
 
 type Payment struct {
 	Id        string
-	Inputs    []VtxoKey
+	Inputs    []Vtxo
 	Receivers []Receiver
 }
 
-func NewPayment(inputs []VtxoKey) Payment {
+func NewPayment(inputs []Vtxo) Payment {
 	return Payment{
 		Id:     uuid.New().String(),
 		Inputs: inputs,


internal/core/application/service.go
```diff
@@ -107,14 +177,22 @@ func (s *service) finalizeRound() {
 		return
 	}
 
+	forfeitTxs, leftUnsigned := s.forfeitTxs.pop()
+	if len(leftUnsigned) > 0 {
+		err := fmt.Errorf("%d forfeit txs left so sign", len(leftUnsigned))
```
louisinger:
```suggestion
		err := fmt.Errorf("%d forfeit txs left to sign", len(leftUnsigned))
```

internal/core/domain/payment.go
```diff
@@ -14,7 +27,20 @@ func (p Payment) TotOutputAmount() uint64 {
 	return tot
 }
 
-type Vtxo struct {
+func (p Payment) validate(ignoreOuts bool) error {
```
louisinger, line 30: ignoreOuts seems unused, do we really need it ?

internal/core/domain/round.go
```diff
@@ -77,30 +77,31 @@ func (r *Round) On(event RoundEvent, replayed bool) {
...
 	case RoundFinalizationStarted:
 		r.Stage.Code = FinalizationStage
-		r.ForfeitTxs = append([]string{}, e.ForfeitTxs...)
 		r.CongestionTree = append([]string{}, e.CongestionTree...)
 		r.Connectors = append([]string{}, e.Connectors...)
 		r.TxHex = e.PoolTx
 	case RoundFinalized:
 		r.Stage.Ended = true
 		r.Txid = e.Txid
+		r.ForfeitTxs = append([]string{}, e.ForfeitTxs...)
```
louisinger, line 86: Just curious about this, what's the aim of using `append` instead of `r.ForfeitTxs = e.ForfeitTxs` ?

internal/core/domain/round_repo.go
```diff
@@ -16,3 +16,9 @@ type RoundRepository interface {
 		ctx context.Context, id string, updateFn func(r *Round) (*Round, error),
 	) error
 }
+
+type VtxoRepository interface {
+	AddVtxos(ctx context.Context, vtxos []Vtxo) error
```
louisinger, line 21: Not sure we need `AddVtxos` because they are persisted via congestionTree in `AddRound` call (moreover it looks unused)

internal/core/domain/round.go
altafan: This way r.ForfeitTxs is a copy of e.ForfeitTx, otherwise with `r.ForfeitTxs = e.ForfeitTxs` they would share the same variable.

internal/core/domain/round_repo.go
altafan: this is a projection store, the idea is to listen for round events and store the status of all vtxos in this projection store.

internal/core/domain/payment.go
altafan: It's used [here](https://github.com/ark-network/ark/blob/core-functions/internal/core/domain/round.go#L185).

The idea is that, initially, we want to validate only the spending vtxos when a payment is registered, then we want to validate also receivers when they are registered, in a second moment.

PR #14, opened 2023-11-24
`TxBuilder` implementation

louisinger:
**This PR adds `infrastructure/txbuilder` package implementing `ports.TxBuilder` interface**

TL;DR

* `BuildPoolTx` uses ocean's Transfer rpc
* `BuildCongestionTree` uses a "centralized" algorithm to create scriptPubKeys (branches nodes owned by ASP, leaves by receivers)
* `BuildForfeitTxs` creates the connectors list and their associated forfeit transactions.
* `builder_test.go` calls the `BuildForfeitTxs` & `BuildCongestionTree` functions, checking that it does not panic and create the right transactions

@altafan please review


internal/infrastructure/tx-builder/builder.go
```diff
@@ -0,0 +1,210 @@
...
+					b.net,
+				)
+				if err != nil {
+					return nil, nil, err
+				}
+
+				forfeitTxs = append(forfeitTxs, forfeitTx)
+			}
+		}
+	}
```
altafan: @louisinger so we create N forfeit txs for each user, where N is the number of connectors, correct?

internal/infrastructure/tx-builder/builder.go
```diff
@@ -0,0 +1,210 @@
...
+	ctx := context.Background()
+
+	return wallet.Transaction().Transfer(ctx, []ports.TxOutput{
+		newOutput(aspScript, sharedOutputAmount),
+		newOutput(aspScript, connectorOutputAmount),
+	})
+}
+
+func connectorsToInputArgs(connectors []string) ([]psetv2.InputArgs, error) {
+	inputs := make([]psetv2.InputArgs, len(connectors))
```
altafan: from my experience, better to do the following to prevent any kind of panic:
```suggestion
	inputs := make([]psetv2.InputArgs, 0, len(connectors))
```

internal/infrastructure/tx-builder/builder.go
```diff
@@ -0,0 +1,210 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		input := psetv2.InputArgs{
+			Txid:    txID,
+			TxIndex: 0,
+		}
+
+		inputs[i] = input
```
altafan: and here:
```suggestion
		inputs = append(inputs, input)
```

internal/infrastructure/tx-builder/builder.go
```diff
@@ -0,0 +1,210 @@
...
+	feeAmount       = 500
+	connectorAmount = 450
+)
+
+type txBuilder struct {
+	net          *network.Network
+	aspPublicKey *secp256k1.PublicKey
+}
+
+func NewTxBuilder(aspPublicKey *secp256k1.PublicKey, net *network.Network) *txBuilder {
```
altafan: The factory should return a interface, not its implementation

```suggestion
func NewTxBuilder(aspPublicKey *secp256k1.PublicKey, net *network.Network) ports.TxBuilder {
```

internal/infrastructure/tx-builder/builder.go
```diff
@@ -0,0 +1,210 @@
...
+
+func sumReceivers(receivers []domain.Receiver) uint64 {
+	var sum uint64
+	for _, r := range receivers {
+		sum += r.Amount
+	}
+	return sum
+}
+
+var _ ports.TxBuilder = (*txBuilder)(nil)
```
altafan: with the changes to the factory function, this can be removed.

internal/infrastructure/tx-builder/builder.go
louisinger: Yes, for each VTXO to be totally exact

internal/infrastructure/tx-builder/builder.go
altafan: Also, the consumer of this service, ie the app service, doesn't make use of this data type.
What if we accept here an arg of type `common.Network` from which we derive the corresponding liquid network needed by this service?

internal/infrastructure/tx-builder/builder_test.go
```diff
@@ -0,0 +1,285 @@
...
+		},
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return pset.ToBase64()
+}
+
+func TestBuildCongestionTree(t *testing.T) {
```
altafan: In this test we basically verify a tree by hand. We should consider adding a new function `VerifyCongestionTree(poolTx string, tree []string)` to the `common` package, useful also for clients. We could use it here instead to increase the readability of this test.

internal/infrastructure/tx-builder/builder_test.go
```diff
@@ -0,0 +1,285 @@
...
+					Amount: 400,
+				},
+			},
+		},
+	}
+
+	connectors, forfeitTxs, err := builder.BuildForfeitTxs(poolTx, payments)
+	require.NoError(t, err)
+
+	require.Equal(t, 2, len(connectors))
```
altafan: there's a `require.Len` function for this ;)

internal/infrastructure/tx-builder/builder_test.go
```diff
@@ -0,0 +1,285 @@
...
+	forfeitTxsPsets := make([]*psetv2.Pset, 4)
+	for i, pset := range forfeitTxs {
+		forfeitTxsPsets[i], err = psetv2.NewPsetFromBase64(pset)
+		require.NoError(t, err)
+	}
+
+	// each forfeit tx should have 2 inputs and 2 outputs
+	for _, pset := range forfeitTxsPsets {
+		require.Equal(t, 2, len(pset.Inputs))
+		require.Equal(t, 2, len(pset.Outputs))
```
altafan: why 2 outs? I expected only one, ie the ASP. What's the 2nd output for?

internal/infrastructure/tx-builder/forfeit.go
```diff
@@ -0,0 +1,47 @@
...
+	err = updater.AddOutputs([]psetv2.OutputArgs{
+		{
+			Asset:  net.AssetID,
+			Amount: vtxoAmount - feeAmount,
+			Script: aspScript,
+		},
+		{
+			Asset:  net.AssetID,
+			Amount: feeAmount,
+		},
```
altafan: Remove this output. Fees will be payed by the unilateral redeem tx eventually. ALL virtual txs just DON'T pay for any network fees.

internal/infrastructure/tx-builder/tree.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// internal struct to build a binary tree of Pset
+type node struct {
+	receivers          []domain.Receiver
+	left               *node
+	right              *node
+	createOutputScript outputScriptFactory
+	network            *network.Network
+}
```
altafan: didn't know it was possible to define recursive structs in go? 😂

internal/infrastructure/tx-builder/forfeit.go
louisinger: not sure it's the case of forfeit transactions, they should be broadcastable directly no? CPFP in case of forfeit will increase the cost for the ASP

louisinger: you were right, fees removed in connectors and forfeit transactions

PR #22, opened 2023-11-29
Add badger implementations of repos

altafan:
This adds a badger implementation of the repositories useful for testing purposes but also as an initial persistency layer.

This also adds some changes to the domain layer, required after writing their consumers:
- Removed `PaymentsClaimed` event for sake of simplicity since it wasn't used by the app layer
- Added a dedicated method to add receivers to a payment + validation checks
- Added a dedicated method to get the hash of a vtxo key

This adds logic to the application layer meant to keep the projection store up-to-date with the event store: whenever an event is persisted, the aggregate (round/vtxo) resulting from replaying all events is saved to the proper projection store.

Please @louisinger review.

### DIFF
--- a/asp/internal/core/domain/payment.go
+++ b/asp/internal/core/domain/payment.go
@@ -7,7 +7,6 @@ import (
 	"hash"
 
 	"github.com/google/uuid"
-	"golang.org/x/crypto/ripemd160"
 )
 
 type Payment struct {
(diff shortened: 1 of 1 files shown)


asp/internal/infrastructure/db/badger/event_repo.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+type eventRepository struct {
+	store     *badgerhold.Store
+	lock      *sync.Mutex
+	chUpdates chan *domain.Round
+	handler   func(round *domain.Round)
+}
+
+func NewRoundEventRepository(config ...interface{}) (dbtypes.EventStore, error) {
```
louisinger, line 28: any reason to do not type config ?

altafan: because different impls can require different config params.
This way, every impl makes sure that the passed args are exactly those required, instead of having a predefined config that includes all params for the different impls.

PR #23, opened 2023-11-30
add Streaming rpc + implement grpc interface

louisinger:
This PR modifies protobuf spec & implements grpc interface

### Protobuf

 * add `GetEventStream` rpc letting to subscribe to round events: `RoundFinalizationEvent` or `RoundFinalizedEvent`. 
 * change parameter  of `RegisterPaymentRequest`, now expect a set of `Input` (VTXO outpoints to spend).
 * `FinalizePaymentRequest` now accept an array of forfeit transactions (1 per VTXO).
 
### GRPC interface

 * map grpc requests to  `application.Service` methods.
* remove `ListRounds` rpc
 
@altafan please review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,0 @@
-go.work
-go.work.sum


asp/api-spec/protobuf/ark/v1/service.proto
```diff
@@ -56,14 +58,13 @@ message ClaimPaymentRequest {
...
-  string signed_vtx = 1;
+  // Forfeit txs signed by the user. 
+  repeated string signed_forfeits = 1;
 }
 message FinalizePaymentResponse {}
 
 message ListRoundsRequest {
-  int64 start = 1;
-  int64 end = 2;
+  repeated string ids = 1;
```
altafan: @louisinger This rpc is meant to return info about rounds in a specified time window.

The round id is not a public info, rather it's an identifier that means something only to the application layer.
Users could retrieve rounds' info by `txid`s, as this is indeed a public info, but they should have to find those txids first.

Please, revert these changes.

asp/api-spec/protobuf/ark/v1/service.proto
```diff
@@ -34,17 +34,19 @@ service ArkService {
       get: "/v1/round/{txid}"
     };
   };
+  rpc GetEventStream(GetEventStreamRequest) returns (stream GetEventStreamResponse) {
```
altafan, line 31: this is the rpc called by a user once he registers his inputs. Let's try to give it a more descriptive name, one that describes what's happening... something like `ConnectToFinalize` (super ugly IMO)

asp/api-spec/protobuf/ark/v1/service.proto
```diff
@@ -77,13 +78,42 @@ message GetRoundResponse {
...
 message Round {
-  int64 start = 1;
-  int64 end = 2;
-  string txid = 3;
-  repeated Output outputs = 4;
+  string id = 1;
+  int64 start = 2;
+  int64 end = 3;
+  string txid = 4;
+  repeated Output outputs = 5;
```
altafan: let's remove this outputs field in favour of congestion tree and forfeit txs. Maybe event better if we abstract them to something like new vtxos and spent vtxos?

asp/api-spec/protobuf/ark/v1/service.proto
```diff
@@ -77,13 +78,42 @@ message GetRoundResponse {
...
 
 message Output {
   string pubkey = 1;
   uint64 amount = 2;
+}
+
+message RoundFinalizationEvent {
+  string id = 1;
+  string pool_partial_tx = 2;
+  map<string, string> forfeit_txs = 3;
```
altafan, line 93: what would be the key for each txs?

louisinger: transaction ID but thinking about it and it could be a list of `string` instead :thinking: what do you think ?

asp/api-spec/protobuf/ark/v1/service.proto
louisinger: SubscribeToRoundEvents ? ̀SubscribeToRound ? GetRoundUpdates ?

I wouldn't add the "finalize" term into the name, just in case we'll add other type of event (like ̀RoundFailedEvent` for instance)

asp/api-spec/protobuf/ark/v1/service.proto
louisinger: discussed offline, the RPC will be removed (may be re add later if needed by the CLI)

asp/api-spec/protobuf/ark/v1/service.proto
louisinger: **Congestion Tree**
LGTM to add the whole tree instead of abstracting vtxos. The tree has the advantage to be verifiable.

**Forfeit txs**
Not really useful from the wallet point of view IMO: the user doesn't need to know about forfeits once the round is done. He have to take care of them only while finalizing a round, handled by ` RoundFinalizationEvent` message.

altafan: ok let's replace output with congestion tree

asp/api-spec/protobuf/ark/v1/service.proto
altafan: yeah let's go for a list of txs

PR #36, opened 2023-12-04
Add Faucet endpoint

altafan:
This adds a /faucet endpoint to the asp in order to fund a wallet that will be used as a faucet ßervice for end users.

Internally, this pushes a payment to the queue with fixed inputs (txid "0x00..00", vout 0 with of 100k sats) and receivers (10 receivers with user pubkey, 10k sats each).
This also adds some changes with the purpose of NOT keeping track of such fake input vtxos.

No need to start pinging when fauceting, but it is required to open the server side stream to receive forfeit txs that can either be signed or sent back unsigned to complete the flow.

Please @louisinger review.

### DIFF
--- a/asp/internal/core/application/service.go
+++ b/asp/internal/core/application/service.go
@@ -130,7 +130,7 @@ func (s *service) FaucetVtxos(ctx context.Context, pubkey string) error {
 		return err
 	}
 
-	payment.AddReceivers([]domain.Receiver{
+	if err := payment.AddReceivers([]domain.Receiver{
 		{Pubkey: pubkey, Amount: 10000},
 		{Pubkey: pubkey, Amount: 10000},
 		{Pubkey: pubkey, Amount: 10000},
@@ -141,7 +141,9 @@ func (s *service) FaucetVtxos(ctx context.Context, pubkey string) error {
 		{Pubkey: pubkey, Amount: 10000},
 		{Pubkey: pubkey, Amount: 10000},
 		{Pubkey: pubkey, Amount: 10000},
-	})
+	}); err != nil {
+		return err
+	}
 
 	if err := s.paymentRequests.push(*payment); err != nil {
 		return err


asp/internal/core/application/service.go
```diff
@@ -170,49 +206,52 @@ func (s *service) startFinalization() {
 		return
 	}
 
+	var changes []domain.RoundEvent
+	defer func() {
+		if err := s.repoManager.Events().Save(ctx, round.Id, changes...); err != nil {
+			log.WithError(err).Warn("failed to store new round events")
```
louisinger, line 214: shoudn't return an error in the main routine `SignVtxos` ?

asp/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -87,28 +86,28 @@ func (h *handler) ClaimPayment(ctx context.Context, req *arkv1.ClaimPaymentReque
...
 
-		unsignedTx, err := pset.UnsignedTx()
-		if err != nil {
-			return nil, err
-		}
+	return &arkv1.FinalizePaymentResponse{}, nil
+}
 
-		forfeits[unsignedTx.TxHash().String()] = b64
+func (h *handler) Faucet(ctx context.Context, req *arkv1.FaucetRequest) (*arkv1.FaucetResponse, error) {
```
louisinger, line 100: nit: I would add a check of the available amount and return a "not enough funds" error if the wallet is empty

asp/internal/core/application/service.go
altafan: no, a round failed message will be sent to the users instead of the pool tx hash that signals the successful finalization of the round.

asp/internal/interface/grpc/handlers/arkservice.go
altafan: a faucet request is added to the queue just like any other payment. Such a message will be sent to the users through the server-side stream whenever it happens. Rounds are handled by the asp in async.

PR #42, opened 2023-12-04
Noah CLI implementation

louisinger:
**This PR implements the Noah CLI using ark RPCs**

* `send` selects vtxos, handle periodic ping and manage round creation requests.
* `faucet` requests the Faucet RPC. Note: the `FaucetResponse` needs to return RoundID in order to send `GetEventStream`
*  `balance` makes use of the ListVtxos RPC.

Please review @altafan

### DIFF
--- a/noah/config.go
+++ b/noah/config.go
@@ -3,15 +3,9 @@ package main
 import (
 	"fmt"
 
-	"github.com/ark-network/ark/common"
+	arkv1 "github.com/ark-network/ark/api-spec/protobuf/gen/ark/v1"
 	"github.com/urfave/cli/v2"
 )
-
-var rpcUrlFlag = cli.StringFlag{
-	Name:  "rpc",
-	Usage: "ark rpc URL",
-	Value: "",
-}
 
 var configCommand = cli.Command{
 	Name:   "config",
@@ -20,11 +14,8 @@ var configCommand = cli.Command{
 	Subcommands: []*cli.Command{
 		{
 			Name:   "connect",
-			Usage:  "connect <ARK_URL> [--rpc <RPC_URL>]",
+			Usage:  "connect <ARK_URL>",
 			Action: connectAction,
-			Flags: []cli.Flag{
-				&rpcUrlFlag,
-			},
 		},
 	},
 }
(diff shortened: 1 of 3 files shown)


noah/config.go
```diff
@@ -7,15 +7,24 @@ import (
...
+
 var configCommand = cli.Command{
 	Name:   "config",
 	Usage:  "Print local configuration of the Noah CLI",
 	Action: printConfigAction,
 	Subcommands: []*cli.Command{
 		{
 			Name:   "connect",
-			Usage:  "connect <ARK_URL>",
+			Usage:  "connect <ARK_URL> [--rpc <RPC_URL>]",
```
altafan: ~~Let's overwrite temporary `connect <ARK_URL>` with  `init --rpcserver <ASP_URL>`. The first is meant for a nostr-like transpoort, while the second is for a direct connection.~~

~~We'll fix this once we move the asp to nostr-like transpoort.~~

The ark URL is needed to know the asp pubkey. Let's keep both for now, need to think about a better way to configure the client wallet.

noah/send.go
```diff
@@ -37,24 +48,180 @@ func sendAction(ctx *cli.Context) error {
...
+	stream, err := client.GetEventStream(ctx.Context, &arkv1.GetEventStreamRequest{})
+	if err != nil {
+		return err
+	}
+
+	pingStop := ping(ctx, client, &arkv1.PingRequest{
+		PaymentId: registerResponse.GetId(),
+	})
+
+	defer pingStop()
```
altafan: we should stop pinging once we receive the forfeit txs and find some owned by us.

noah/send.go
```diff
@@ -37,24 +48,180 @@ func sendAction(ctx *cli.Context) error {
...
 	return nil
 }
 
-type receiverJSON struct {
-	To     string `json:"to"`
-	Amount int64  `json:"amount"`
+// send 1 ping message every minute (55 secs) to signal to the ark service that we are still alive
+// returns a function that can be used to stop the pinging
+func ping(ctx *cli.Context, client arkv1.ArkServiceClient, req *arkv1.PingRequest) func() {
+	ticker := time.NewTicker(55 * time.Second)
```
altafan: Let's ping every 5 or 10 secs, 55 secs is too much!

noah/balance.go
```diff
@@ -13,6 +13,19 @@ var balanceCommand = cli.Command{
...
+	}
+	defer close()
+
+	vtxos, err := getVtxos(ctx, client)
+	if err != nil {
+		return err
+	}
+
+	balance := computeBalance(vtxos)
+	fmt.Println(balance)
```
altafan: let's always return a json response in all commands, it might turn out useful to build apps on top of this cli if we'll ever need.

noah/send.go
louisinger: > we should stop pinging once we receive the forfeit txs..

Right !

> ...and find some owned by us.

This is done while receiving the forfeits (test on pset inputs)

noah/send.go
```diff
@@ -37,24 +48,181 @@ func sendAction(ctx *cli.Context) error {
...
+						if inputTxid == coin.txid {
+							// TODO: sign the vtxo input
+							signedForfeits = append(signedForfeits, forfeit)
+						}
+					}
+				}
+			}
+
+			if len(signedForfeits) == 0 {
+				return fmt.Errorf("no forfeit to sign")
```
altafan: we shouldn't return error here, if there are no forfeit txs to sign, it means we need to wait for the next round, so we keep the stream open and keep pinging to not loose the priority acquired, like when you call the customer support by phone 🤣

PR #50, opened 2023-12-07
express congestion tree as a matrix instead of a list

### LINKED ISSUE #49: [arkd] Change tree representation to matrix
As things start to take a shape, we saw that representing a tree as a list of base64 txs limits us in many ways. It's hard to understand which of these txs are leaves and unless all txs are decodes, it's not possible to know the parents and child of a node.

Let's change the representation of the tree to a matrix of structs. Every row of the matrix represents a tree level. Every node of the tree is represented by the following struct:

```
type Node struct {
    Txid string
    Tx string
    Parent string
}
```

We can use a specular data structure for the proto message sent to the users.

louisinger:
**This PR adds the domain `CongestionTree` replacing `[]string` type in applications, ports and infrastructure packages.**

TL;DR
* add `Node` type and `CongestionTree` = `[][]Node` in `domain` layer.
* change occurences of `[]string` to express congestionTree members in domain's events and application services.
* update `TxBuilder` interface
* update `txbuilder/dummy` implementation.
* update tests data (tree as matrix)

@altafan please review

it closes #49

### DIFF
--- a/asp/internal/core/application/service.go
+++ b/asp/internal/core/application/service.go
@@ -369,8 +369,6 @@ func getNewVtxos(net network.Network, round *domain.Round) []domain.Vtxo {
 	vtxos := make([]domain.Vtxo, 0)
 	for _, node := range leaves {
 		tx, _ := psetv2.NewPsetFromBase64(node.Tx)
-		utx, _ := tx.UnsignedTx()
-		txid := utx.TxHash().String()
 		for i, out := range tx.Outputs {
 			for _, p := range round.Payments {
 				var pubkey string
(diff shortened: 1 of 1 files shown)


asp/internal/core/application/service.go
```diff
@@ -366,11 +365,10 @@ func (s *service) propagateEvents(round *domain.Round) {
...
-	treeDepth := math.Log2(float64(len(round.CongestionTree) + 1))
-	leaves := round.CongestionTree[int(math.Pow(2, treeDepth)-1):]
+	leaves := round.CongestionTree.Leaves()
 	vtxos := make([]domain.Vtxo, 0)
-	for _, ptx := range leaves {
-		tx, _ := psetv2.NewPsetFromBase64(ptx)
+	for _, node := range leaves {
+		tx, _ := psetv2.NewPsetFromBase64(node.Tx)
 		utx, _ := tx.UnsignedTx()
 		txid := utx.TxHash().String()
```
altafan: this can be replaced with `txid := node.Txid`

PR #57, opened 2023-12-11
Cleanup & Add config and launcher

altafan:
This adds config to be able to configure the daemon at startup via env vars.

This adds app config that takes care of configuring the app service and the services it consumes by validating them first.

This adds a proper interface layer with factory.

This adds `cmd/arkd/main.go` launcher to startup the service.

This also contains several fixes I made to be able to start the service and solves the most evident bugs that I found while achieving the purpose of doing a manual integration tests - that means starting the service, using 2 noah wallets, 1 to faucet&send, 1 to receive and check their balance are correct.

Please, @louisinger review. I made all the commits once everything was final. I suggest you review them 1 by one and read the commit messages as I tried to summarize all the changes in bullet lists.

### DIFF
--- a/asp/Makefile
+++ b/asp/Makefile
@@ -34,6 +34,7 @@ lint:
 run: clean
 	@echo "Running arkd in dev mode..."
 	@export ARK_WALLET_ADDR=localhost:18000; \
+	export ARK_ROUND_INTERVAL=30; \
 	go run ./cmd/arkd
 
 ## test: runs unit and component tests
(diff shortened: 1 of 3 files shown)


asp/internal/infrastructure/scheduler/gocron/service.go
```diff
@@ -26,9 +26,9 @@ func (s *service) Stop() {
 
 func (s *service) ScheduleTask(interval int64, immediate bool, task func()) error {
 	if immediate {
-		_, err := s.scheduler.Every(interval).Seconds().Do(task)
+		_, err := s.scheduler.Every(int(interval)).Seconds().Do(task)
```
louisinger, line 29: nit: change type of `interval` parameter (from int64 to int) ?

PR #60, opened 2023-12-12
Application service: broadcast `RoundFailed` events

### LINKED ISSUE #59: [arkd] RoundFailed events are not broadcasted by GetEventStream rpc
If a `RoundFailed` event is raised in application service, the client does not receive any error and thus continue to ping the daemon.

louisinger:
This PR fixes the `propagateEvents` function to handle `RoundFailed` events.

it closes #59 

@altafan please review

### DIFF
--- a/asp/internal/core/application/service.go
+++ b/asp/internal/core/application/service.go
@@ -388,7 +388,7 @@ func (s *service) propagateEvents(round *domain.Round) {
 			PoolTx:             e.PoolTx,
 			UnsignedForfeitTxs: forfeitTxs,
 		}
-	case domain.RoundFinalized:
+	case domain.RoundFinalized, domain.RoundFailed:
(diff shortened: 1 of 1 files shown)


asp/internal/core/application/service.go
```diff
@@ -388,7 +388,7 @@ func (s *service) propagateEvents(round *domain.Round) {
 			PoolTx:             e.PoolTx,
 			UnsignedForfeitTxs: forfeitTxs,
 		}
-	case domain.RoundFinalized:
+	case domain.RoundFinalized, domain.RoundFailed:
```
altafan, line 391: what if we propagate any event instead by doing something like the following?
 ```suggestion
	default:
```

louisinger: It will increase memory usage for events we won't handle no?

PR #62, opened 2023-12-14
[txbuilder] Instrospection based congestion tree

### LINKED ISSUE #56: Covenant tx builder
Let's add another impl of the tx builder that uses covenants (introspection opcodes) for the output scripts of the txs of the tree.

We can start by focusing on the congestion tree only, and once ocean supports taproot, we can also focus on the pool tx to stop mocking it. 

Note that `TxBuilder` interface might need some changes to support creating the real pool tx.

Reference:
https://docs.google.com/presentation/d/1Z5j31oBLws0RU_wl6PBL-Yf9KgRflW-Ru_t3lyFzR14/edit#slide=id.g29de08d3ad9_0_22

louisinger:
This PR adds a new `covenant/txbuilder` package implementing a covenant based `ports.TxBuilder`.

TL;DR
* implements binary tree using introspections covenants
* add the sweep branch letting the ASP to claim the funds after a relative timelock
* shared output script becomes a taproot script
* add `GetLeafOutputScript` method in `TxBuilder` interface
* remove `BuildCongestionTree` method, `BuildPoolTx` returns the congestion tree

### The taproot covenant

The congestion tree is a binary tree with taproot output scripts. 
Every nodes (leaves and branches) outputs have the sweep taproot leaf. 
The leaves outputs also have a simple checksig leaf associated with the vtxo owner. 
The branches outputs have an introspection script enforcing the spender to follow the tree format. 
Every taproot scripts are using an unspendable internal key.

__leaf taproot script__

```
root
├─ 14 days `CHECKSEQUENCEVERIFY` asp `CHECKSIG`
├─ vtxo_owner `CHECKSIG`
```

**branch taproot script**

branch pset = 1 input 2 outputs. First output (index 0) is the "left" child, the last one (index 1) is the "right" child.

```
root
├─ 14 days CHECKSEQUENCEVERIFY asp CHECKSIG
├─ 0 INSPECT_SCRIPTPUBKEY left_script EQUALVERIFY 0 INSPECT_AMOUNT left_amount EQUALVERIFY            
   1 INSPECT_SCRIPTPUBKEY right_script EQUALVERIFY 1 INSPECT_AMOUNT right_amount EQUAL 
```

it closes #56 

### DIFF
--- a/asp/internal/app-config/config.go
+++ b/asp/internal/app-config/config.go
@@ -127,7 +127,7 @@ func (c *Config) txBuilderService() error {
 	case "covenant":
 		svc = txbuilder.NewTxBuilder(net)
 	default:
-		err = fmt.Errorf("unknown db type")
+		err = fmt.Errorf("unknown tx builder type")
 	}
 	if err != nil {
 		return err


asp/internal/app-config/config.go
```diff
@@ -121,6 +123,8 @@ func (c *Config) txBuilderService() error {
 	net := c.mainChain()
 	switch c.TxBuilderType {
 	case "dummy":
+		svc = txbuilderdummy.NewTxBuilder(net)
+	case "covenant":
 		svc = txbuilder.NewTxBuilder(net)
 	default:
 		err = fmt.Errorf("unknown db type")
```
bordalix:
```suggestion
		err = fmt.Errorf("unknown tx builder type")
```

asp/internal/infrastructure/tx-builder/covenant/builder.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	aspScript, err := p2wpkhScript(aspPubkey, b.net)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	numberOfConnectors := numberOfVTXOs(payments)
```
bordalix, line 100: Do we need 1 connector per input or 1 connector per output?

Consider Alice has 3 VTXO of 10, 20 and 30 and wants to send Bob 60 sats (no change).
The forfeit tx would have 4 inputs (3 for Alice's VTXOs) and just 1 for the connector (not 3).

Or am I think it wrong?

louisinger: > Consider Alice has 3 VTXO of 10, 20 and 30 and wants to send Bob 60 sats (no change). The forfeit tx would have 4 inputs (3 for Alice's VTXOs) and just 1 for the connector (not 3).

A forfeit tx spending the 3 VTXOs is invalid because the forfeit will be broadcastable **if and only if**  the 3 VTXOs are onchain at the same time. In case Alice redeems only 1 of them, the forfeit transaction will be rejected by nodes because the others VTXOs are still "virtual", letting Alice scams the ASP. 

You need 1 connector per VTXO spent in the pool transaction in order to create "independant" forfeit txs that always have the same format: 2 inputs (VTXO + connector), 1 output (ASP script). 

Regarding your exemple, Alice needs 3 connectors to forfeit 3 VTXOs.

bordalix: If the Server has all 3 Alice's VTXOs, and Alice try to double spend VTXO `#1`, can't the Server broadcast VTXO `#2` and `#3` and then the forfeit transaction?

louisinger: Theorically I would say yes totally possible but:
- increase fees for ASP (need to pay for the redeems)
- increase the block space need for a forfeit

PR #70, opened 2023-12-20
add `SweeperService` claiming expired rounds  

louisinger:
This PR add the `RoundSwept` event & `SweeperService` looking for sweep task.

depends on:
- go-elements v0.4.9-rc.4
- https://github.com/vulpemventures/ocean/pull/61

TL;DR
* add `ExpirationTimestamp` & `SweepTxid` to Round domain
* finalization event sets up `ExpirationTimestamp`
* config lets to set up `ARK_ROUND_LIFETIME` defining the relative locktime of outputs.
* start the sweeper at start up
* every 30 secondes, the sweeper tries to check for expired rounds. If it finds some outputs to sweep, it will create a sweep transaction spending them and sending the funds to ASP wallet. Then if succeed sweeper raises `RoundSwept` events.

### DIFF
--- a/asp/internal/app-config/config.go
+++ b/asp/internal/app-config/config.go
@@ -49,6 +49,7 @@ type Config struct {
 	wallet    ports.WalletService
 	txBuilder ports.TxBuilder
 	scanner   ports.BlockchainScanner
+	scheduler ports.SchedulerService
 }
 
 func (c *Config) Validate() error {
@@ -86,6 +87,9 @@ func (c *Config) Validate() error {
 		return err
 	}
 	if err := c.scannerService(); err != nil {
+		return err
+	}
+	if err := c.schedulerService(); err != nil {
 		return err
 	}
 	if err := c.appService(); err != nil {
@@ -190,10 +194,28 @@ func (c *Config) scannerService() error {
 	return nil
 }
 
+func (c *Config) schedulerService() error {
+	var svc ports.SchedulerService
+	var err error
+	switch c.SchedulerType {
+	case "gocron":
+		svc = scheduler.NewScheduler()
+	default:
+		err = fmt.Errorf("unknown scheduler type")
+	}
+	if err != nil {
+		return err
+	}
+
+	c.scheduler = svc
+	return nil
+}
+
 func (c *Config) appService() error {
 	net := c.mainChain()
 	svc, err := application.NewService(
-		c.RoundInterval, c.Network, net, c.wallet, c.repo, c.txBuilder, c.scanner, c.MinRelayFee, c.RoundLifetime, scheduler.NewScheduler(),
+		c.Network, net, c.RoundInterval, c.RoundLifetime, c.MinRelayFee,
+		c.wallet, c.repo, c.txBuilder, c.scanner, c.scheduler,
 	)
 	if err != nil {
 		return err
--- a/asp/internal/core/application/service.go
+++ b/asp/internal/core/application/service.go
@@ -41,31 +41,30 @@ type Service interface {
 }
 
 type service struct {
-	minRelayFee   uint64
-	roundLifetime int64
-	roundInterval int64
 	network       common.Network
 	onchainNework network.Network
 	pubkey        *secp256k1.PublicKey
+	roundLifetime int64
+	roundInterval int64
+	minRelayFee   uint64
 
 	wallet      ports.WalletService
 	repoManager ports.RepoManager
 	builder     ports.TxBuilder
 	scanner     ports.BlockchainScanner
+	sweeper     *sweeper
 
 	paymentRequests *paymentsMap
 	forfeitTxs      *forfeitTxsMap
 
 	eventsCh chan domain.RoundEvent
-
-	sweeper *sweeper
 }
 
 func NewService(
-	interval int64, network common.Network, onchainNetwork network.Network,
+	network common.Network, onchainNetwork network.Network,
+	roundInterval, roundLifetime int64, minRelayFee uint64,
 	walletSvc ports.WalletService, repoManager ports.RepoManager,
 	builder ports.TxBuilder, scanner ports.BlockchainScanner,
-	minRelayFee uint64, roundLifetime int64,
 	scheduler ports.SchedulerService,
 ) (Service, error) {
 	eventsCh := make(chan domain.RoundEvent)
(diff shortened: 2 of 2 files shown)


asp/internal/app-config/config.go
```diff
@@ -74,13 +77,42 @@ func (c *Config) Validate() error {
 	if err := c.appService(); err != nil {
 		return err
 	}
+	if err := c.sweeperService(); err != nil {
+		return err
+	}
+	// round life time must be a multiple of 512
+	if c.RoundLifetime <= 0 || c.RoundLifetime%512 != 0 {
+		return fmt.Errorf("invalid round lifetime, must be greater than 0 and a multiple of 512")
+	}
```
altafan, line 101: do you think it's possible to let the operator specify a lifetime in seconds and we "approximate" it to a correct value instead of returning error?

asp/internal/app-config/config.go
```diff
@@ -137,6 +169,33 @@ func (c *Config) txBuilderService() error {
...
+			func(msg string) {
+				log.Debug(msg)
+			})
+
+	default:
+		return fmt.Errorf("unknown sweeper type")
+	}
+	if err != nil {
+		return err
+	}
```
altafan: this way, if we try to use the dummy tx builder, we won't be able to startup the asp!

Instead, I'm going to add a very dummy sweeper that fires a domain event `RoundExpired` and makes the vtxos expired in the projection repo

asp/internal/config/config.go
```diff
@@ -44,6 +46,7 @@ var (
 	defaultInsecure      = true
 	defaultNetwork       = "testnet"
 	defaultLogLevel      = 5
+	defaultRoundLifetime = (60 * 60 * 24 * 14) - 256 // ~2 weeks in seconds (modulus 512 for sequence)
```
altafan: so this means that we can't set a lifetime smaller than 2 weeks for a demo? 😱

tiero: Dont think sweep its something important for a demo anyway.

louisinger: yes we can, the minimum sequence value in sequence is `512` (seconds)

asp/internal/app-config/config.go
altafan: @louisinger

asp/internal/config/config.go
altafan: @louisinger let's export the env var to set the round lifetime to 8 minutes when running `make run`

asp/internal/core/application/service.go
```diff
@@ -341,7 +360,17 @@ func (s *service) finalizeRound() {
 		return
 	}
 
-	changes, _ = round.EndFinalization(forfeitTxs, txid)
+	now := time.Now().Unix()
+
+	lifetime, err := s.builder.GetLifetime(round.CongestionTree)
```
altafan: why do we ask the tx builder for the round lifetime is that's only a config param that can be customized via env var? The app service shouldn't need to ask for this info to anybody. Can you explain this?

asp/internal/core/application/service.go
```diff
@@ -419,7 +448,12 @@ func (s *service) getNewVtxos(round *domain.Round) []domain.Vtxo {
 
 					buf, _ := hex.DecodeString(r.Pubkey)
 					pk, _ := secp256k1.ParsePubKey(buf)
-					script, _ := s.builder.GetLeafOutputScript(pk, s.pubkey)
+					script, err := s.builder.GetLeafOutputScript(pk, s.pubkey)
+					if err != nil {
+						log.WithError(err).Warn("failed to get leaf output script")
+						continue
```
altafan: can you revert this change?  Here we're parsing the vtxos from the round domain. We can assume that the domain cannot store corrupted data, otherwise it would have been impossible to arrive at this stage.

Also, most likely it would be better to add a `Script` field to the `Vtxo` domain so we don't need to recompute its locking script everytime we need, but this will be done in other pr.

asp/internal/core/application/sweeper.go
```diff
@@ -0,0 +1,274 @@
...
+	repoManager ports.RepoManager
+	builder     ports.TxBuilder
+
+	stop chan struct{}
+}
+
+// Start implements ports.SweeperService.
+func (s *sweeper) Start() error {
+	s.stop = make(chan struct{}, 1)
+	timer := time.NewTicker(30 * time.Second)
```
altafan: This is against the design pattern we are using. No need to poll the db to know if there are vtxos to be swept, we should fire and react to events for that

asp/internal/core/application/sweeper.go
```diff
@@ -0,0 +1,274 @@
...
+				eventsRepo := s.repoManager.Events()
+				roundsRepo := s.repoManager.Rounds()
+				vtxoKeys := make([]domain.VtxoKey, 0)
+
+				for _, sweptOutput := range toSweepOutputs {
+					events, err := sweptOutput.Round.Sweep(
+						sweptOutput.InputArgs.Txid,
+						sweptOutput.InputArgs.TxIndex,
+						txid,
+					)
```
altafan: the round **expires** only once and during this operation one or multiple outputs are going to be **swept**. If we agree on this terminology, we should reflect this statement in code. It doesn't make much sense to me seeing the round being swept many times

asp/internal/core/application/service.go
louisinger: right, let's remove that one :+1:

asp/internal/core/application/sweeper.go
louisinger: It is possible in case someone in the round broadcasted a part of the tree at different times. If, for instance, at least 2 unilateral redeem happens in the round. Some outputs will be swept at redeem_time_1+lifetime, some others at redeem_time_2+lifetime.

altafan: IMO, from the ASP pov, the round could expire at a date X, and then, it could compute if the sweep action needs to be delayed based on the status of the congestion tree as you just described.

My point is that we don't really have a reason to store all the shared outputs as introduced by this pr. When a round expires, the ASP knows that it has to sweep all the non-spent and non-redeemed vtxos. We could already fetch all vtxos for a round since the `Vtxo` domain contains a Txid field.

louisinger: The reasons to store the shared outputs are:
1. Do not recompute the "state of the tree" at each sweep routine. In case of big trees, it means a lot of requests saved.
2. Persist in DB several expirations timestamps, in order to handle the case described above in my first comment
3. Later, for monitoring, it lets the operator knows which sweep tx has swept which part of the tree without any requests to explorer (sweepTxid field in SharedOutput).

PR #79, opened 2024-01-12
Unilateral exit implementation

louisinger:
_This PR implements the  `noah redeem --force` command and fix the congestion tree_

### Congestion tree

* Fix the introspection scripts
* Add a minimal fee output to all transactions. The ASP covers that extra amount
* Remove the last level of the tree (useless). It divides by 2 the number of transactions to broadcast

### CLI 

* add a `bufferutils` pkg letting to encode input witness (in order to finalize congestion tree branches)
* redeem with --force flag will launch a unilateral exit
* redeem all the VTXOs of the wallet
* Subtract fee amount from VTXOs output. Thus, **no more coin selection from on-chain address is needed to redeem**

### Usage

```
> noah redeem --address tex1qdlm2e08aggyuajws4vtxu26sr5fv72dtfcha09 --force
WARNING: unilateral exit (--force) will redeem all your VTXOs
redeeming 10 VTXOs (10000 sats)...
password: 
(1/9) broadcasted tx 989adca325dc1cbb099a1692c337733a8e9fb7ba421785e94c4c1594e62c15b9
(2/9) broadcasted tx be7c949576c1424f53dbad2378d7bcaf1952a1a2b235cf4d29621dac176f8d8c
(3/9) broadcasted tx be136bf39b4bb5186e2e8973d45fa125502b6f01528c938a5c949a313dc410e5
(4/9) broadcasted tx dfa98f6704db9f3259ebb68dfef40bdff0e748bfe3fd4a2d6b69e997ac36147d
(5/9) broadcasted tx ba7c6a79ff33d711d7a82a615eba5964f3ebe170490b9bb4015e7907f106c936
(6/9) broadcasted tx 7d9cc9383f9019eebd3938756c8126d5fae83897f4a756b495ef65e606399833
(7/9) broadcasted tx b0d648305a56922da71be33c8c506f670f30ec44675c4d569a4c4d7ecf7a1da2
(8/9) broadcasted tx 1a3e40af4d7eddefeb16cb20b12e71973737f27eeebafa7a931b4575f48611d8
(9/9) broadcasted tx 729bef7939d41d18096d33cf8facf01a1f195b3ee9d155cda78d9a795dbaedd9
(final) redeem tx d88514e641f4cf0b6a8427b40d467f6a692af947364709bbaa3e8632f3b8ee6d
```
[The last transaction](https://liquid.network/testnet/tx/d88514e641f4cf0b6a8427b40d467f6a692af947364709bbaa3e8632f3b8ee6d) spends all the VTXOs output previously created by broadcasting the tree's branches. The sum of all the VTXOs is 10000 sats, the CPFP fees amount is 650 sats. Thus, the address receives 9350 sats.

### Further work

* https://github.com/ark-network/ark/issues/78
* Update the VTXOs database while an exit happens

Please, @altafan @bordalix @tiero review


asp/internal/core/application/service.go
```diff
@@ -353,7 +355,6 @@ func (s *service) updateProjectionStore(round *domain.Round) {
 				time.Sleep(100 * time.Millisecond)
 				continue
 			}
-			log.Debugf("added %d new vtxos", len(newVtxos))
```
altafan: @louisinger what's the reason for removing this log?

louisinger: It was a debugging one (added by a previous commit). I've just forget to clean log printing.

asp/internal/infrastructure/tx-builder/covenant/tree.go
```diff
@@ -29,24 +29,23 @@ type pluggableCongestionTree func(outpoint psetv2.InputArgs) (domain.CongestionT
...
 		txscript.OP_DATA_32,
-	}...)
+	}
 
 	script = append(script, taprootWitnessProgram...)
 	script = append(script, []byte{
 		txscript.OP_EQUALVERIFY,
 	}...)
-	script = append(script, index...)
+	script = append(script, byte(index))
```
altafan: Unnecessary casting:
```suggestion
	script = append(script, index)
```

asp/internal/infrastructure/tx-builder/covenant/tree.go
```diff
@@ -341,21 +344,53 @@ func newBranch(
...
-// is it the final node of the tree
 func (n *node) isLeaf() bool {
-	return len(n.receivers) == 1
+	return n.left.isSomething() && (n.right == nil || n.right.isSomething())
+}
+
+// is it the final node of the tree
+func (n *node) isSomething() bool {
+	return n.left == nil && n.right == nil
+}
```
altafan: 😂 maybe we can find a better name
```suggestion
func (n *node) isEmpty() bool {
	return n.left == nil && n.right == nil
}
```

noah/redeem.go
```diff
@@ -217,7 +223,177 @@ func collaborativeRedeem(ctx *cli.Context, addr string, amount uint64) error {
...
+	if err != nil {
+		return err
+	}
+
+	if changeAmount > 0 {
+		return fmt.Errorf("unilateral redemption does not allow change, it will redeem all the selected VTXOs for a value of %d", totalVtxosAmount)
+	}
+
+	fmt.Printf("WARNING: unilateral exit (--force) will redeem all your VTXOs\n")
+	fmt.Printf("redeeming %d VTXOs (%d sats)...\n", len(selectedCoins), totalVtxosAmount)
```
altafan: The first check does not make any sense to me: the target amount is the sum of all vtxos, if there's a change it means `coinSelect` is bugged!

I would print a warning message only if the user specified a `--amount` flag to inform him that it would be ignored - and show the tot amount that is gonna be redeemed as we already do.

I would show only the amount redeemed, the number of vtxos is not really interesting to see.

I would also ask for user confirmation before asking for pwd and complete the process.

noah/send.go
```diff
@@ -240,7 +240,11 @@ func ping(ctx *cli.Context, client arkv1.ArkServiceClient, req *arkv1.PingReques
 
 	go func(t *time.Ticker) {
 		for range t.C {
-			client.Ping(ctx.Context, req)
+			_, err := client.Ping(ctx.Context, req)
+			if err != nil {
+				fmt.Printf("error while pinging ASP: %s\n", err)
+				return
+			}
```
altafan: revert and add a `// nolint` comment above L243

noah/signer.go
```diff
@@ -0,0 +1,199 @@
...
+) error {
+	updater, err := psetv2.NewUpdater(pset)
+	if err != nil {
+		return err
+	}
+
+	for i, input := range pset.Inputs {
+		if input.WitnessUtxo != nil {
+			continue
+		}
```
altafan, line 32: why this check? we should not sign the input only if we already did it.. consider that also the asp has to sign such inputs so this check can only cause troubles

louisinger: I may misunderstand your point here, but this for loop aims to populate the pset with witnessUtxo (in order to, later, compute the taproot preimage to sign). This check is here to prevent fetching the input prevout if it is already in the pset.

altafan: why the witness utxos are not yet there? Why do we need to do this operation client-side?

noah/explorer.go
```diff
@@ -0,0 +1,56 @@
...
+	tx, err := transaction.NewTxFromHex(txHex)
+	if err != nil {
+		return "", err
+	}
+	txid := tx.TxHash().String()
+	e.cache[txid] = txHex
+
+	txid, err = broadcast(txHex)
+	if err != nil {
+		if strings.Contains(err.Error(), "Transaction already in block chain") {
```
altafan: This check seems too fragile like this. I'd check the lower case eventually:
```suggestion
		if strings.Contains(strings.ToLower(err.Error()), "transaction already in block chain") {
```

noah/signer.go
louisinger: Because transaction is created client side

PR #87, opened 2024-01-25
Detect redemptions and mark vtxos has redeemed

### LINKED ISSUE #78: The ASP daemon should watch for unilateral exit events

altafan:
This adds support on the asp for detecting redeemed vtxos and mark them as redeemed.

To do so, this adds a new bool `Redeemed` field to the `Vtxo` domain and logic to the application service to start watching vtxos (their locking scripts) when they're created with the help of a new `BlockchainScanner` component (ocean). 

As soon as the app service detects some redemptions, it stops watching for those scripts as we assume unilateral exit to be done over all the offchain funds, and not by cherry-picking vtxos manually. I left some [comments](https://github.com/altafan/ark/blob/detect-redemption/asp/internal/core/application/service.go#L355-L370) about this.

Closes #78.

@louisinger please review.


asp/internal/core/application/service.go
```diff
@@ -440,6 +467,80 @@ func (s *service) getNewVtxos(round *domain.Round) []domain.Vtxo {
...
+		script := hex.EncodeToString(buf)
+
+		indexedScripts[script] = struct{}{}
+	}
+	scripts := make([]string, 0, len(indexedScripts))
+	for script := range indexedScripts {
+		scripts = append(scripts, script)
+	}
+
+	for {
```
louisinger: why this for loop ? may not block the main thread is something wrong happens ocean side ?

asp/internal/core/application/service.go
```diff
@@ -345,6 +348,28 @@ func (s *service) finalizeRound() {
...
 
+func (s *service) listenToRedemptions() {
+	ctx := context.Background()
+	chVtxos := s.scanner.GetNotificationChannel(ctx)
+	for vtxoKeys := range chVtxos {
+		for {
+			// TODO: make sure that the vtxos haven't been already spent, otherwise
+			// broadcast the corresponding forfeit tx and connector to prevent
+			// getting cheated.
+			vtxos, err := s.repoManager.Vtxos().RedeemVtxos(ctx, vtxoKeys)
```
louisinger: I think the `RedeemVtxos` call should occur only if the VTXO has been spent. Here, the notification tells that the vtxo exists on-chain but as long as the user didn't spend it, he can still forfeit it and the ASP is able to sweep it (the VTXO output script includes a sweep taproot branch).

Thus, before calling the RedeemVtxos function, the service should check if the VTXO outpoint is spent or not by the "redeem transaction" = the last tx done while unilateral exit occurs (the one not in the tree but built by clients).

Without that check, it means that if I share a tree's leaf tx with a guy who redeemed, my VTXO will be marked as "redeemed" which is unexpected because I didn't unilateral exit.

asp/internal/core/application/service.go
altafan: It's to ensure that we stop watching the redeemed vtxos. This assumes that the connection with the scanner never drops but if it does so, the service shuts down, so we don't really need to worry about this loop

asp/internal/core/application/service.go
```diff
@@ -81,15 +83,29 @@ func NewService(
...
 func (s *service) Start() error {
 	log.Debug("starting app service")
-	return s.start()
+	go s.start()
+	return nil
 }
 
 func (s *service) Stop() {
+	// nolint
+	vtxos, _ := s.repoManager.Vtxos().GetSpendableVtxos(
```
louisinger, line 102: we should also stop watching the spent vtxos (where spent = true & redeem = false)

asp/internal/core/application/service.go
```diff
@@ -81,15 +83,29 @@ func NewService(
 			svc.propagateEvents(round)
 		},
 	)
+
+	if err := svc.restoreWatchingVtxos(); err != nil {
```
louisinger, line 87: not sure it's working in case a vtxo has been redeemed during the downtime of the ASP (the script is already spent)

I would watch for all not redeemed vtxo's tx: if onchain -> mark as redeemed

altafan: How do we know if the vtxo has been redeemed if we don't start watching it?

asp/internal/core/application/service.go
altafan: no, we stop watching vtxos only at shutdown do clean ocean. The idea is that with only the info coming from ocean notifications we can't know when a script can be unwatched (we should check the balance of its pubkey todo so)

asp/internal/core/application/service.go
altafan: The elegant solution would be always receive the notification from ocean instead of leveraging GetTransaction API.

Let's merge this and add the fix for this use case in a second moment. Maybe we can figure this out by receving txs events instead of utxos from ocean

PR #91, opened 2024-02-08
Support docker

altafan:
This moves the dockerfile to root folder otherwise it's not possible to install `common` inside the container.

This also adds a docker compose file and related volumes to facilitate serving locally with docker.

Please @louisinger @tiero @bordalix review this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,8 +9,7 @@ ARG TARGETARCH
 
 WORKDIR /app
 
-COPY asp ./asp
-COPY common ./common
+COPY . .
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -9,9 +9,10 @@ ARG TARGETARCH
 
 WORKDIR /app
 
-COPY . .
+COPY asp ./asp
+COPY common ./common
```
tiero, line 13: just personal preference if you user WORKDIR to get the right folder, you retain docker layers caching

PR #100, opened 2024-02-09
adds instructions on how to join an ark

bordalix:
@tiero please review

closes https://github.com/ark-network/ark/issues/99

### DIFF
--- a/website/docs/user/ark_cli.md
+++ b/website/docs/user/ark_cli.md
@@ -20,7 +20,7 @@ Available binaries:
 $ OS=$(uname | tr '[:upper:]' '[:lower:]')
 $ ARCH=$(uname -m)
 $ if [ $ARCH = 'x86_64' ]; then GOARCH='amd64'; else GOARCH=$ARCH; fi
-$ curl -Ls https://install-latest-cli.arkdev.info/latest-release/ark-$OS-$GOARCH
+$ curl -O https://install-latest-cli.arkdev.info/latest-release/ark-$OS-$GOARCH
 ```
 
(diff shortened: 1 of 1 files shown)


website/docs/user/ark_cli.md
```diff
@@ -1,16 +1,38 @@
...
+Available binaries:
+
+- ark-darwin-amd64
+- ark-darwin-arm64
+- ark-linux-amd64
+- ark-linux-arm64
+
+```bash
+$ curl -Ls https://install-latest-cli.arkdev.info/latest-release/<binary_name>
+```
```
tiero: Can you re do it in a way you prepare the command for each binary? Just to avoid user extra step to copy-paste and edit manually the curl command

bordalix: 555742b changes to this:

```bash
$ OS=$(eval "go env GOOS")
$ ARCH=$(eval "go env GOARCH")
$ curl -Ls https://install-latest-cli.arkdev.info/latest-release/ark-$OS-$ARCH
```

ok?

bordalix: Ok, that was dumb, since it required Go installed.

I think afd4203 has a better solution:

```
$ OS=$(uname | tr '[:upper:]' '[:lower:]')
$ ARCH=$(uname -m)
$ if [ $ARCH = 'x86_64' ]; then GOARCH='amd64'; else GOARCH=$ARCH; fi
$ curl -Ls https://install-latest-cli.arkdev.info/latest-release/ark-$OS-$GOARCH
```

Tested on Mac and Ubuntu.

PR #101, opened 2024-02-12
adds dumpPrivKey command

bordalix:
ark-cli has a `--prvkey` option on `init`, but I wasn't able to find a command to dump the private key.

```
$ ark init --help
NAME:
   Ark CLI init - initialize the wallet with an encryption password, and connect it to an ASP

USAGE:
   Ark CLI init [command options] [arguments...]

OPTIONS:
   --password value  password to encrypt private key
   --prvkey value    optional, private key to encrypt ❗❗❗
   --network value   network to use (mainnet, testnet) (default: "testnet")
   --ark-url value   the url of the ASP to connect to
   --help, -h        show help
```

As a first experiment on Go, I decided to make one:

```
$ ark dumpPrivKey
password: 
key unlocked
{
	"privateKey": "9c83f3cd5e36a7169e494bf99b632c72399a383a175cdc0f39a945bcbbdf1f88"
}
```

@altafan @louisinger your feedback will be very welcome.

### DIFF
--- a/client/dump.go
+++ b/client/dump.go
@@ -7,7 +7,7 @@ import (
 )
 
 var dumpCommand = cli.Command{
-	Name:   "dumpPrivKey",
+	Name:   "dump-privkey",
 	Usage:  "Dump private key of the Ark wallet",
 	Action: dumpAction,
 }


client/dump.go
```diff
@@ -0,0 +1,24 @@
+package main
+
+import (
+	"encoding/hex"
+
+	"github.com/urfave/cli/v2"
+)
+
+var dumpCommand = cli.Command{
+	Name:   "dumpPrivKey",
```
altafan: the convenience is to have _snake-case-commands_:
```suggestion
	Name:   "dump-privkey",
```

PR #104, opened 2024-02-12
cli logs + handle dust amount

### LINKED ISSUE #53: Increase noah verbosity
Let's add some logs to stdout describing what's happening when `send`ing.
We could add a log once every step is completed.

louisinger:
_This PR improves the CLI verbosity_

bonus: prevent change amount to be lower than dust amount (450) when possible in coin selection.

please @altafan review

it closes #53

### DIFF
--- a/client/common.go
+++ b/client/common.go
@@ -561,7 +561,7 @@ func handleRoundStream(
 			if err != nil {
 				return "", err
 			}
-			fmt.Print("done. \n")
+			fmt.Print("done.\n")
 			fmt.Println("waiting for round finalization...")
 
 			continue
--- a/client/common.go
+++ b/client/common.go
@@ -26,6 +26,10 @@ import (
 	"golang.org/x/term"
 )
 
+const (
+	DUST = 450
+)
+
 func hashPassword(password []byte) []byte {
 	hash := sha256.Sum256(password)
 	return hash[:]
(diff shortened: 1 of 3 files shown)


client/common.go
```diff
@@ -144,10 +147,12 @@ func getServiceProviderPublicKey() (*secp256k1.PublicKey, error) {
 
 func coinSelect(vtxos []vtxo, amount uint64) ([]vtxo, uint64, error) {
 	selected := make([]vtxo, 0)
+	notSelected := make([]vtxo, 0)
```
bordalix, line 150: Why an array? It seems to me it will only be one vtxo or none.

client/common.go
```diff
@@ -161,6 +166,13 @@ func coinSelect(vtxos []vtxo, amount uint64) ([]vtxo, uint64, error) {
 
 	change := selectedAmount - amount
 
+	if change < DUST {
```
bordalix, line 169: Will not prevent case where user has a single 10.300 sats vtxo and is trying to pay 10.000 sats.

louisinger: Yes, made it on purpose. What do you suggest? Remove the change and send the remaining to receiver?

client/common.go
louisinger: Right, in my mind it was to later do a better strategy like "select the VTXO closer to expiry in the array".

Btw the coin selection is really simple now, it would need a complete refacto to be optimal. This is just a quick fix cause I was tired of getting a dust error ^^

client/common.go
```diff
@@ -533,21 +545,24 @@ func handleRoundStream(
...
-			fmt.Printf("%d forfeit txs signed, finalizing payment...\n", len(signedForfeits))
+			fmt.Printf("%d signed\n", len(signedForfeits))
+			fmt.Print("finalizing payment... ")
 			_, err = client.FinalizePayment(ctx.Context, &arkv1.FinalizePaymentRequest{
 				SignedForfeitTxs: signedForfeits,
 			})
 			if err != nil {
 				return "", err
 			}
+			fmt.Print("done. \n")
```
altafan:
```suggestion
			fmt.Print("done.\n")
```
```suggestion
			fmt.Print("done. \n")
```

altafan (approved):

PR #108, opened 2024-02-14
`balance --expiry-details` flag + improve coin selector

### LINKED ISSUE #103: Client should show expiration time to user
Do we show the global expiration time set by ASP ?
Do we show all the vtxos expirations ? Mapped by outpoints ? Mapped by balance ?

### LINKED ISSUE #105: Coin select oldest vtxos first

louisinger:
This PR adds a way to get the expiry timestamp for a given vtxo

* add `--expiry-details` flag on balance letting to show the expiration timestamp(s) of the balance
* use the expiry time in coin selector: sort vtxos by "older first"

@altafan I decided to not add any cache logic for the balance. `ark balance` (without the flag) is instant. I do think it's fine to have 1 seconds loading when we use the flag (combined with a fancy loader, it should be a good UX in my opinion). Let me know what do u think 

it closes #105 
it closes #103 

please @altafan @bordalix review this

### DIFF
--- a/client/common.go
+++ b/client/common.go
@@ -146,33 +146,19 @@ func getServiceProviderPublicKey() (*secp256k1.PublicKey, error) {
 	return pubKey, nil
 }
 
-// olderFirst implements sort.Interface
-type olderFirst []vtxo
-
-func (ls olderFirst) Len() int {
-	return len(ls)
-}
-
-// older vtxos first
-func (ls olderFirst) Less(i, j int) bool {
-	if ls[i].expireAt == nil || ls[j].expireAt == nil {
-		return false
-	}
-
-	return ls[i].expireAt.Before(*ls[j].expireAt)
-}
-
-func (ls olderFirst) Swap(i, j int) {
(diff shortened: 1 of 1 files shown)


client/common.go
```diff
@@ -145,11 +146,34 @@ func getServiceProviderPublicKey() (*secp256k1.PublicKey, error) {
...
+}
+
 func coinSelect(vtxos []vtxo, amount uint64) ([]vtxo, uint64, error) {
 	selected := make([]vtxo, 0)
 	notSelected := make([]vtxo, 0)
 	selectedAmount := uint64(0)
 
+	// sort vtxos by expiration (older first)
+	sort.Sort(olderFirst(vtxos))
+
```
altafan: instead of this, it would be simpler and cleaner to use `sort.SliceStable(vtxos, less&swapFunc())`. Check this example https://github.com/tdex-network/tdex-daemon/blob/v1/internal/core/application/operator/market_account.go#L142-L144

PR #110, opened 2024-02-16
drop private key encoding

### LINKED ISSUE #69: No need to make a different prefix for `apub` and `asec`

louisinger:
This PR removes common functions related to `asec` private key encoding.

@altafan please review

it closes #69

### DIFF
--- a/client/balance.go
+++ b/client/balance.go
@@ -1,18 +1,31 @@
 package main
 
 import (
+	"fmt"
+	"math"
 	"sync"
+	"time"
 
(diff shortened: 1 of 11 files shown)


common/network.go
```diff
@@ -1,25 +1,17 @@
 package common
 
 type Network struct {
-	Name     string
-	SecKey   string
-	PubKey   string
-	RelayKey string
-	Addr     string
+	Name   string
+	SecKey string
```
altafan: Remove `SecKey`

PR #122, opened 2024-02-26
Support connectors validation client-side

altafan:
This makes the client validate the connectors sent by the asp during the round finalization by checking that each tx of the chain contains the expected prevout as input.

For this, the list of connectors has been added to the proto event in order to be received by the client.

Please @louisinger review this.


server/api-spec/protobuf/ark/v1/service.proto
```diff
@@ -116,9 +116,10 @@ message Output {
 
 message RoundFinalizationEvent {
   string id = 1;
-  string pool_partial_tx = 2;
+  string pool_tx = 2;
   repeated string forfeit_txs = 3;
   Tree congestion_tree = 4;
+  repeated string connectors = 5;
```
louisinger, line 122: we decided offline that connectors should not be sent but rebuilt in the client, isn't it ?

altafan: it's easier to directly share them, in a beta version, instead, they would be rebuilt client-side, as the congestion tree i would say

PR #126, opened 2024-03-01
Support connectors validation client-side & Minor fixes

### LINKED ISSUE #86: While signing a forfeit transaction, the client should be able to verify connector input
Forfeit transactions include a connector input related to the current round. Before signing them, the client should check the validity of that connector (is it linked to the next round transaction)

altafan:
This makes the client validate the connectors sent by the asp during the round finalization by checking that each tx of the chain contains the expected prevout as input.

For this, the list of connectors has been added to the proto event in order to be received by the client.

Closes #86 

Please @louisinger review this.

### DIFF
--- a/client/common.go
+++ b/client/common.go
@@ -537,6 +537,15 @@ func handleRoundStream(
 
 			explorer := NewExplorer()
 
+			connectorsTxids := make([]string, 0, len(connectors))
+			for _, connector := range connectors {
+				p, _ := psetv2.NewPsetFromBase64(connector)
+				utx, _ := p.UnsignedTx()
+				txid := utx.TxHash().String()
+
+				connectorsTxids = append(connectorsTxids, txid)
+			}
+
 			for _, forfeit := range forfeits {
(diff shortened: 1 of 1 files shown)


client/common.go
```diff
@@ -430,12 +432,16 @@ func handleRoundStream(
...
 			// validate the congestion tree
 			if err := tree.ValidateCongestionTree(
-				congestionTree, poolTxStr, aspPubkey, int64(roundLifetime),
+				congestionTree, poolTx, aspPubkey, int64(roundLifetime),
 			); err != nil {
 				return "", err
 			}
 
-			exitDelay, err := getUnilateralExitDelay()
+			if err := common.ValidateConnectors(poolTx, connectors); err != nil {
```
louisinger, line 440: I think we should also check that the connectors in forfeit transactions are the one we are validated (an ASP could send a valid connectors list but use others outpoints while building the forfeits.

louisinger: done by 2421382

PR #128, opened 2024-03-04
Add integration tests and regtest support

### LINKED ISSUE #127: Integration test
We need an integration test, at least, for the happy path as:
- start up server
- init 2 clients
- onboard from client 1
- send from client 1 to client 2
- redeem collaboratively from client 1
- redeem unilaterally from client 2

louisinger:
_This PR adds e2e tests using the CLI, the ark daemon & a running ocean wallet_

TL;DR
* [CLI] add hidden & optional flag `--password` wherever it is necessary. If set, skip any user input.
* [Server] support rounds with empty shared output (no vtxos created).
* [Server] add `e2e` package running the `docker-compose.regtest.yaml` file to test CLI commands.
* Support `regtest` network.
* [CLI] add optional `--explorer` flag to `init` command. 
* run `make integrationtest` in Github CI.

it closes #127 

@altafan please review

### DIFF
--- a/client/common.go
+++ b/client/common.go
@@ -202,19 +202,21 @@ func getUnilateralExitDelay() (int64, error) {
 	return int64(redeemDelay), nil
 }
 
-func coinSelect(vtxos []vtxo, amount uint64) ([]vtxo, uint64, error) {
+func coinSelect(vtxos []vtxo, amount uint64, sortByExpirationTime bool) ([]vtxo, uint64, error) {
 	selected := make([]vtxo, 0)
 	notSelected := make([]vtxo, 0)
 	selectedAmount := uint64(0)
 
-	// sort vtxos by expiration (older first)
-	sort.SliceStable(vtxos, func(i, j int) bool {
-		if vtxos[i].expireAt == nil || vtxos[j].expireAt == nil {
-			return false
-		}
+	if sortByExpirationTime {
+		// sort vtxos by expiration (older first)
+		sort.SliceStable(vtxos, func(i, j int) bool {
+			if vtxos[i].expireAt == nil || vtxos[j].expireAt == nil {
+				return false
+			}
 
-		return vtxos[i].expireAt.Before(*vtxos[j].expireAt)
-	})
+			return vtxos[i].expireAt.Before(*vtxos[j].expireAt)
+		})
+	}
 
 	for _, vtxo := range vtxos {
 		if selectedAmount >= amount {
--- a/client/init.go
+++ b/client/init.go
@@ -63,7 +63,7 @@ func initAction(ctx *cli.Context) error {
 		explorerURL = explorer
 		_, network := networkFromString(net)
 		if err := testEsploraEndpoint(network, explorerURL); err != nil {
-			return fmt.Errof("failed to connect with explorer: %s", err)
+			return fmt.Errorf("failed to connect with explorer: %s", err)
 		}
 	} else {
 		explorerURL = explorerUrl[net]
--- a/server/internal/app-config/config.go
+++ b/server/internal/app-config/config.go
@@ -78,24 +78,6 @@ func (c *Config) Validate() error {
 	if c.MinRelayFee < 30 {
 		return fmt.Errorf("invalid min relay fee, must be at least 30 sats")
 	}
-	if err := c.repoManager(); err != nil {
-		return err
-	}
-	if err := c.walletService(); err != nil {
-		return fmt.Errorf("failed to connect to wallet: %s", err)
-	}
-	if err := c.txBuilderService(); err != nil {
-		return err
-	}
-	if err := c.scannerService(); err != nil {
-		return err
-	}
-	if err := c.schedulerService(); err != nil {
-		return err
-	}
-	if err := c.appService(); err != nil {
-		return err
-	}
 	// round life time must be a multiple of 512
 	if c.RoundLifetime < minAllowedSequence {
 		return fmt.Errorf(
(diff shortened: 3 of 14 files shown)


.github/workflows/ark.integration.yaml
```diff
@@ -6,6 +6,11 @@ on:
       - master
     paths:
       - "server/**"
+  pull_request:
+    branches: 
+      - master
+    paths:
+      - "server/**"
```
altafan: nope we don't want to run the integration tests in every pr, it may take too long. Rather, we want to run them only when commits are pushed on master, so please revert this change

server/test/e2e/e2e_test.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		fmt.Printf("error creating ocean wallet: %s", err)
+		os.Exit(1)
+	}
+
+	_, err = runOceanCommand("wallet", "unlock", "--password", password)
+	if err != nil {
+		fmt.Printf("error unlocking ocean wallet: %s", err)
+		os.Exit(1)
+	}
```
altafan, line 38: not needed, we can setup OCEAN_MNEMONIC and OCEAN_PASSWORD env vars

server/test/e2e/e2e_test.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		fmt.Printf("error funding ocean account: %s", err)
+		os.Exit(1)
+	}
+
+	_, err = runCommand("docker", "restart", "arkd")
+	if err != nil {
+		fmt.Printf("error restarting arkd: %s", err)
+		os.Exit(1)
+	}
```
altafan: why restarting the asp?

server/test/e2e/e2e_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+	time.Sleep(2 * time.Second)
+
+	_, err = runArkCommand("init", "--ark-url", "localhost:6000", "--password", password, "--network", "regtest", "--explorer", "http://host.docker.internal:3001")
```
altafan: we should be able to use docker DNS resolution:
```suggestion
	_, err = runArkCommand("init", "--ark-url", "localhost:6000", "--password", password, "--network", "regtest", "--explorer", "nigiri-chopsticks:3001")
```

server/test/e2e/e2e_test.go
louisinger: I do think it's better to have a fresh wallet each time we run the tests (firstly to test the create accounts flow + do not mess with the utxos/txs reusing.)

client/common.go
```diff
@@ -64,22 +72,28 @@ func verifyPassword(password []byte) error {
...
-func readPassword() ([]byte, error) {
+func readPassword(ctx *cli.Context) ([]byte, error) {
 	fmt.Print("unlock your wallet with password: ")
-	passwordInput, err := term.ReadPassword(int(syscall.Stdin))
-	fmt.Println() // new line
-	if err != nil {
-		return nil, err
+	password := []byte(ctx.String("password"))
+
+	if len(password) == 0 {
```
altafan, line 78: Let's move this check at the very beginning of the function so that we don't see the output message unlock your wallet with password: in case the --password flag is used

client/init.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	passwordFlag = cli.StringFlag{
+	passwordSetFlag = cli.StringFlag{
```
altafan: why can't we just use a single `passwordFlag` both to encrypt/decrypt the wallet? we could just change `Usage` a little bit like: `password to encrypt/decrypt the wallet`

client/init.go
```diff
@@ -34,34 +34,48 @@ var (
...
-	Flags:  []cli.Flag{&passwordFlag, &privateKeyFlag, &networkFlag, &urlFlag},
+	Flags:  []cli.Flag{&passwordSetFlag, &privateKeyFlag, &networkFlag, &urlFlag, &explorerFlag},
 }
 
 func initAction(ctx *cli.Context) error {
 	key := ctx.String("prvkey")
 	password := ctx.String("password")
 	net := strings.ToLower(ctx.String("network"))
 	url := ctx.String("ark-url")
+	explorer := ctx.String("explorer")
```
altafan, line 51: We need this flag for the e2e tests on ci?

server/internal/app-config/config.go
```diff
@@ -239,11 +239,16 @@ func (c *Config) appService() error {
...
-		net = network.Testnet
+	switch c.Network.Name {
+	case "liquid":
+		return network.Liquid
+	case "testnet":
+		return network.Testnet
+	case "regtest":
+		return network.Regtest
+	default:
+		return network.Liquid
```
altafan: nit: to avoid code repetition in switch we can write:
```suggestion
	switch c.Network.Name {
	case "testnet":
		return network.Testnet
	case "regtest":
		return network.Regtest
	case "liquid":
		fallthrough
	default:
		return network.Liquid
```

client/init.go
louisinger: The idea of PasswordFlag was to be "hidden"

client/init.go
louisinger: yes + it allows the user to use a custom endpoint

client/init.go
altafan: I liked the idea, is it bad to reuse it also for the `init` cmd?

louisinger: I would say yes cause in the init the flag is required.
OR we can use an input in init if the flag is not set ?

client/init.go
```diff
@@ -68,6 +61,10 @@ func initAction(ctx *cli.Context) error {
 
 	if len(explorer) > 0 {
 		explorerURL = explorer
+		_, network := networkFromString(net)
+		if err := testEsploraEndpoint(network, explorerURL); err != nil {
+			return err
```
altafan:
```suggestion
			return fmt.Errof("failed to connect with explorer: %s", err)
```

PR #129, opened 2024-03-07
wait for boarding confirmation server side

### LINKED ISSUE #125: Onboard: make the asp wait for tx confirmation
In the onboarding flow it is required to wait for the tx to be confirmed in blockchain to complete the flow.
ATM we wait for such confirmation client-side, exposing to side effects in case the user aborts the process.

Let's wait for the onboard tx to be confirmed server-side to get rid of the mentioned adversarial edge case.

louisinger:
This PR modifies the flow of the `/Onboard` rpc: the server is now responsible to broadcast the tx and wait for its confirmation (async).

@altafan please review

it closes #125

### DIFF
--- a/server/internal/core/application/service.go
+++ b/server/internal/core/application/service.go
@@ -405,16 +405,16 @@ func (s *service) handleOnboarding(onboarding onboarding) {
 	for !isConfirmed {
 		select {
 		case <-timeout.C:
-			log.WithError(fmt.Errorf("operation timed out").Warnf("failed to get confirmation for boarding tx %s", txid)
+			log.WithError(fmt.Errorf("operation timed out")).Warnf("failed to get confirmation for boarding tx %s", txid)
 			return
 		default:
 			var err error
 			isConfirmed, _, err = s.wallet.IsTransactionConfirmed(ctx, txid)
 			if err != nil {
 				log.WithError(err).Warn("failed to check tx confirmation")
-				return
 			}
-			if !isConfirmed {
+
+			if err != nil || !isConfirmed {
 				time.Sleep(30 * time.Second)
 			}
 		}


server/internal/core/application/service.go
```diff
@@ -200,24 +211,30 @@ func (s *service) Onboard(
...
-		dustAmount, pubkey, txid, boardingTx, congestionTree, payments,
-	)
+	log.Debugf("broadcasted boarding tx %s", txid)
 
-	return s.saveEvents(ctx, round.Id, round.Events())
+	s.onboardingCh <- onboarding{
+		tx:             boardingTx,
+		congestionTree: congestionTree,
+		userPubkey:     userPubkey,
+	}
```
altafan, line 235: this should be done inside a go routine otherwise it could block the return response to the client

server/internal/core/application/service.go
```diff
@@ -367,6 +384,53 @@ func (s *service) finalizeRound() {
...
+	// wait for the tx to be confirmed with a timeout
+	timeout := time.NewTimer(5 * time.Minute)
+	defer timeout.Stop()
+
+	isConfirmed := false
+
+	for !isConfirmed {
+		select {
+		case <-timeout.C:
+			log.Warnf("boarding tx %s not confirmed (timed out)", txid)
```
altafan: just have a better log:

```suggestion
			log.WithError(fmt.Errorf("operation timed out").Warnf("failed to get confirmation for boarding tx %s", txid)
```

server/internal/core/application/service.go
```diff
@@ -367,6 +384,53 @@ func (s *service) finalizeRound() {
...
+			return
+		default:
+			var err error
+			isConfirmed, _, err = s.wallet.IsTransactionConfirmed(ctx, txid)
+			if err != nil {
+				log.WithError(err).Warn("failed to check tx confirmation")
+				return
+			}
+			if !isConfirmed {
+				time.Sleep(5 * time.Second)
```
altafan: No need to be so much stressful, we can sleep for half the blocktime period here:

```suggestion
				time.Sleep(30 * time.Second)
```

server/internal/core/application/service.go
```diff
@@ -367,6 +384,53 @@ func (s *service) finalizeRound() {
...
+		select {
+		case <-timeout.C:
+			log.Warnf("boarding tx %s not confirmed (timed out)", txid)
+			return
+		default:
+			var err error
+			isConfirmed, _, err = s.wallet.IsTransactionConfirmed(ctx, txid)
+			if err != nil {
+				log.WithError(err).Warn("failed to check tx confirmation")
+				return
```
altafan: Let's consolidate the warning msg, but also, why if instead of returning here and stop watching the tx, we retry and eventually wait for the operation to time out instead?

```suggestion
				log.WithError(err).Warn("failed to get confirmation for boarding tx %s", txid)
				time.Sleep(30 * time.Seconds)
				continue
```

altafan: if you agree, we should consider adding a `defer time.Sleep(30 * time.Seconds) at the beginning of the for loop to remove the repetition of code here and few lines below

server/internal/core/application/service.go
louisinger: the function reading the channel (`listenToOnboarding`) runs go routines, thus it never blocks L231 while we add a new onboarding task

PR #152, opened 2024-04-26
add `--datadir` flag

### LINKED ISSUE #149: CLI: add --datadir flag as other option beside ENV VAR to use a different location for the state
Its' more ergonomic when using CLI for liquid, testnet and regtest at same time

```
ark --datadir /path/data 
```

louisinger:
Move datadir from ENV variable to global `ark` flag

**Usage**

`ark balance` -> default datadir (`~/.ark-cli`)
`ark --datadir ./ balance` -> datadir in the current directory

it closes #149 

@tiero @altafan  please review

### DIFF
--- a/client/main.go
+++ b/client/main.go
@@ -16,8 +16,6 @@ import (
 const (
 	DATADIR_ENVVAR = "ARK_WALLET_DATADIR"
 
-	version = "alpha-0.1.0"
-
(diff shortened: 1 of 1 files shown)


client/state.json
tiero, line 1: remove it

PR #168, opened 2024-05-03
ListVtxos rpc: return swept vtxos

louisinger:
TL;DR

* [protos] add `swept` boolean in `Vtxo` message
* [repository] add `Vtxo.GetAllSweepableVtxos` → replace former `GetAllVtxos` call while stopping the service (in order to unsubscribe scripts). 
* [infrastructure] include Swept vtxos in `GetAllVtxos` method


bordalix (approved):

tiero:
does merging this break the app?

louisinger:
> does merging this break the app?

No, but it might need an update on the app (exclude Vtxos where Swept = true)

PR #172, opened 2024-05-23
do not set expireAt=0 when vtxo is swept

### LINKED ISSUE #170: Return swept timestamp instead
Today, `/v1/vtxos/<ark address>` returns a boolean `swept` to inform if the VTXO was swept by the ASP.

In order for the client to properly know when this swept occurred, it would be better to return a unix timestamp instead (and 0 if is not swept).

@louisinger do you think is possible/easy to make this change?

louisinger:
This PR removes the reset statement on expireAt member. Thus, a swept Vtxo will keep its `expireAt` timestamp value.

it closes #170 

@bordalix @tiero please review

### DIFF
--- a/server/.gitignore
+++ b/server/.gitignore
@@ -22,3 +22,5 @@ DS_Store
 
 go.work
 go.work.sum
+
+.vscode/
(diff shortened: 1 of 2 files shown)


bordalix (approved):

bordalix:
Easiest PR review ever 😄

tiero (approved):

PR #173, opened 2024-05-23
Dynamic amount in unroll closure

### LINKED ISSUE #153: Use the final Boarding Script

louisinger:
This PR reworks the unroll closure in `common` package. The new version uses `OP_INSPECTINPUTVALUE` to compute and verify output amount. It lets to have a **constant onboarding address**.

TL;DR
* `UnrollClosure` with 1 child now use the "dynamic introspection script"
* `TrustedOnboarding` do not stop watching onboarding scripts.
* `ark-cli onboard --trusted` does not expect an amount flag anymore.
* add a new test for Trusted onboard in `e2e_test.go` 

`UnrollClosure.Decode` is still able to decode former version of the script in order to do not break rounds made before this change.

related to https://github.com/ark-network/tapscripts/pull/2

it closes #153 

@tiero @altafan @bordalix please review


tiero:
I think as we are moving to different approach with the baording address, we dont need anymore the "trusted" concept imho. I would replace this new script in the current Boarding RPC directly, no need to maintain old script.

altafan (approved):

louisinger:
> I think as we are moving to different approach with the baording address, we dont need anymore the "trusted" concept imho. I would replace this new script in the current Boarding RPC directly, no need to maintain old script.

sorry, OP was maybe not clear:
all onbaording types are using the new script type. We need backward compatibly (it's 1 line in Decode function) for the rounds created before (in case someone wants to decode the scripts to unilateral exit). In other words, Boarding and TrustedOnboarding are using the same "dynamic" script version.

bordalix (approved):

tiero:
Feel free to merge if you tested @altafan

PR #176, opened 2024-05-23
add AdminService 

### LINKED ISSUE #151: Admin RPC
- Start & Stop Round
 - Update Round terms for next round (round lifetime, exit delay)
 - Query transaction and round history
 - Balance Breakdown (how much is collected from fees and from sweep)

louisinger:
This PR adds a gRPC admin service to the ASP daemon. This service is protected using basic authorization token.

TL;DR
* [config] `AUTH_USER` & `AUTH_PASS` lets to define the authorization token (default to admin:admin).
* [application] new AdminService implements `admin.proto`.
* [interface] authorization interceptor + admin handler.

### Admin API

* `GetBalance` returns ocean accounts balance (connectors and main account)
* `GetScheduledSweep` returns the rounds outputs that will be swept (with schedulet_at timestamp).
* `GetRounds` lets to search for round ids between a range of timestamp
* `GetRoundDetails` returns stats about the round (vtxos amount, vtxos txid, forfeit amounts, exit amount ...)

it closes #151 

@tiero @altafan @bordalix please review


server/api-spec/protobuf/ark/v1/admin.proto
```diff
@@ -0,0 +1,79 @@
...
+  repeated string exit_addresses = 9;
+}
+
+message GetRoundsRequest {
+  int64 after = 1;
+  int64 before = 2;
+}
+
+message GetRoundsResponse {
+  repeated string rounds = 1;
```
tiero, line 83: is this round ID?

server/api-spec/protobuf/ark/v1/admin.proto
```diff
@@ -0,0 +1,79 @@
...
+    };
+  }
+}
+
+message GetBalanceRequest {}
+
+message GetBalanceResponse {
+  string main_account = 1;
+  string connectors_account = 2;
+}
```
tiero: The most important thing imho is not the balance available, but a complete breakdown of "available" capacity and "locked" capacity. I know can be computed via GetRoundDetails, but also mean listing all the round IDs and call many times the

server/api-spec/protobuf/ark/v1/admin.proto
louisinger: yes

server/api-spec/protobuf/ark/v1/admin.proto
louisinger: the problem is that the "locked" capacity is related to the expirationTime. How should I show this balance ? the sum of all "scheduled sweeps" ? or something like this:
```json
{
    "24h": "0.0006",
    "1week": "0.002",
    "1month": "0.0001",
}
```
meaning that in 1 day the wallet will sweep 0.0006, in 1 week 0.002 etc...

bordalix (approved):

server/api-spec/protobuf/ark/v1/admin.proto
altafan: I think for now it could be enough to show the total locked amount instead of having such a high level of detail

altafan: along with the availble balance, for both main and connector accounts

louisinger: fixed by 5ff9c379f126a35d66c270ed39058b6b6d96b6d5

PR #180, opened 2024-05-30
Implements SQLite repositories

### LINKED ISSUE #177: Add SQLite repository
https://pkg.go.dev/modernc.org/sqlite

louisinger:
This PR adds a new db type `sqlite` in config associated with `RoundRepository` and `VtxoRepository` implementations in `infrastructure/db/sqlite` package.

it closes #177 

@altafan @tiero please review


tiero (commented):
Dockerfile may need an update?

louisinger:
> Dockerfile may need an update?

SQLite isn't embedded? `ARK_DB_TYPE=sqlite` should be enough to switch on sqlite DB (beside migrations)

server/internal/app-config/config.go
```diff
@@ -147,6 +150,22 @@ func (c *Config) repoManager() error {
 			RoundStoreConfig: []interface{}{c.DbDir, logger},
 			VtxoStoreConfig:  []interface{}{c.DbDir, logger},
 		})
+	case "sqlite":
+		var database *sql.DB
+		database, err = sqlitedb.OpenDB(c.DbDir + "/sqlite.db")
+		if err != nil {
+			return err
+		}
```
altafan: this must be moved to `db.NewService()`!

altafan: we can pass

```
RoundStoreConfig: []interface{}{c.DbDir, logger},
VtxoStoreConfig:  []interface{}{c.DbDir, logger},
```

just like the badger db config

server/internal/app-config/config.go
```diff
@@ -147,6 +150,22 @@ func (c *Config) repoManager() error {
...
+
+		svc, err = db.NewService(db.ServiceConfig{
+			EventStoreType: c.DbType,
+			RoundStoreType: c.DbType,
+			VtxoStoreType:  c.DbType,
+
+			EventStoreConfig: []interface{}{c.DbDir, log.New()},
+			RoundStoreConfig: []interface{}{database},
+			VtxoStoreConfig:  []interface{}{database},
```
altafan: we can just pass the same config we use for badger in this case:
```suggestion
			RoundStoreConfig: []interface{}{c.DbDir, logger},
			VtxoStoreConfig:  []interface{}{c.DbDir, logger},
```

common/tree/congestion_tree.go
```diff
@@ -1,18 +1,20 @@
 package tree
 
 import (
+	"encoding/json"
```
altafan: @louisinger revert the changes to this file, we can't afford to store nodes as json strings in the db. Let's store a tree as a list of nodes with a foreign key referencing t the round, maybe we can use also another field to facilitate reconstructing the tree as [][]Node.

tiero: Agreed, better to deconstruct and make the necessary SQL relations between data

tiero:
> > Dockerfile may need an update?
> 
> SQLite isn't embedded? `ARK_DB_TYPE=sqlite` should be enough to switch on sqlite DB (beside migrations)

Exactly, we need to copy the migration file inside the container as well. We can do later, but test a docker image locally first

server/internal/app-config/config.go
louisinger: done

server/internal/app-config/config.go
louisinger: `c.DbDir`: isn't better to open the DB only once and then pass the `*sql.Db` to repositories ? If I pass the path instead, I will execute 2 times the `OpenDb` function (expensive ?)

I don't see the point to pass the logger ? Do you mean we should use a custom logger to print some "Debug" logs ?

louisinger:
**This PR is not ready to be merged.** The repositories are not handling correctly concurrent requests (SQLITE_BUSY errors while running integration tests using sqlite). WIP

server/internal/app-config/config.go
altafan: > c.DbDir: isn't better to open the DB only once and then pass the *sql.Db to repositories ? If I pass the path instead, I will execute 2 times the OpenDb function (expensive ?)

totally agree, but let's do it in the repo manager, not here in config.. we did the same for badger, the change is as simple as copy-pasting the existing code (mostly)

altafan:
> Exactly, we need to copy the migration file inside the container as well. We can do later, but test a docker image locally first

The migrations are hardcoded currently. We need to move to `sqlc` for better handling migrations and query definitions.
cc @louisinger

tiero:
We have always used this go package in other go  projects so far. You have to bring the files inside the docker container 

https://github.com/golang-migrate/migrate

server/internal/infrastructure/db/sqlite/round_repo.go
```diff
@@ -104,50 +122,25 @@ ON CONFLICT(payment_id, pubkey) DO UPDATE SET
...
-
-	selectSweepableRounds = `
-SELECT round.id, round.starting_timestamp, round.ending_timestamp, round.ended, round.failed, round.stage_code, round.txid, round.unsigned_tx, round.congestion_tree, round.forfeit_txs, round.connectors, round.connector_address, round.dust_amount, round.version, round.swept, payment.id, receiver.payment_id, receiver.pubkey, receiver.amount, receiver.onchain_address, vtxo.txid, vtxo.vout, vtxo.pubkey, vtxo.amount, vtxo.pool_tx, vtxo.spent_by, vtxo.spent, vtxo.redeemed, vtxo.swept, vtxo.expire_at, vtxo.payment_id
+	selectRound = `
+SELECT round.id, round.starting_timestamp, round.ending_timestamp, round.ended, round.failed, round.stage_code, round.txid, 
+round.unsigned_tx, round.connector_address, round.dust_amount, round.version, round.swept, payment.id, receiver.payment_id, 
+receiver.pubkey, receiver.amount, receiver.onchain_address, vtxo.txid, vtxo.vout, vtxo.pubkey, vtxo.amount, 
+vtxo.pool_tx, vtxo.spent_by, vtxo.spent, vtxo.redeemed, vtxo.swept, vtxo.expire_at, vtxo.payment_id, 
+tx.tx, tx.type, tx.position, tx.txid, 
+tx.tree_level, tx.parent_txid, tx.is_leaf
```
tiero, line 130: do we really need ALL of theese fields in every repo fetch? I do think we can have multiple repo methods mapped to different subset of fields, to be mor efficient to only query the minimum data that you need for the method

louisinger: I agree, however it needs a complete refactor of the interface (all getters return the whole Round object), it also means rewrite badger implementation and tests. Let's do that in a dedicated PR no ?

tiero:
To avoid concurrent writes you should a) refactor tables in a way if two components likely write at same time, best to work against two tables b) use mutex on a holder structure so only one writer per time easily  c) if you get SQLite Busy error, simply retry in a second or so

tiero:
Forcing the pool to only one option is the OPPOSITE of what you want to achieve :) you want to maximize reusing connection pool across multiple clients

https://chatgpt.com/share/83bcaf40-ed87-4f5e-9d4a-ede07a2f94ee

tiero:
Why we don't have one events table in SQLite as well? So we have single DB for all repos no?

tiero:
> > To avoid concurrent writes you should a) refactor tables in a way if two components likely write at same time, best to work against two tables b) use mutex on a holder structure so only one writer per time easily c) if you get SQLite Busy error, simply handle that in the code and keep retry-ing until the write does go through
> 
> 
> 
> https://github.com/ark-network/ark/pull/180/commits/8f993926a3a9330a4933b1ab06666083bd644f7a fixes that SQLITE_BUSY errors.  Still the integrations tests are hanging (but does not return this SQLITE_BUSY errors)

But if we don't allow more than one open connection it means we can serve a single user and a single requests at the time? (that are writes? I guess read are not counted but SQL busy errors?)

server/internal/infrastructure/db/sqlite/round_repo.go
tiero: Open an issue about it! Will likely be a easy optimization

louisinger:
> Why we don't have one events table in SQLite as well? So we have single DB for all repos no?

It seems that we are never reading the event DB (we only push events without reading them once inserted). That's why we decided to re-use the badger implementation. cc @altafan 

> But if we don't allow more than one open connection it means we can serve a single user and a single requests at the time? (that are writes? I guess read are not counted but SQL busy errors?)

yes readers are not counted, several processes may read at the same time.

tiero:
Ok but if we can only open a connection at the time, if 100 users are joining a round (and need to write?) what will happen?

tiero:
https://chatgpt.com/share/851e4ef6-a082-459e-89a5-193bb756b4c6

PR #191, opened 2024-06-25
Admin UI

bordalix:
WIP


tiero:
Stylistic, nit: I think it's nicer to have the /server' level ie. `/server/web` or `/server/dashboard`

bordalix:
> Stylistic, nit: I think it's nicer to have the /server' level ie. `/server/web` or `/server/dashboard`

This is definitely work in progress 🤣 

I made the PR because I need @altafan help in "merging" the two servers into one:

- API is running on port 8000
- Dashboard is running on port 7000

I want them to be running on the same port, which means endpoint paths will need to change.

PR #193, opened 2024-07-02
Client SDK

### LINKED ISSUE #192: Client SDK
We want to create a reusable library that clients can consume for interaction with the ASP.

The idea is to move (most of) the code within the CLI commands to a standalone package that can then be imported into other client wallet projects.

Let's provide both grpc and rest clients.

sekulicd:
This adds client SDK(grpc, rest, wasm) for accessing Ark server.

@tiero @altafan please review.

Closes #192.


sekulicd:
> No need for grpc-web support, we can drop the grpc-web client too.
> 
> @tiero suggested to put the SDK in pkg/client-sdk, is it ok to have it in common? Also, it would be appropriate to create a dedicated module for the SDK (ie cd into client-sdk and `go mod init github.com/ark-network/ark-sdk`)
> 
> Let's proceed then by using the SDK within the CLI, this way we have the integration tests as testing suite for the SDK.

@altafan what about @tiero [comment](https://github.com/ark-network/ark/issues/192#issuecomment-2200434451) ?

tiero:
> pkg/client-sdk,

yes lets move it in root, not inside common

>  we can drop the grpc-web client too.

From what I understood, when compiling to WASM targets, we should use gRPC-web like the example I found. Not sure if using JSON interface instead, we can then do code-gen of client stubs like with gRPC and gRPC web?

tiero:
Dumb question @sekulicd what happens if you compile to WASM classic gRPC client instead? does it gives you error? I understand browser would then not suppport http2 streaming out of the box, but should not be an issue for unary requets?

Just a curiosity

common/client-sdk/grpcweb_client.go
```diff
@@ -0,0 +1,90 @@
...
+	port := 80
+	if strings.HasPrefix(aspUrl, "https://") {
+		aspUrl = strings.TrimPrefix(aspUrl, "https://")
+		creds = credentials.NewTLS(nil)
+		port = 443
+	}
+	if !strings.Contains(aspUrl, ":") {
+		aspUrl = fmt.Sprintf("%s:%d", aspUrl, port)
+	}
+	conn, err := grpc.NewClient(aspUrl, grpc.WithTransportCredentials(creds))
```
tiero: is this really using a gRPC web client? Seems to be same as gRPC native?

common/client-sdk/rest_client.go
```diff
@@ -0,0 +1,357 @@
...
+
+func (c *arkRestClient) Admin() AdminServiceClient {
+	return &restAdminServiceClient{baseURL: c.baseURL, client: c.httpClient}
+}
+
+func (c *arkRestClient) Service() ArkServiceClient {
+	return &restArkServiceClient{baseURL: c.baseURL, client: c.httpClient}
+}
+
+type restAdminServiceClient struct {
```
tiero: is this all written by hand? NO way to do code-gen from Swagger?

altafan: https://goswagger.io/go-swagger/about/

common/client-sdk/grpcweb_client.go
sekulicd: > From what I understood, when compiling to WASM targets, we should use gRPC-web like the example I found. Not sure if using JSON interface instead, we can then do code-gen of client stubs like with gRPC and gRPC web?

Wasm [file](https://github.com/johanbrandhorst/grpcweb-wasm-example/blob/master/frontend/html/test.wasm) is generated using [this](https://github.com/johanbrandhorst/grpcweb-wasm-example/blob/50cb9e39fa989ec5c2af9d3200e2d278cf72c034/Makefile#L5) cmd, if u check that [file](https://github.com/johanbrandhorst/grpcweb-wasm-example/blob/master/frontend/frontend.go) u will see in main it creates usual grpc connection. If you check [server](https://github.com/johanbrandhorst/grpcweb-wasm-example/blob/50cb9e39fa989ec5c2af9d3200e2d278cf72c034/main.go#L40) which is hosting FE files(wasm, html) and exposing grpc API's you will see that it is using [grpcweb](https://github.com/johanbrandhorst/grpcweb-wasm-example/blob/50cb9e39fa989ec5c2af9d3200e2d278cf72c034/main.go#L40).

common/client-sdk/rest_client.go
sekulicd: I will check this, code was written by hand.

tiero:
In the commit that removes `grpc-web` we are also removing the WASM toolchain and glue code from examples. Can you reintroduce it, but now using the `rest` package?

altafan:
> lets move client-sdk from `common` to `pkg` already as it is being imported by CLI

Let's also create a dedicated submodule:

```
$ cd pkg/client-sdk
$ go mod init github.com/ark-network/ark-sdk
```

in client's go.mod we need to add also the following line:

```
replace github.com/ark-network/ark-sdk => ../pkg/client-sdk
```

common/client-sdk/wasm/service_client.go
```diff
@@ -0,0 +1,380 @@
...
+var (
+	invalidArgsError = errors.New("invalid number of arguments")
+)
+
+func main() {
+	js.Global().Set("RestServiceClient", RestServiceClientWrapper())
+	js.Global().Set("InvokeRestServiceClientMethod", RestServiceClientWrapper())
+	js.Global().Set("Log", Log())
+	select {}
+}
```
tiero: Here we should expose things like Balance, Send etc..

tiero:
The Admin service is not needed

sekulicd:
@tiero @altafan i created [sdk](https://github.com/sekulicd/ark/blob/05644dad2adcd06433f4c98267d06164c77e5bc3/pkg/client-sdk/ark_sdk.go#L44) and added [example](https://github.com/sekulicd/ark/blob/client-sdk/pkg/client-sdk/example/main.go).
Note that for REST protocol some of sdk methods wont work cause [this](https://github.com/sekulicd/ark/blob/05644dad2adcd06433f4c98267d06164c77e5bc3/pkg/client-sdk/inner_client.go#L102) is not implemented yet as i need to check if grpc-gateway supports REST(i guess websockets are needed) to Stream. 
Also WASM wrapper is missing.
@altafan please check if signing is properly done.
I guess we can work on wallet abstraction and think about if general cleaning of ark_sdk.go is possible so its more maintainable/readable.

pkg/client-sdk/example/main.go
```diff
@@ -0,0 +1,157 @@
...
+		ctx         = context.Background()
+		explorerSvc = arksdk.NewExplorer(explorerUrl)
+	)
+
+	alice, err := arksdk.NewSingleKeyWallet(
+		explorerSvc, network, "",
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
```
altafan: would be ideal to instantiate this into the factory function below `arkdsk.New()` by providing a `walletType="singleKey"` or similar param

pkg/client-sdk/example/main.go
```diff
@@ -0,0 +1,157 @@
...
+
+	txID, err := aliceArkClient.Onboard(ctx, 20000)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Infof("Alice onboarded with txID: %s", txID)
+
```
altafan: we should mint a block with `nigiri rpc --liquid generatetoaddress 1 $(nigiri rpc --liquid getnewaddress)` and check that the balance has been updated correctly

pkg/client-sdk/example/main.go
```diff
@@ -0,0 +1,157 @@
...
+				To:     bobOffchainAddr,
+				Amount: 10000,
+			},
+		})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Infof("Alice sent 10000 to Bob offchain with txID: %s", txID)
+}
```
altafan: same here, let's mine a block and check the balances of both wallets

altafan (commented):
@sekulicd atm the "database" management is in the CLI, ie reading the wallet keypair from the json file in the cli datadir.

Considering that we could support an HD wallet with btc-wallet, that comes with a full managed db, wouldn't it be better to move this part of code to the SDK too?

tiero:
> this part of code to the SDK too?

Yes would be best to abstract also the storage component with a common interface, so that based in the target (linux server, wasm browser etc..) we can use the most appropriate:

- SQLite for simple of deployemnt of Ark nodes
- Postgres for enterprise Ark node
- IndexDB  for light clients in the browser 
- AsyncStorage for mobile apps with React native

etc..


Also BitcoinDevKit does the same, and provdes some reference implementations 👇 
https://docs.rs/bdk_wallet/latest/bdk_wallet/#persistence

We can have SQLite driver using the native Go impl (which it may also compile to WASM? maybe! big if possible)

common/Makefile
```diff
@@ -1,2 +1,6 @@
 test:
-	go test -v ./...
\ No newline at end of file
+	go test -v ./...
+
+genrest:
```
tiero: Does this genrest belongs to common package?

pkg/client-sdk/example/main.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	_, err = runCommand("nigiri", "faucet", "--liquid", aliceOnchainAddr)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	txID, err := aliceArkClient.Onboard(ctx, 20000)
```
tiero: Does this Onbord does all the dance under the hood?

- Liquid is like creating an addess and sending a tx
- Bitcoin is doing two txs, Funding and Redeem, interactively with ASP (pending API/RPC I guess @altafan )

common/Makefile
sekulicd: No, its leftover, will move it.

pkg/client-sdk/example/main.go
sekulicd: Code is mostly copy of what is already done in cli.

sekulicd:
Regarding comments:
> the "database" management is in the CLI, ie reading the wallet keypair from the json file in the cli datadir.
Considering that we could support an HD wallet with btc-wallet, that comes with a full managed db, wouldn't it be better to move this part of code to the SDK too?

> Yes would be best to abstract also the storage component with a common interface

IMO sdk should be stateless(separated from wallet) its not an ark wallet but sdk others can use to create ark wallet which will have persistence.
For this reason i provided Wallet interface as arg to Ark Sdk factory func in [here](https://github.com/sekulicd/ark/blob/f3c7d1b02e0e8a24f74e9f031802e678f4bc8062/pkg/client-sdk/ark_sdk.go#L61)

 @tiero @altafan

pkg/client-sdk/example/main.go
sekulicd: Check [this](https://github.com/ark-network/ark/pull/193#issuecomment-2217200874) comment, i think it would be ideal if we come up with good wallet abstraction and pass it as interface than to "leave" sdk to do creation of concrete wallets

pkg/client-sdk/example/main.go
sekulicd: Check https://github.com/ark-network/ark/pull/193#issuecomment-2217200874 comment, i think it would be ideal if we come up with good wallet abstraction and pass it as interface than to "leave" sdk to do creation of concrete wallets

pkg/client-sdk/example/main.go
altafan: The idea is to abstract all the dance behind this method. For now we are working on the general SDK interface and its liquid implementation, we'll work on the bitcoin impl as soon as the bitcoin version of the ASP is released.

altafan:
> IMO sdk should be stateless(separated from wallet) its not an ark wallet but sdk others can use to create ark wallet which will have persistence. For this reason i provided Wallet interface as arg to Ark Sdk factory func in [here](https://github.com/sekulicd/ark/blob/f3c7d1b02e0e8a24f74e9f031802e678f4bc8062/pkg/client-sdk/ark_sdk.go#L61)
> 
> @tiero @altafan WDYT?

Try to take a look at [btcwallet](https://github.com/btcsuite/btcwallet/): it is a complete wallet SDK that we used to implement the embedded wallet for the bitcoin version of the service.

It comes with a fully managed db and the operator can decide which kind of db he wants to use. It's super handy and leaves to the dev only the job of creating an interface around the imported package. We want to have something similar for our SDK.

I agree with what tiero suggested [in his comment](https://github.com/ark-network/ark/pull/193#issuecomment-2214802452).

tiero:
A Bitcoin wallet is stateful by nature and the main source of errors is usually efficient caching and retrieval of the UTXO set (or in this case VTXO set eheh)

pkg/client-sdk/example/main.go
tiero: I see the benefit of decoupling wallet to be standalone packages, but in the end we are worsening the Dev UX, as it will result in more lines of code.

In this specific I am not sure we should call it as SingleKey, rather we should identify it based on the secret to be passed:

- PrivateKey 
- Mnemonic 


The fact in our implementation we decide to do public key reuse doesn't matter if the SDK is loaded with a mnemonic for example.

tiero:
It's totally ok to define a DB interface and pass the concrete impl when instantiating the SDK

common/client-sdk/wasm/service_client.go
sekulicd: Done

pkg/client-sdk/example/main.go
sekulicd: Done

pkg/client-sdk/example/main.go
sekulicd: Done

sekulicd: Done

common/Makefile
sekulicd: Done

sekulicd:
> It's totally ok to define a DB interface and pass the concrete impl when instantiating the SDK

@tiero @altafan please review Config and Wallet abstraction, i guess this can be improved, but current solution covers code in cli, we can think of HD wallet abstraction and how to support it.

pkg/client-sdk/wallet.go
```diff
@@ -0,0 +1,273 @@
...
+	"github.com/vulpemventures/go-elements/network"
+	"github.com/vulpemventures/go-elements/payment"
+	"github.com/vulpemventures/go-elements/psetv2"
+	"github.com/vulpemventures/go-elements/transaction"
+)
+
+type Wallet interface {
+	PubKey() *secp256k1.PublicKey
+	PubKeySerializeCompressed() []byte
+	SignPsetForAddress(pset *psetv2.Pset, address string) error
```
tiero: SignTransactionForAddress would work both on bitcoin and elements.

pkg/client-sdk/ark_sdk.go
```diff
@@ -0,0 +1,1413 @@
...
+		network.Testnet.Name: "https://blockstream.info/liquidtestnet/api",
+		network.Regtest.Name: "http://localhost:3001",
+	}
+)
+
+type ArkClient interface {
+	Connect(ctx context.Context) error
+	Balance(ctx context.Context, computeExpiryDetails bool) (*BalanceResp, error)
+	Onboard(ctx context.Context, amount uint64) (string, error)
+	TrustedOnboard(ctx context.Context) (string, error)
```
tiero: This can be removed I guess @altafan

altafan: Yes

altafan: cc @sekulicd

pkg/client-sdk/ark_sdk.go
```diff
@@ -0,0 +1,1413 @@
...
+	Receive(ctx context.Context) (string, string, error)
+	SendOnChain(ctx context.Context, receivers []Receiver) (string, error)
+	SendOffChain(
+		ctx context.Context, withExpiryCoinselect bool, receivers []Receiver,
+	) (string, error)
+	ForceRedeem(ctx context.Context) error
```
altafan: Let's rename this:
```suggestion
	UnilateralRedeem(ctx context.Context) error
```

pkg/client-sdk/wallet.go
altafan: would be better to change this method to:
```suggestion
	SignTransaction(tx string, inputs []uint32) (string, error)
```

This way we can pass either a psbt, pset or raw transaction and also we extend its definition to HD wallets as expecting to pass a single address might fit well for a single-key but not for other kind of wallets

altafan: cc @sekulicd

sekulicd:
@tiero @altafan client sdk is working and tested for grpc,rest [here](https://github.com/sekulicd/ark/blob/client-sdk/pkg/client-sdk/example/alice_to_bob.go) and wasm [here](https://github.com/sekulicd/ark/tree/client-sdk/pkg/client-sdk/example/wasm), bellow u can see wasm example which simulates sending of vtxo's from one wallet to another.

Screen Shot 2024-07-16 at 12 12 32

pkg/client-sdk/ark_sdk.go
sekulicd: Done

pkg/client-sdk/ark_sdk.go
sekulicd: Done

docker-compose.regtest.yml
```diff
@@ -26,13 +26,13 @@ services:
...
       - ARK_WALLET_ADDR=oceand:18000
-      - ARK_ROUND_INTERVAL=10
+      - ARK_ROUND_INTERVAL=30
       - ARK_NETWORK=regtest
       - ARK_LOG_LEVEL=5
       - ARK_ROUND_LIFETIME=512
       - ARK_DB_TYPE=sqlite
     ports:
-      - "6000:6000"
+      - "8080:6000"
```
altafan, line 35:
```suggestion
      - "6000:6000"
```

docker-compose.regtest.yml
```diff
@@ -26,13 +26,13 @@ services:
       - oceand
     environment:
       - ARK_WALLET_ADDR=oceand:18000
-      - ARK_ROUND_INTERVAL=10
+      - ARK_ROUND_INTERVAL=30
```
altafan:
```suggestion
      - ARK_ROUND_INTERVAL=10
```

common/Makefile
```diff
@@ -1,2 +1,2 @@
-test:
+test:
```
altafan, line 1:
```suggestion
test:
```

pkg/client-sdk/example/wasm/main.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"context"
+
+	arksdkwasm "github.com/ark-network/ark-sdk/wasm"
+)
+
+func main() {
+	var (
+		aspUrl = "http://localhost:8080"
```
altafan, line 14:
```suggestion
		aspUrl = "http://localhost:6000"
```

PR #199, opened 2024-07-05
Fix errors on round finalization

### LINKED ISSUE #175: Do not persist empty rounds
The ASP is[ saving round](https://github.com/ark-network/ark/blob/a6e0b6f408c58055751e2cc23e705bedaac51374/server/internal/core/application/service.go#L351) even if the "no payments registered" error happened. We should not store empty rounds in DB

### LINKED ISSUE #194: Send offchain `error: round failed: failed to create pool tx: failed to add input: input already existing`
This error happens when the list of swept rounds passed to [this function](https://github.com/ark-network/ark/blob/master/server/internal/infrastructure/tx-builder/covenant/builder.go#L145-L147) is not empty and one or more of them have the ConnectorAddress field empty.

This situation leads the coinselection to select the same utxos for all the rounds that have no ConnectorAddress.

The error the user gets in response comes from pset package and means that the same utxo(s) is added as input to the tx. I take this opportunity to say also that giving this kind of error to the end user is wrong, we should wrap it around a more general message like "something went wrong, retry later"

altafan:
This contains the following fixes:
- SQLite round repo: exclude rounds with empty connector address in `GetSweptRounds()`, align with what done in other impls
- Store the current round in memory to prevent sourcing it from the db in the several stages
- Thanks to the fix above, prevent storing aborted rounds (failed because empty)
- Prevent passing to ocean an empty connector address

All the bug fixed with this caused a round not to be finalized, @louisinger please review.

Testable version at https://asp.arkdev.info 

Closes #175.
Closes #194.

### DIFF
--- a/server/internal/core/application/service.go
+++ b/server/internal/core/application/service.go
@@ -325,6 +325,7 @@ func (s *service) start() {
 
 func (s *service) startRound() {
 	round := domain.NewRound(dustAmount)
+	//nolint:all
 	round.StartRegistration()
 	s.currentRound = round
 
--- a/server/internal/core/application/service.go
+++ b/server/internal/core/application/service.go
@@ -79,6 +79,7 @@ type service struct {
 
 	trustedOnboardingScriptLock *sync.Mutex
 	trustedOnboardingScripts    map[string]*secp256k1.PublicKey
+	currentRound                *domain.Round
 }
 
 func NewService(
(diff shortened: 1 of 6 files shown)


tiero:
linting check fails, other than that utACK for me

louisinger (approved):

PR #207, opened 2024-07-19
Update client sdk

### LINKED ISSUE #206: client-sdk: add persisten storage option
sqlite? it should be able to be cross-compile being native go

altafan:
This contains a refactor of the abstractions of the client SDK.

Explanation of the new interfaces:
- [`store.Store`](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/store/store.go#L27-L31) represents the main db where the [configuration data](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/store/store.go#L15-L25) are stored. Replaces the old config store
- [`client.Client`](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/client/client.go#L29-L60) represents the client used based on transport protocol, either _grpc_ or _rest_. Replaces old inner client
- [`explorer.Explorer`](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/explorer/explorer.go#L16-L29) the old explorer interface has been moved to a dedicated folder and properly scaffolded to be ready to "host" the bitcoin implementation (very soon)
- [`wallet.Wallet`](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/wallet/wallet.go#L23-L39) represents the wallet as a generic multi-key wallet, with methods for deriving new address and get all the derived ones for example . This is an evolution of the old wallet interface. The singlekey wallet is now basically an extremely simplified version of a (non) multi-key wallet. The interface definition should be ok for a future bitcoin HD wallet (built on top of btcwallet). Just so you know, this interface might be subject to changes in the future when we implement an HD wallet.*

Now setting up the SDK client is as simple as selecting the desired store type (json file, inmemory currently supported) or providing a custom store implementation - we'll get back to this in a few.
[Example](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/example/alice_to_bob.go#L31-L33).
If the store is empty, it's necessary to [initialize the SDK client](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/example/alice_to_bob.go#L38-L47) - config data are persisted in store -, otherwise it is initialized with the data retrieved from the store, just like the CLI works to make a comparison.

The list of features of the SDK client has been maintained - beside some renaming - with the addition of 2 methods to Unlock/Lock the inner wallet. This way, it's not mandatory to always provide the password for the required methods if the wallet is already unlocked. It is something similar to an optional login/logout session.

In the WASM pkg, I added a [local-storage implementation](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/wasm/local_storage_store_wasm.go) of `store.Store` and made it consumed by the [SDK client](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/example/wasm/main.go#L21-L23) - with a fallback strategy just for safety.
I successfully executed the wasm example in the browser! 🥳 

Last but not lease, this adds unit tests for the implementations of the interfaces listed above - not all of them. The gh action has been updated to run unit tests for server and SDK modules.
In a later PR we will make the current alice-bob example a proper integration test. This is already huuuge, no need to make it even bigger.

* The singlekey wallet makes uses of a [`walletstore.WalletStore`](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/wallet/singlekey/store/store.go#L14-L18) that is an extension of `store.Store` in charge of persisting the [wallet encrypted data](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/wallet/singlekey/store/store.go#L8-L12). The json file impl of `store.Store` and `walletstore.Store` can write on the same file without risk of loss of information. This interface is needed only and exclusively by the singlekey wallet impl that needs a persistency layer. A future btcwallet impl will have all the db completely managed, so all this won't be needed at all.

Closes #206.

Please @louisinger @sekulicd  review.


common/network.go
```diff
@@ -10,12 +12,31 @@ var Liquid = Network{
 	Addr: "ark",
 }
 
-var TestNet = Network{
+var LiquidTestNet = Network{
+	Name: "liquidtestnet",
+	Addr: "tark",
```
tiero, line 17: `tlark` :D

altafan: i might agree but it's off topic 🤣

tiero:
Any comments? @sekulicd @louisinger

pkg/client-sdk/internal/utils/utils.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/ark-network/ark/common"
+	"github.com/ark-network/ark/common/tree"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
+	"github.com/vulpemventures/go-elements/address"
+	"github.com/vulpemventures/go-elements/network"
+	"github.com/vulpemventures/go-elements/payment"
+	"github.com/vulpemventures/go-elements/taproot"
+)
+
+func ComputeVtxoTaprootScript(
```
louisinger: Probably this could be move in `common/tree` pkg in order to, later, implement the bitcoin version in `common/bitcointree`.

pkg/client-sdk/example/alice_to_bob.go
```diff
@@ -10,46 +10,47 @@ import (
...
-		//grpcAspUrl = "localhost:8080"
-		restAspUrl = "http://localhost:8080"
-		//grpcProtocol = arksdk.Grpc
-		restProtocol = arksdk.Rest
-		ctx          = context.Background()
+var (
+	aspUrl      = "http://localhost:8080"
+	clientType  = "rest"
+	walletType  = wallet.SingleKeyWallet
+	network     = "liquidregtest"
```
louisinger: nit: let's use `common.LiquidRegtest.Name` ?

sekulicd:
I have some general comments regarding this PR:

Firstly, while there are some improvements such as the addition of a local storage option for the browser (wasm), the refactoring introduces significant changes compared to the existing state. It would be very helpful to have a clearer explanation of the intentions behind these changes and the benefits they bring. For instance, could you please elaborate on the benefit of refactoring inner_client and making its methods public?

My primary concern is the use of the factory pattern. It feels like a premature optimization aimed at anticipating future scenarios like different types of wallets (HD, or different types of stores). As mentioned in the PR description, the Wallet interface "might be subject to changes in the future."

In the factory methods, we are attempting to cover all cases within a single New method. For example, [here](https://github.com/ark-network/ark/blob/3de2ffbb189c2684804aa8bb6585b6b7fbdf1517/pkg/client-sdk/ark_sdk.go#L79), I would suggest adding an additional factory method NewWithCustomWalletStore instead. The current implementation of the [wallet factory method](https://github.com/ark-network/ark/blob/3de2ffbb189c2684804aa8bb6585b6b7fbdf1517/pkg/client-sdk/wallet/wallet.go#L43) appears complex.

The logic for creating a wallet based on the type of store could be simplified by using separate constructors instead of a switch-case within a single function. I assume the single factory method will eventually be used for both liquid and BTC wallets. In general, I prefer the previous version where interfaces for wallet and store were passed directly, avoiding the need for switch cases in factory methods.

Additionally, variadic functions can be error-prone, and I believe the current function signatures are not user-friendly for clients. In the previous version, wallet and config stores were separated, which I think is a better design. The wallet doesn't need all the configuration information; it primarily uses network, aspPubKey, and UnilateralExit.

Furthermore, some configuration fields that were removed in the previous version because they could always be fetched from the ASP using the GetInfo API seem to have been reintroduced. I don't see the need for storing these fields.

Regarding naming conventions, exposed structs should not start with the same name as their package. For example:

store.Store
client.Client
explorer.Explorer
wallet.Wallet
Lastly, there has been a lot of code movement without clear benefits, such as the changes to inner_client. Additionally, there are now utils in both pkg/utils.go and internal/utils, which could lead to confusion.

I hope these points help in refining the changes and ensuring the refactoring brings clear benefits while maintaining the simplicity and usability of the codebase.

client/client.go
```diff
@@ -86,6 +89,29 @@ func getClientFromState(ctx *cli.Context) (arkv1.ArkServiceClient, func(), error
 }
 
 func getClient(addr string) (arkv1.ArkServiceClient, func(), error) {
-	client, cleanFn, err := arkgrpcclient.New(addr)
-	return client.Service(), cleanFn, err
+	creds := insecure.NewCredentials()
+	port := 80
```
louisinger, line 93: this shouldn't be configurable? Especially for testing purposes where we run the daemon under 6000

pkg/client-sdk/client/client.go
```diff
@@ -0,0 +1,62 @@
...
+
+type Vtxo struct {
+	Amount    uint64
+	Txid      string
+	VOut      uint32
+	RoundTxid string
+	ExpiresAt *time.Time
+}
+
+type Client interface {
```
louisinger: I would abstract the types generated by protobuf because `arkv1` stubs are specific to covenant version

pkg/client-sdk/client/grpc/client.go
```diff
@@ -0,0 +1,290 @@
...
+	errAspUrlEmpty = errors.New("asp url is empty")
+)
+
+type grpcClient struct {
+	conn     *grpc.ClientConn
+	svc      arkv1.ArkServiceClient
+	eventsCh chan client.RoundEventChannel
+}
+
+func NewClient(args ...interface{}) (client.Client, error) {
```
louisinger: Personal preference for typed arguments, because here you can't get how to configure the client by looking at the signature

common/encoding.go
```diff
@@ -40,7 +40,7 @@ func DecodeAddress(
 	if err != nil {
 		return
 	}
-	if prefix != Liquid.Addr && prefix != TestNet.Addr {
+	if prefix != Liquid.Addr && prefix != LiquidTestNet.Addr {
```
louisinger: let's also check if the prefix is the regtest hrp?

pkg/client-sdk/ark_sdk.go
```diff
@@ -30,262 +34,240 @@ const (
...
-		network.Testnet.Name: "https://blockstream.info/liquidtestnet/api",
-		network.Regtest.Name: "http://localhost:3001",
-	}
-)
-
 type ArkClient interface {
-	Connect(ctx context.Context) error
-	Balance(ctx context.Context, computeExpiryDetails bool) (*BalanceResp, error)
-	Onboard(ctx context.Context, amount uint64) (string, error)
+	Store() store.Store
```
louisinger: any reason for the Ark developer to access the store ?

pkg/client-sdk/ark_sdk.go
```diff
@@ -603,16 +591,27 @@ func (a *arkClient) sendOnchain(receivers []Receiver) (string, error) {
 }
 
 func (a *arkClient) sendOffchain(
-	ctx context.Context, withExpiryCoinselect bool, receivers []Receiver,
+	ctx context.Context,
+	withExpiryCoinselect bool, receivers []Receiver, password string,
```
louisinger: From UX point of view, I think would be better to ask for password *after* the tx creation. Here it looks dangerous to give access to the wallet without any way to "verify" the tx/fees/payment infos what do you think ?

pkg/client-sdk/ark_sdk.go
```diff
@@ -437,19 +419,29 @@ func (a *arkClient) Onboard(
...
+	}); err != nil {
 		return "", err
 	}
 
 	return txid, nil
 }
 
-func (a *arkClient) sendOnchain(receivers []Receiver) (string, error) {
+func (a *arkClient) SendOnchain(
+	ctx context.Context, receivers []Receiver, password string,
```
louisinger: In the CLI we craft the tx then print it and then we ask the password. Looks safer from UX point of view to unlock *after* you check the transaction.

louisinger (commented):
I don't like the fact that the password concept appears in the Client interface. First of all it seems to be specific to the wallet implementation (other wallets could use a key to unlock, or ask for Hardware device etc...).
Also it reduces the interactivity possibilities with the user like asking password after printing data about the round/tx. 

Can't we move the password logic outside of the Client interface and make it customizable (like either unlock 1 time or ask every time we need the secret keys etc..) ?

tiero:
I do agree regarding having multiple factories instead a single one with lot of config options that can led to errors as we scale.

I do also agree that password should be abstracted as is not all the key signers will be protected by it (it can be a dedicated interface)

For the rest, can't speak much in terms of implementation details, but the current APIs are pretty user friendly (looking at the example)

altafan:
> For instance, could you please elaborate on the benefit of refactoring inner_client and making its methods public?

For development purpose, if somebody wants to develop a custom wallet he could make use of the rest or grpc or rest clients.
Also, there wasn't a real reason to handle rest or grpc in a dedicated struct with if/switch, they became 2 different impl of the same interface
This is the reason why we expose also the store, wallet, etc interfaces: we provide a our way of abstracting an ark client but if it doesn't fit somebody's need, he can make use of the "inner components" to build his own ark client

> My primary concern is the use of the factory pattern. It feels like a premature optimization aimed at anticipating future scenarios like different types of wallets (HD, or different types of stores). As mentioned in the PR description, the Wallet interface "might be subject to changes in the future."

Not so premature since this abstraction comew after writing the actual cli, the bitcoin cli coming with the related pr, and the definition of other external services.

I said that the wallet interface might be subject to changes because i'm not 100% sure if it's ok for a later hd implementation. The current one is taylored for a single-key impl tho and 100% is not ok for the hd impl.

> NewWithCustomWalletStore instead. The current implementation of the wallet factory method appears complex.

Agreed, seems much cleaner

> The logic for creating a wallet based on the type of store could be simplified by using separate constructors instead of a switch-case within a single function. I assume the single factory method will eventually be used for both liquid and BTC wallets. In general, I prefer the previous version where interfaces for wallet and store were passed directly, avoiding the need for switch cases in factory methods.

That was the original idea but i had to add this factory to avoid circular dependencies

> Additionally, variadic functions can be error-prone, and I believe the current function signatures are not user-friendly for clients. In the previous version, wallet and config stores were separated, which I think is a better design. The wallet doesn't need all the configuration information; it primarily uses network, aspPubKey, and UnilateralExit.

Can you make a list of the funca you consider error-prone? Better to review them together

Wallet and store are still separated! Wallet must take care of persisting only the keys, while store is meant to persist asp and config data.
I think the separation of concerns is more clear now than before.
Furthermore, Walletstore is the persistency layer for the singlekey wallet only.
An hd wallet like btcwallet doesn't need and won't use this inteface, it comes with a fully managed persistency layer for the keys.
Walletstore's exception is that it's meant to persist data in the same place where store does, exactly like we currently do in the actual cli. Take it as a poethic license :) but remember that at some point the singlekey wallet will be used only for dev purposes, so i think it's fine to take this license on a non-prod tool
The hd wallet will need a store, or either the store data becuase it need to acccess info like asp pubkey for things like address derivation. Or maybe even better to pass these required data as func args noe that i thing about it, i'll reson on this. I'll try to find the time to sync offline in the next days.

> Furthermore, some configuration fields that were removed in the previous version because they could always be fetched from the ASP using the GetInfo API seem to have been reintroduced. I don't see the need for storing these fields.

Can you make a list of such fields? Consider just that some info like the asp pubkey are static so it's fine to fetch them only once. In a second moment would be nice to add cache invalidation in the ark client so we are sure the stored info are always up-to-date, but too soon now to reason about this.

> Regarding naming conventions, exposed structs should not start with the same name as their package. 

Do you agree on adding a `Service` suffix to those interfaces?

altafan:
> Can't we move the password logic outside of the Client interface and make it customizable (like either unlock 1 time or ask every time we need the secret keys etc..) ?

It should be already possible by making use of the client's Unlock/Lock methods wherever you need and pass an empty string as password in send because you already unlocked the wallet by yourself. If you don't need such "custom" scenario you can just let the send method to handle the unlock.

Basically you can choose whether managing the authenticated session or let it be managed by the client itself.

cc @louisinger

altafan:
> I do also agree that password should be abstracted as is not all the key signers will be protected by it (it can be a dedicated interface)

I agree but i thought it was too early for this as i don't have a clear vision of such component.
These lock/unlock methods are an initial step for that final goal.
Consider that the current ark client is unauthenticated

cc @tiero

louisinger:
> It should be already possible by making use of the client's Unlock/Lock methods wherever you need and pass an empty string as password in send because you already unlocked the wallet by yourself. If you don't need such "custom" scenario you can just let the send method to handle the unlock.

Most of "external" wallets like Hardware devices or browser extensions handle the authentication independently. So I'd revert this change if it's not necessary to have it. Let's assume the `Wallet` interface is handling the lock/unlock process when it's necessary ?

sekulicd:
> Can you make a list of such fields?

We use aspPubKey, network, round lifetime, unilateral exit
```
message GetInfoResponse {
  string pubkey = 1;
  int64 round_lifetime = 2;
  int64 unilateral_exit_delay = 3;
  int64 round_interval = 4;
  string network = 5;
  int64 min_relay_fee = 6;
}
```

> Do you agree on adding a Service suffix to those interfaces?
For store i would do similar to prev stare store.Config store.Wallet interfaces, for client maybe transport.ArkClient or some combo of these words, for rest yeah i guess better to choose Service over same word for pkg and struct/interface.

> Can you make a list of the funca you consider error-prone? Better to review them together

Its related to this factory pattern, that, i can see, u like to use :), for eg [here](https://github.com/ark-network/ark/blob/3de2ffbb189c2684804aa8bb6585b6b7fbdf1517/pkg/client-sdk/wallet/wallet.go#L43).

tiero:
FWIW check BDK who just release his v1 with extensive refactor after years of adoption among Bitcoin dev community

https://x.com/bitcoindevkit/status/1815412036723171501?s=46

pkg/client-sdk/internal/utils/utils.go
altafan: Better to do this in #214 where we have the bitcoin impl of the method? This pr would be merged before it ideally

pkg/client-sdk/ark_sdk.go
altafan: not sure the following is a strong enough reason:

this is used only by the wasm wrapper that exposes some ["extra methods"](https://github.com/ark-network/ark/blob/update-client-sdk/pkg/client-sdk/wasm/ark_sdk_wasm.go#L28-L36) compared to this interface for accessing asp info.

We should define the same methods here so they are consumed instead of accessing the store. Ideally, it makes much more sense to expose the wallet rather than the store but can be done in a second moment if needed.

Nice catch, I'll get rid of this in favour of getters for asp info

client/client.go
altafan: I simply reverted the changes. L#99 handles what you pointed out.

pkg/client-sdk/client/client.go
altafan: this is the covenant client and will be moved to client/covenant/client.go once we work on the covenant-less version. We can keep it like this

sekulicd (approved):

Issue #213, opened 2024-07-21
Make the bitcoin wallet restorable


louisinger:
Add a way to restore the embedded wallet (read/write mnemonic & accounts)

Issue #212, opened 2024-07-21
Add clArk clients tree signing round


louisinger:
The clArk clients should be able to sign the vtxo tree in order to make the service trustless.

Issue #211, opened 2024-07-21
Bitcoin wallet password


louisinger:
The `btc-embedded` package implements a bitcoin wallet for clArk. This wallet makes use of lnd's `btcwallet` interface. The secret keys of the wallet are encrypted and can be unlocked by a password. This password should be configurable.

Issue #208, opened 2024-07-19
Dynamic min relay fee


altafan:
We need to change the behavior of the SDK client and make it fetch the min relay fee whenever required instead of expecting it to be fixed (ie. source it from the persistency layer).

PR #214, opened 2024-07-21
Implement covenant-less `WalletInterface`

### LINKED ISSUE #182: Bitcoin wallet
Let's implement a bitcoin implementation of [`ports.WalletService`](https://github.com/ark-network/ark/blob/master/server/internal/core/ports/wallet.go#L9) as single-key wallet.

Actually 2 keys, one for the ark account and one for the ark-connector account.

As a single-key wallet, we don't need to persist utxos/txs as it's affordable to restore them at startup and not slow down the bitcoin version's development.

louisinger:
This PR implements a Bitcoin wallet aiming to own and spend funds of cl-Ark ASPs. The wallet is using [lnd/btcwallet](https://pkg.go.dev/github.com/lightningnetwork/lnd@v0.18.2-beta/lnwallet/btcwallet) pkg.

TL;DR

- `btc-embedded` implements `WalletInterface` and `BlockchainScanner` using btcwallet pkg backed by a neutrino instance.
- `application/covenantless.go` file is a dummy clArk application service letting to run clArk. Dummy because only the ASP is signing the congestion tree.
- several changes in `config`, add new variables to configure clARK. 
- add `docker-compose.clark.regtest.yml` file and the associated e2e tests.
- clArk CLI (`client/covenantless`)

### TODOs

_This PR makes clArk possible on regtest but needs extra work to be done before a real use on mainchain_

- #211 
- #208 
- #212
- #213
- #215 

it closes #182 

@sekulicd @tiero @altafan @bordalix please review


server/internal/core/application/covenantless.go
```diff
@@ -0,0 +1,1004 @@
...
+		PubKey:              pubkey,
+		RoundLifetime:       s.roundLifetime,
+		UnilateralExitDelay: s.unilateralExitDelay,
+		RoundInterval:       s.roundInterval,
+		Network:             s.network.Name,
+		MinRelayFee:         int64(s.minRelayFee),
+	}, nil
+}
+
+// TODO clArk changes the onboard flow (2 rounds ?)
```
tiero, line 203: if we use the simple "boarding wihtout luggage" its one-step boarding, user sends to address.

```
- A + S 
- A after 3 months 
```

common/bitcointree/psbt.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"bytes"
+
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
+)
+
+var (
+	COSIGNER_PSBT_KEY_PREFIX = []byte("cosigner")
```
tiero, line 11: whats this for?

louisinger: It lets you write the cosigners public keys in the PSBTs. These keys have been used to create the aggregated musig2 key signing the tree txs. All cosigners need them to sign the tree. 

Note that now, the cosigners keys are only owned by the ASP so later, when we'll add client-side signing, we would need to add extra data (the aggregated nonce) but the idea is to use the PSBT as transport layer for musig2 fields

server/internal/core/application/covenantless.go
louisinger: hm yes but Alice and ASP need an extra round to cosign together the tree no ?

tiero:
README should be updated with nend to end testing of the ASP on Bitcoin network (regtest or signet is fine)

PR #216, opened 2024-07-24
Payment Charge API

### LINKED ISSUE #183: new payments API
Let's add the APIs for the shortcut so users don't need to join rounds to make off-chain payments.

Instead, receivers will settle their "shortcut vtxos" as soon as they get back online and fetch them from the explorer (ASP).

This way the users would need to join rounds only to (collaboratively) exit or to send to themselves.

louisinger:
This PR implements async payment known as "Somsen shortcut". 

docs : https://arkdev.info/docs/learn/payments

it depends on #214 

TL;DR

- `domain.AsyncPaymentTxs` is a new struct embedding the "shortcut tx" and the unconditionnal tx related to an async payment. The struct can be attached to a `domain.Vtxo` struct.
- add 2 new RPCs `CreateAsyncPayment` & `CompleteAsyncPayment`.
- add a new method in `ports.TxBuilder` crafting the async payment transactions.

### User flow (Alice --- async pay w/ ASP ---> Bob)

Async payment introduces 2 new transactions:
* the "redeem tx" spends the VTXO outpoint and send it to a "shortcut" script (A+S or B+S or A in x weeks).
* the "unconditional forfeit tx" spends the VTXO outpoint and send it to the ASP wallet.

The flow is the following:

1. Alice uses `CreateAsyncPayment` RPC to signal the ASP she wants to spend a vtxo in a async way.
2. ASP creates the redeem and the unconditional forfeit transactions. He signs *only* the redeem one.
3. Alice receives the transactions, verifies them, and signs *both transactions*.
4. Alice sends the signed PSBTs with the `CompleteAsyncPayment` RPC.
5. ASP verifies the signatures and creates the new vtxo in DB.
6. Bob received the new VTXO, he can use a normal send to settle the VTXO in a round (do a self-send).

`client/covenantless/async_payment.go` implements this flow.

### TODOs

> This PR is a draft and should not be merged as it is.

* [ ] add locktimes to transactions in `TxBuilder.BuildAsyncPaymentTransactions`
* [ ] handle "prevent getting cheated" scenario for async vtxos
* [ ] handle change output in "redeem tx"
* [ ] add `sender & ASP` closure to the shortcut script (output script of redeem tx) + the related flow

it closes #183 

@tiero @sekulicd @bordalix please review


tiero:
We need a CancelAsyncPayment RPC

stevenroose:
This is the "out of round" payments, right? "async payments" are a thing in lightning, might be worth it to try not overload the term. Lightning's async payments are actually pretty useful for doing LN receives in Ark.

altafan:
@stevenroose Thank you for your feedback and for pointing out the potential naming overlap with Lightning's async payments.

To clarify, this PR is about the shortcut and it's still a draft. We chose "Async Payment" as a placeholder name because we felt it was descriptive of the feature's behavior from the user's perspective.
However, we understand your concern about potential confusion with Lightning's async payments. While we believe our use of "async" accurately describes the functionality, we acknowledge the importance of avoiding terminology conflicts in the space.

We're open to considering alternative names that maintain clarity about the feature's purpose and benefits. That said, options like "Shortcut Payment" don't fully capture the user benefit we're trying to convey.

We'll revisit the naming with your feedback in mind. If you have any specific suggestions that you feel would be both accurate and distinct, we'd be interested in hearing them.

Thank you for bringing this to our attention and helping us refine our terminology.

server/api-spec/protobuf/ark/v1/service.proto
```diff
@@ -66,8 +66,39 @@ service ArkService {
...
+    option (google.api.http) = {
+      post: "/v1/payment/async/complete"
+      body: "*"
+    };
+  }
+}
+
+message CreateAsyncPaymentRequest {
+  Input input = 1;
+  string receiver_pubkey = 2;
```
tiero, line 87:
```suggestion
  string address = 2;
```

server/api-spec/protobuf/ark/v1/service.proto
```diff
@@ -66,8 +66,39 @@ service ArkService {
...
+}
```
tiero, line 88:
```suggestion
uint64 amount = 3;
}
```

PR #224, opened 2024-07-31
CLI with ARK Sdk

### LINKED ISSUE #218: [CLI] Use sdk
Now that we have the first version of the client SDK we can use it within the covenant/covenant-less CLIs and simplify a lot the code of the actions

sekulicd:
This updates CLI so that it uses arksdk.
Note that this depends on arksdk covenantless update, once arksdk is updated changes should be simple and applicable in [here](https://github.com/sekulicd/ark/blob/72a649ffb4c055ba53186de249d9f3213d176fee/client/main.go#L265)

This closes #218


client/main.go
```diff
@@ -170,54 +228,312 @@ func main() {
...
+func config(ctx *cli.Context) error {
+	cfg, err := arkSdkClient.GetConfigData(cntx)
+	if err != nil {
+		return err
+	}
+
+	return printJSON(cfg)
+}
+
+func onboard(ctx *cli.Context) error {
```
altafan: this can be dropped once the pr is synced with master

client/main.go
```diff
@@ -170,54 +228,312 @@ func main() {
...
+		return err
+	}
+
+	return printJSON(map[string]interface{}{
+		"offchain_address": offchainAddr,
+		"onchain_address":  onchainAddr,
+	})
+}
+
+func redeem(ctx *cli.Context) error {
```
altafan, line 361: if `--force` flag is set we must call `arkSdkClient.UnilateralRedeem`

client/main.go
```diff
@@ -170,54 +228,312 @@ func main() {
...
+func receive(ctx *cli.Context) error {
+	offchainAddr, onchainAddr, err := arkSdkClient.Receive(cntx)
+	if err != nil {
+		return err
+	}
+
+	return printJSON(map[string]interface{}{
+		"offchain_address": offchainAddr,
+		"onchain_address":  onchainAddr,
+	})
```
altafan: `Receive` now returns the boarding address as onchain address:
```suggestion
	offchainAddr, boardingAddr, err := arkSdkClient.Receive(cntx)
	if err != nil {
		return err
	}

	return printJSON(map[string]interface{}{
		"offchain_address": offchainAddr,
		"boarding_address":  boardingAddr,
	})
```

client/main.go
```diff
@@ -1,232 +1,525 @@
...
+			return err
 		}
+		return printJSON(map[string]interface{}{"txid": txID})
+	}
+
+	return nil
+}
 
-		path = strings.Replace(path, "~", homeDir, 1)
+func decrypt(encrypted, password []byte) ([]byte, error) {
```
louisinger, line 506: `decrypt` and `encrypt` are also implemented in the SDK package. Also they are used only to print the private key. Should we consider exporting the functions or maybe implement `dumpPrivateKey` in sdk ? cc @altafan @sekulicd

louisinger (commented):
Integration tests are not running in CI ? Is it expected ?

sekulicd:
> Integration tests are not running in CI ? Is it expected ?

Looks like they are triggered only if there are changes in /server which is wrong.

client/main.go
```diff
@@ -278,6 +280,13 @@ func receive(ctx *cli.Context) error {
 }
 
 func claim(ctx *cli.Context) error {
+	if err := arkSdkClient.Unlock(
```
tiero, line 283: wouldnt be more correct to check is IsUnlocked before trying to unlock an already unlocked SDK? to avoid extra roundtrip

client/main.go
```diff
@@ -59,6 +59,7 @@ func main() {
 			return fmt.Errorf("error initializing ark sdk client: %v", err)
 		}
 		arkSdkClient = sdk
+
```
tiero, line 62: Are we sure in using a global variable here? Wouldnt be better to have it inside the context?

client/main.go
sekulicd: This is the only way, AFAIK, that SDK loads primary key from memory, in the [example](https://github.com/sekulicd/ark/blob/52b21c5e687111250b365e89438503e2327ce107/pkg/client-sdk/example/covenant/alice_to_bob.go#L34) itself similar flow is used.

client/main.go
sekulicd: Dont see an issue with this, its one file, all function calls, purpose is just to avoid code repetition by setting up client in each func, so basically before each cmd call setup arksdk client and set global var to func.

pkg/client-sdk/types.go
```diff
@@ -19,13 +20,16 @@ var (
 		GrpcClient: grpcclient.NewClient,
 		RestClient: restclient.NewClient,
 	}
+
+	liquidRegTestUrl  = getEnv("LIQUID_REGTEST_URL", "http://localhost:3001")
+	bitcoinRegTestUrl = getEnv("BITCOIN_REGTEST_URL", "http://localhost:3000")
```
tiero, line 25: shouldn't we use here viper/urfave directly? Usually ENV are automatically `ARK` prefixed

client/main.go
tiero: The SDK exposes `IsLocked` https://github.com/sekulicd/ark/blob/52b21c5e687111250b365e89438503e2327ce107/pkg/client-sdk/wallet/wallet.go#L21

client/main.go
tiero: The SDK exposes IsLocked https://github.com/sekulicd/ark/blob/52b21c5e687111250b365e89438503e2327ce107/pkg/client-sdk/wallet/wallet.go#L21

pkg/client-sdk/types.go
sekulicd: wanted to use quicker solution without 3rd party pkg's since this is only relevant for regtest.

tiero: You can use EnvVars of `cli` when creating flags, so you have both a flag and an env var https://github.com/sekulicd/ark/blob/52b21c5e687111250b365e89438503e2327ce107/client/main.go#L79

sekulicd: yeah, but var supportedNetworks is [inside](https://github.com/sekulicd/ark/blob/52b21c5e687111250b365e89438503e2327ce107/pkg/client-sdk/types.go#L26) sdk, imo this is simplest way to inject diff values for regtest without touching existing arksdk interface, otherwise we would need to pass network and this is not good since network comes from ASP via GetInfo. Instead of passing it via CLI env var i decided to pass it throuugh regtest dockerfile's.

client/main.go
sekulicd: Yeah, but that wont help, when u init arksdk, pk is not loaded by default, and only place where pk is set is in [Unlock](https://github.com/sekulicd/ark/blob/52b21c5e687111250b365e89438503e2327ce107/pkg/client-sdk/wallet/singlekey/wallet.go#L112) method.
Calling IsLocked is meaningless in this case.

pkg/client-sdk/types.go
tiero: Also..we used to have an --explorer flag!!

Look at the current doc https://arkdev.info/docs/quick-start/client/#setting-up-alice-and-bobs-clients

sekulicd: This is not relevant after sdk introduction, as there is no way to pass explorer url when creating instance, as said now explorer is created based on net retreived from ASP with GetInfo RPC.

tiero: This is for sure wrong on the SDK side then. Let's fix it! 

The explorer URL SHOULD NOT be mandated by the connected ASP, as is used to fetch blockchain data which is an open protocol. (I guess to check onchain balance? Compute timestamps from blocks?). 

It's very likely user of the SDK wants to use their explorer, instead being us forcing them with some well-known explorer URLs.

PR #225, opened 2024-08-02
Covenant-less ark sdk

altafan:
This adds support for the covenant-less version of the ark SDK:

* adds bitcoin impl of single key wallet: instead of having the impl in 2 different folders, they are under the same [folder](../blob/bitcoin-sdk/pkg/client-sdk/wallet/singlekey). `singleKeywallet` exposes the common functions (wallet creation, locking/unlocking), while `liquidWallet` and `bitcoinWallet` expose those depending on the chain (address generation, tx signing).
* adds bitcoin explorer: the current interfaces and impl are already enough to support the bitcoin version, so the latter has been moved from _explorer/liquid/_ to _explorer/_. It's enough to init the explorer with the right esplora base URL and network
* refactors client: stuff related to redeem branches has been moved to _internal/utils/redemption/_ (to prevent circular dep with explorer). Maybe worth moving to common? Also, as @louisinger  [suggested](https://github.com/ark-network/ark/pull/207#discussion_r1686351311)  in the prev PR, this frees the client interface to depend on proto stubs. The interface has been simplified a bit by removing unnecessary or redundant methods. 
* adds covenant-less ark client: like done for the wallet, instead of having the impls in 2 different folders, they are defined as extensions of the common [ark SDK client](../blob/bitcoin-sdk/pkg/client-sdk/client.go) (`arkClient`) that exposes all the common methods (init, lock/unlock, address generation). The `covenantArkClient` and `covenantlessArkClient`, instead, expose methods depending on the chain and script type (onboard, send, redeem, get balance)
* all utils func spreaded in the module have been merged in _internal/utils/_
* adds tests and examples like done for covenant


pkg/client-sdk/client.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/ark-network/ark-sdk/wallet"
+	singlekeywallet "github.com/ark-network/ark-sdk/wallet/singlekey"
+	walletstore "github.com/ark-network/ark-sdk/wallet/singlekey/store"
+	filestore "github.com/ark-network/ark-sdk/wallet/singlekey/store/file"
+	inmemorystore "github.com/ark-network/ark-sdk/wallet/singlekey/store/inmemory"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
+)
+
+const (
+	DUST = 450
```
tiero, line 23: Is it for Bitcoin also?

altafan: yes, even tho it could be a little bit lower than 450 sats based on [this](https://bitcoin.stackexchange.com/a/41082/145037), for now it's fine for sake of simplicity

tiero: Use the right Dust Amount for Bitcoin

PR #232, opened 2024-08-09
Support macaroons and TLS && Add arkd wallet cmds

altafan:
This adds support for macaroons auth for the admin and wallet interfaces (replaces the previous basic auth-based one) and TLS encryption with a self-signed key pair.

At startup the server generates the default [macaroon files](https://github.com/ark-network/ark/blob/admin-cli/server/internal/interface/grpc/macaroons.go#L15-L18) within the `macaroons/` folder inside the datadir if needed.

Similarly, it also generated the default [tls keypair](https://github.com/ark-network/ark/blob/admin-cli/server/internal/interface/grpc/service.go#L35-L36) into the `tls/` folder inside the datadir if needed.

This also adds a set of commands to the `arkd` binary to interact with the internal wallet (create/restore, lock/unlock, get status, address and balance)

**NOTE:** I found out restoring a bitcoin wallet doesn't work, left a TODO for later to fix.


server/cmd/arkd/main.go
```diff
@@ -18,19 +22,40 @@ var (
...
+		Value: fmt.Sprintf("http://localhost:%d", config.DefaultPort),
+	}
+	macaroonFlag = &cli.StringFlag{
+		Name:  "macaroon-path",
+		Usage: "the path where to find the admin macaroon file",
+		Value: filepath.Join(common.AppDataDir("arkd", false), "macaroons", "admin.macaroon"),
+	}
+	tlsCertFlag = &cli.StringFlag{
+		Name:  "tls-cert-path",
+		Usage: "the path where to find the TLS certificate",
```
tiero, line 44: Can we make it so if not passed it assume its' unsecure? (ideally because we have a proxy that does TLS termination already ie. coolify)

server/pkg/kvdb/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (C) 2015-2022 Lightning Labs and The Lightning Network Developers
```
tiero, line 1: I assume was hard to use badger or sqlite directly?

server/internal/interface/grpc/permissions/.gitkeep
tiero, line 1: do we need this

server/cmd/arkd/main.go
altafan: Certainly, and it's already supported! If the URL starts with `http://` this flag is just ignored.

In general, the CLI understands by itself if the tls cert is required or not depending on the given url

server/pkg/kvdb/LICENSE
altafan: This package is a copy/paste of the [lnd one](https://github.com/lightningnetwork/lnd/tree/master/kvdb).

I simplified it by removing some impl, they already have the sqlite one, therefore it should be possible to use that kind of db. I'm gonna apply the changes

server/internal/interface/grpc/permissions/.gitkeep
altafan: it's been removed

tiero:
test fails, but LGTM other than that! utACK

server/pkg/kvdb/LICENSE
altafan: the sqlite db depends on [sqldb](https://github.com/lightningnetwork/lnd/tree/master/sqldb) module which is a tailored db for the lnd node (check inside the [sqlc folder](https://github.com/lightningnetwork/lnd/tree/master/sqldb/sqlc/queries)). 

We'll add other macaroon dbs later

tiero:
remove test for trusted onboiarding

```
Incorrect Usage: flag provided but not defined: -trusted
```

tiero (approved):

PR #266, opened 2024-08-19
IsInitialized > walletLoaded & panic recovery

### LINKED ISSUE #253: arkd wallet status: return initialized = true if wallet db exists
If you run an Arkd instance ie. `make run` and create the wallet, if you stop the binary it will not go load the wallet db from datadir.

sekulicd:
This related to closed [PR](https://github.com/ark-network/ark/pull/264) and updates few things:
1. Original isInitialized is renamed to walletLoaded which holds same code as 
2. New isInitialized is added that checks if walled db exists 
3. isInitialized is used in Status RPC
4. Prev PR avoids panics but it doesnt add general panic handler 

This closes #253 

@tiero @louisinger please review.


tiero:
utACK

louisinger (approved):
Nice ! `recover` is a thing, never used it, powerful !

PR #268, opened 2024-08-19
check go work sync in unit test

tiero:
`go work sync` runs `tidy` under the hood on each module. The current master doesnt has the last version. I have introduced a gh action to automatically check this during unit tests


sekulicd (approved):
LGTM

PR #271, opened 2024-08-21
Make the round participants sign the vtxo tree

### LINKED ISSUE #184: [Bitcoin] Add tree signing stage APIs
Requires #183 

Since users now join rounds exclusively to send to themselves, let's add the "tree signing" stage before the "forfeit signing" one.

Instead of adding APIs to collect signers pubkeys, we can gather them at payment registration, as the senders and receivers are the same identity

### LINKED ISSUE #212: Add clArk clients tree signing round

louisinger:
_This PR adds APIs allowing to sign the vtxo tree before the round finalization_

TL;DR

- add new APIs to send musig2 nonces and signatures of tree's txs.
- add new Events returning the unsigned tree and the aggregated nonces.
- implement musig2 coordinator in application service.
- rework `PingResponse` in order to return all type of round events.
- update the `client` package.
- update the `sdk/client-sdk` package.
- several fixes and improvements in `common/musig2.go`

### Round interval

The ASP is dividing the round finalization into three parts:
1. nonces exchange step (ASP sends public key and receive nonces from signers). Once everybody generated the nonces, the ASP combine them and send the result to signers via event.
2. signing step: ASP collects signatures from signers, then aggregate and finalize tree txs.
3. legacy finalization: sign forfeits, verify tree etc...

image

Thus, the time to sign forfeits (and the tree) is divided by 3. It works well on a local device using the default round interval (10 seconds) but it may be possible this interval is too short for a remote ark running rounds involving a lot of signers.

it closes #212 
it closes #184 

@sekulicd @bordalix @tiero please review


api-spec/protobuf/ark/v1/service.proto
```diff
@@ -17,13 +17,24 @@ service ArkService {
...
+      post: "/v1/payment/tree/nonces"
+      body: "*"
+    };
+  }
+  rpc SendTreeSignatures(SendTreeSignaturesRequest) returns (SendTreeSignaturesResponse) {
+    option (google.api.http) = {
+      post: "/v1/payment/tree/signatures"
+      body: "*"
+    };
+  }
```
tiero, line 31: Some new proposal for renaming current proto has been made in https://github.com/ark-network/ark/pull/228/files

We could call these Submit..

client/covenantless/claim.go
```diff
@@ -69,8 +72,19 @@ func selfTransferAllPendingPayments(
 		return err
 	}
 
+	ephemeralKey, err := secp256k1.GeneratePrivateKey()
```
tiero, line 75: why do we use a random key here?

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -91,6 +102,7 @@ message CompletePaymentResponse {}
 
 message RegisterPaymentRequest {
   repeated Input inputs = 1;
+  optional string ephemeral_pubkey = 2;
```
tiero, line 105: is this the best name we can think of? What the user has to know amount this to register a payment?

pkg/client-sdk/client/client.go
```diff
@@ -43,6 +45,12 @@ type ASPClient interface {
 	CompletePayment(
 		ctx context.Context, signedRedeemTx string, signedUnconditionalForfeitTxs []string,
 	) error
+	SendTreeNonces(
+		ctx context.Context, roundID, cosignerPubkey string, nonces bitcointree.TreeNonces,
+	) error
+	SendTreeSignatures(
+		ctx context.Context, roundID, cosignerPubkey string, signatures bitcointree.TreePartialSigs,
+	) error
 	Close()
```
tiero, line 54: Can we do bundle these two in a single RPC?

server/internal/core/application/covenant.go
```diff
@@ -50,7 +52,7 @@ func NewCovenantService(
 	builder ports.TxBuilder, scanner ports.BlockchainScanner,
 	scheduler ports.SchedulerService,
 ) (Service, error) {
-	eventsCh := make(chan domain.RoundEvent)
+	eventsCh := make(chan interface{})
```
tiero: 🤔

pkg/client-sdk/client/client.go
louisinger: No, musig2 **must** be 2 rounds as far as I get it

server/internal/core/application/covenant.go
louisinger: I know it looks weird, but if you check how we are using this channel, you will see that we are not using the interface at all. The idea behind this is to extend the event type without modifying the `domain`. 
The new musig2 events are not `domain.RoundEvent` because **they are not modifying the Round state**.

This `eventsCh` is systematically red by a type checking `switch` on each event structs. The idea was to make possible to send any type of events (not only domain ones).

api-spec/protobuf/ark/v1/service.proto
louisinger: ephemeral highlights the fact that the user should delete the key after signing the tree. Open to suggestions

common/bitcointree/musig2.go
```diff
@@ -12,20 +14,45 @@ import (
...
 	ErrAggregateKeyNotSet   = errors.New("aggregate key not set")
+
+	columnSeparator = byte('|')
+	rowSeparator    = byte('/')
 )
 
-type TreeNonces [][][66]byte // public nonces
+type Musig2Nonce struct {
+	PubNonce [66]byte
+}
```
altafan, line 30: no need for struct?
```suggestion
type Musig2Nonce [66]byte
```

common/bitcointree/musig2.go
```diff
@@ -507,3 +523,71 @@ type treePrevOutFetcher struct {
...
+type writable interface {
+	Encode(w io.Writer) error
+}
+
+type readable interface {
+	Decode(r io.Reader) error
+}
+
+// encodeMatrix encode a matrix of serializable objects into a byte stream
+func encodeMatrix[T writable](matrix [][]T) ([]byte, error) {
```
altafan, line 536:
```suggestion
func encodeMatrix(matrix [][]writable) ([]byte, error) {
```

altafan: no need for generics?

server/internal/core/application/covenantless.go
```diff
@@ -38,12 +39,14 @@ type covenantlessService struct {
 	paymentRequests *paymentsMap
 	forfeitTxs      *forfeitTxsMap
 
-	eventsCh     chan domain.RoundEvent
+	eventsCh     chan interface{}
```
altafan: I completely understand why the new "events" are not at the domain but at the application level since they contain ephemeral data that is unnecessary to persist.

That being said, if we define a `isEvent()` method for the new events we shouldn't need to make this change and we could keep this field as `chan domain.RoundEvent` and prevent other changes in the rest of the codebase

server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -145,11 +201,36 @@ func (h *handler) RegisterPayment(ctx context.Context, req *arkv1.RegisterPaymen
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	var cosigner *secp256k1.PublicKey
+
+	if h.svc.IsCovenantLess() {
```
altafan, line 206: this is not really needed. We can just parse and the key if it's not nil, then at app level uses or ignores is depending on its version.

server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -145,11 +201,36 @@ func (h *handler) RegisterPayment(ctx context.Context, req *arkv1.RegisterPaymen
...
+		buf, err := hex.DecodeString(pubkey)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, "invalid cosigner pubkey (expect hex)")
+		}
+
+		cosigner, err = secp256k1.ParsePubKey(buf)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, "invalid cosigner pubkey")
+		}
+	}
```
altafan: this is not really needed. We can just parse and the key if it's not nil, then at app level uses or ignores is depending on its version (the `missing cosigner pubkey` error message would be returned by the app svc):
```suggestion
	if pubkey := req.GetEphemeralPubkey(); len(pubkey) > 0 {
		buf, err := hex.DecodeString(pubkey)
		if err != nil {
			return nil, status.Error(codes.InvalidArgument, "invalid cosigner pubkey (expect hex)")
		}

		cosigner, err = secp256k1.ParsePubKey(buf)
		if err != nil {
			return nil, status.Error(codes.InvalidArgument, "invalid cosigner pubkey")
		}
	}
```

server/internal/core/application/covenant.go
altafan: https://github.com/ark-network/ark/pull/271/files#r1734483986

server/internal/core/application/covenantless.go
```diff
@@ -424,54 +489,60 @@ func (s *covenantlessService) startFinalization() {
 	if num > paymentsThreshold {
 		num = paymentsThreshold
 	}
-	payments := s.paymentRequests.pop(num)
+	payments, cosigners := s.paymentRequests.pop(num)
+	if len(payments) > len(cosigners) {
+		roundAborted = true
```
altafan: `roundAborted` is meant to be true ONLY if the round was aborted because there were no payments in the [queue of requests](https://github.com/ark-network/ark/blob/master/server/internal/core/application/covenantless.go#L418).

In any other case we want to save the failing event and start a new round, as [already handled](https://github.com/ark-network/ark/blob/master/server/internal/core/application/covenantless.go#L399-L406).

Please, remove all the `roundAborted = true` lines you added

louisinger: ok my bad I didn't get it
I'd add a comment explaining this, do u agree?

server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -312,6 +385,84 @@ func (h *handler) GetInfo(ctx context.Context, req *arkv1.GetInfoRequest) (*arkv
...
+
+	cosignerPublicKey, err := secp256k1.ParsePubKey(pubkeyBytes)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid cosigner public key")
+	}
+
+	nonces, err := bitcointree.DecodeNonces(hex.NewDecoder(strings.NewReader(serializedNonces)))
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid tree nonces")
+	}
```
altafan: move to app svc

server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -312,6 +385,84 @@ func (h *handler) GetInfo(ctx context.Context, req *arkv1.GetInfoRequest) (*arkv
...
+
+	cosignerPublicKey, err := secp256k1.ParsePubKey(pubkeyBytes)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid cosigner public key")
+	}
+
+	signatures, err := bitcointree.DecodeSignatures(hex.NewDecoder(strings.NewReader(serializedSignatures)))
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid tree signatures")
+	}
```
altafan: move to app svc

common/bitcointree/musig2.go
louisinger: We need it because the interface is not the argument itself. Instead, the argument is a slice matrix of the interface. I won't speculate why Golang does not accept this but if I make the change you suggest, I need to cast the matrix passed as arguments: "`TreeNonces` is not a `[][]writable`"

server/internal/core/application/utils.go
```diff
@@ -20,8 +21,9 @@ type timedPayment struct {
 }
 
 type paymentsMap struct {
-	lock     *sync.RWMutex
-	payments map[string]*timedPayment
+	lock          *sync.RWMutex
+	payments      map[string]*timedPayment
+	ephemeralKeys map[string]*secp256k1.PublicKey
```
altafan, line 26: either create a clone of `paymentsMap` but for pubkeys, or we add another lock here to control access to the ephemeralKeys map.

Using the same lock can lead to unnecessary and unexpected waiting time for a sender to register his pubkeys in a scenario where there are a lot of users registering their inputs at nearly the same time

altafan (approved):

PR #279, opened 2024-08-25
New boarding protocol 

### LINKED ISSUE #237: reverse boarding address

louisinger:
_This PR replaces all boarding protocols (covenant & covenantless) by a non-interative one.
the new boarding protocol is as simple as "send to boarding address". Once the fund is confirmed onchain, the user will co-sign it with the ASP while claiming the utxo in a round._

TL;DR
- add `REVERSE_BOARDING_EXIT_DELAY` config parameter.
- add `GetBoardingAddress` RPC returning a taproot address `Alice & ASP or Alice after x`.
- allow registering a "boarding utxo" in `RegisterPayment` RPC. Expect the utxo to be signed in `FinalizePayment` via "signedRoundTx" field.
- allow a Round to be finalized without forfeits transactions.
- make the connector output optional.
- add `common/descriptor` pkg allowing to generate boarding addresses from a miniscript descriptor.

please @altafan @tiero @sekulicd @bordalix  review

it closes #237


client/covenantless/onboard.go
```diff
@@ -17,6 +17,29 @@ import (
...
+		resp, err := client.ReverseBoardingAddress(ctx.Context, &arkv1.ReverseBoardingAddressRequest{
+			Pubkey: hex.EncodeToString(pubkey.SerializeCompressed()),
+		})
+		if err != nil {
+			return err
+		}
+
+		return utils.PrintJSON(resp)
+	}
+
```
tiero: Lets move this to receive

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -89,6 +95,15 @@ message CompletePaymentRequest {
...
 
+message ReverseBoardingAddressRequest {
+  string pubkey = 1;
+}
+
+message ReverseBoardingAddressResponse {
+  string address = 1;
+
+}
+
```
tiero: Its' try is handy to have server to generate the address, but we also need to give possibility to build the script "offline" assuming user after first connection caches the descriptor template accepted (ideally we could have an expiry time also attached to each descriptor template)

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -54,6 +54,12 @@ service ArkService {
       get: "/v1/info"
     };
   }
+  rpc ReverseBoardingAddress(ReverseBoardingAddressRequest) returns (ReverseBoardingAddressResponse) {
+    option (google.api.http) = {
+      post: "/v1/reverseboarding"
+      body: "*"
+    };
+  };
```
tiero, line 73: It is for sure handy to have server to generate the address, but we also need to give possibility to build the script "offline" assuming user after first connection caches the descriptor template accepted (ideally we could have an expiry time also attached to each descriptor template). Not needed for now, but I think we should include a way to enumerate also the script types of boarding address. @altafan

tiero: in the pending spec change in #228 we were indeed proposing to pass the descriptor and make the Onboard to be a RPC to be called AFTER the tx is sent to the static address. 



```proto
message OnboardRequest {
  string boarding_txid = 1;
  string descriptor = 2;
}
```

maybe we could still remove the Onboard RPC, and simply add a dedicated `BoardingUtxo` msg as accepted input for a payment, which will include both txid, descriptor. Server can then easily accept/reject registraton based on the descriptor

we need to add a `supported_boarding_script_templates` to `GetInfo`  so clients can catch that at first connection, and cache it

server/internal/core/application/covenant.go
```diff
@@ -114,18 +115,43 @@ func (s *covenantService) Stop() {
 	close(s.onboardingCh)
 }
 
-func (s *covenantService) SpendVtxos(ctx context.Context, inputs []domain.VtxoKey) (string, error) {
-	vtxos, err := s.repoManager.Vtxos().GetVtxos(ctx, inputs)
+func (s *covenantService) CreateReverseBoardingAddress(ctx context.Context, userPubkey *secp256k1.PublicKey) (string, error) {
+	return "", fmt.Errorf("unimplemented")
```
altafan: giving the impact of this kind of boarding on liquidity requirements i think it does make really sense to have it implemented also on the covenant version in this pr

api-spec/protobuf/ark/v1/service.proto
altafan: actually, for the covenant and covenat-less versions, the script for this kind of onboarding is the same (`A + S or A after 6 months`) so we don't really need the enumeration for now.

We don't really need any new rpc at all if we move the script/address creation client-side.

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -65,12 +65,12 @@ service ArkService {
       get: "/v1/info"
     };
   }
-  rpc Onboard(OnboardRequest) returns (OnboardResponse) {
+  rpc ReverseBoardingAddress(ReverseBoardingAddressRequest) returns (ReverseBoardingAddressResponse) {
     option (google.api.http) = {
-      post: "/v1/onboard"
+      post: "/v1/reverseboarding"
```
tiero: Just boarding

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -100,6 +100,15 @@ message CompletePaymentRequest {
 }
 message CompletePaymentResponse {}
 
+message ReverseBoardingAddressRequest {
+  string pubkey = 1;
+}
+
+message ReverseBoardingAddressResponse {
+  string address = 1;
+
```
tiero: Return descriptor also?

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -177,14 +188,7 @@ message GetInfoResponse {
...
   int64 min_relay_fee = 6;
-}
-
-message OnboardRequest {
-  string boarding_tx = 1;
-  Tree congestion_tree = 2;
-  string user_pubkey = 3;
-}
-message OnboardResponse {
+  int64 reverse_boarding_exit_delay = 7;
```
tiero: Better return a descriptor also here?

tiero: Just boarding_descriptor_template

tiero (changes requested):

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -242,6 +246,7 @@ message Round {
 message Input {
   string txid = 1;
   uint32 vout = 2;
+  optional string reverse_boarding_pubkey = 3;
```
tiero: descriptor

tiero (changes requested):

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -239,9 +243,22 @@ message Round {
...
   string txid = 1;
   uint32 vout = 2;
+  string descriptor = 3;
+}
+
+message Input {
+  oneof input {
+    VtxoInput vtxo_input = 1;
+    DescriptorInput descriptor_input = 2;
+  }
```
altafan: why not just add a descriptor field to the existing Input message?
```suggestion
message Input {
  string txid = 1;
  uint32 vout = 2;
  string descriptor = 3;
}
```

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -239,9 +243,22 @@ message Round {
   RoundStage stage = 8;
 }
 
-message Input {
+message VtxoInput {
+  string txid = 1;
+  uint32 vout = 2;
+}
+
+message DescriptorInput {
```
tiero:
```suggestion
message BoardingInput {
```

client/flags/flags.go
```diff
@@ -90,4 +90,10 @@ var (
 		Value:    false,
 		Required: false,
 	}
+	ReverseOnboardingFlag = cli.BoolFlag{
+		Name:     "reverse",
+		Usage:    "reverse onboarding protocol",
+		Value:    false,
+		Required: false,
+	}
```
altafan: not needed anymore?

tiero:
Remember toi update the PR description, so its useful for the future

tiero (approved):

PR #281, opened 2024-08-30
[SDK] Expose transport client

### LINKED ISSUE #263: arkd should return a history of transactions

altafan:
This makes the transport client of the SDK accessible so that the consumer can interact directly with the ASP for custom purposes.

Closes #263.

Please @louisinger @bordalix review.

### DIFF
--- a/pkg/client-sdk/ark_sdk.go
+++ b/pkg/client-sdk/ark_sdk.go
@@ -3,6 +3,7 @@ package arksdk
 import (
 	"context"
 
+	"github.com/ark-network/ark/pkg/client-sdk/client"
 	"github.com/ark-network/ark/pkg/client-sdk/store"
 )
 
@@ -26,6 +27,8 @@ type ArkClient interface {
 	) (string, error)
 	SendAsync(ctx context.Context, withExpiryCoinselect bool, receivers []Receiver) (string, error)
 	ClaimAsync(ctx context.Context) (string, error)
+
+	Client() client.ASPClient
 }
 
 type Receiver interface {
--- a/pkg/client-sdk/client.go
+++ b/pkg/client-sdk/client.go
@@ -47,6 +47,10 @@ type arkClient struct {
 	client   client.ASPClient
 }
 
+func (a *arkClient) Client() client.ASPClient {
+	return a.client
+}
+
 func (a *arkClient) GetConfigData(
 	_ context.Context,
 ) (*store.StoreData, error) {


louisinger (approved):
ACK

sekulicd:
Bellow are methods exposed by newly added transport client:

```
GetRound(ctx context.Context, txID string) -> (*Round, error)
ClaimPayment(ctx context.Context, paymentID string, outputs []Output) -> error
Onboard(ctx context.Context, tx string, userPubkey string, congestionTree tree.CongestionTree) -> error
GetInfo(ctx context.Context) -> (*Info, error)
FinalizePayment(ctx context.Context, signedForfeitTxs []string) -> error
Close() -> interface {}
CompletePayment(ctx context.Context, signedRedeemTx string, signedUnconditionalForfeitTxs []string) -> error
CreatePayment(ctx context.Context, inputs []VtxoKey, outputs []Output) -> (string, []string, error)
GetEventStream(ctx context.Context, paymentID string) -> (<-chan RoundEventChannel, error)
GetRoundByID(ctx context.Context, roundID string) -> (*Round, error)
ListVtxos(ctx context.Context, addr string) -> ([]Vtxo, []Vtxo, error)
Ping(ctx context.Context, paymentID string) -> (*RoundFinalizationEvent, error)
RegisterPayment(ctx context.Context, inputs []VtxoKey) -> (string, error)
```

Feel like we are duplicating stuff, maybe better to expose only methods that are missing?
@altafan i see you are closing txs history issue, which exactly method should be used to fetch these txs?

sekulicd:
> ListVtxos but exposing the whole client allows the user to interact directly with the ASP instead of using our tailored way of > doing so

Seems that before this ArkClient were only missing ListVtxos? should we just add this method instead returning whole client?

tiero:
I think we should have a more high-level `TransactionHistory` that returns all possible extra-metadata beyon ListVtxos, which seems more like low-level down the road. 

Client after all can always use the raw gRPC client if you have very specific needs, as I think we publish the proto

tiero (changes requested):
To close https://github.com/ark-network/ark/issues/263 we should expose a more high level `TransactionHistory` that returns entries sorted by latest first. Each entry should already represent a transaction with his state and details (inbound, outbound, swap, pending, failed) 

Client can already auto-generate client stubs from the public proto API spec, if they need low-level control.

altafan:
one problem is that for example the user could start reading from the event stream channel and compromise the behavior of the ark client. Exposing the client is not safe. Until we provide an high level API for accessing the tx history, any client can  auto-generate stubs from the public proto API spec as suggested by @tiero.

Closing.

tiero:
> About the PR itself, if we allow the user to auto-generate stubs for the client, why not giving him a easier way to access the client? What are the main problems of having arksdk.arkClient.Client()?

You can already use the client stubs from master branch or any other tags with

With latest commit from master 
```go
go get github.com/ark-network/ark/api-spec@c183f992449e588849b02e31097a7a5e91174c40
```

and import as

```go
api "github.com/ark-network/ark/api-spec/protobuf/gen/ark/v1"
```

PR #282, opened 2024-08-30
[SDK] Export ListVtxos

altafan:
This allows users to list vtxos of their offchain wallet via the SDK

Please @bordalix @louisinger @sekulicd review this.


bordalix (commented):
ACK

PR #286, opened 2024-09-02
add ZMQ env vars

tiero:


server/internal/infrastructure/wallet/btc-embedded/wallet.go
```diff
@@ -201,6 +201,62 @@ func WithPollingBitcoind(host, user, pass string) WalletOption {
...
+		bitcoindConn, err := chain.NewBitcoindConn(bitcoindConfig)
+		if err != nil {
+			return fmt.Errorf("failed to create bitcoind connection: %w", err)
+		}
+
+		if err := bitcoindConn.Start(); err != nil {
+			return fmt.Errorf("failed to start bitcoind connection: %w", err)
+		}
+
+		chainClient := bitcoindConn.NewBitcoindClient()
```
tiero, line 290: I have a problem using a local node with `http` instead of `https`: it wont work unless you patch `btcd` method!

```
// NewBitcoindConn creates a client connection to the node described by the host
// string. The ZMQ connections are established immediately to ensure liveness.
// If the remote node does not operate on the same bitcoin network as described
// by the passed chain parameters, the connection will be disconnected.
func NewBitcoindConn(cfg *BitcoindConfig) (*BitcoindConn, error) {
	clientCfg := &rpcclient.ConnConfig{
		Host:                 cfg.Host,
		User:                 cfg.User,
		Pass:                 cfg.Pass,
		DisableAutoReconnect: false,
		DisableConnectOnNew:  true,
		DisableTLS:           **true**,
		HTTPPostMode:         true,
		Params:               cfg.ChainParams.Name,
	}
```

PR #288, opened 2024-09-04
[SDK] Transaction History

### LINKED ISSUE #263: arkd should return a history of transactions

sekulicd:
This adds GetTransactionHistory API to the Client SDK.

This closes #263 

@tiero @altafan @bordalix @louisinger please review.

### DIFF
--- a/pkg/client-sdk/client.go
+++ b/pkg/client-sdk/client.go
@@ -229,6 +229,49 @@ func (a *arkClient) ListVtxos(
 	return
 }
 
+func (a *arkClient) GetTransactionHistory(ctx context.Context) ([]Transaction, error) {
+	var txs []Transaction
+
+	spendableVtxos, spentVtxos, err := a.ListVtxos(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	poolTxIDs := make(map[string]struct{})
+	for _, vtxo := range spendableVtxos {
+		poolTxIDs[vtxo.Txid] = struct{}{}
+		txs = append(
+			txs,
+			Transaction{
+				TxID:   vtxo.RoundTxid,
+				Amount: vtxo.Amount,
+				Type:   TxReceived,
+			},
+		)
+	}
+
+	for _, vtxo := range spentVtxos {
+		// in case bob receives shortcut payment from alice, after claiming the payment,
+		// there will be two vtxos, one in spendable and one in spent vtxo,
+		// with bellow condition we are avoiding to show the spent vtxo in transaction history
+		// as Bob only received the payment and didn't send it to anyone
+		if _, ok := poolTxIDs[vtxo.SpentBy]; ok && vtxo.Pending {
+			continue
+		}
+
+		txs = append(
+			txs,
+			Transaction{
+				TxID:   vtxo.RoundTxid,
+				Amount: vtxo.Amount,
+				Type:   TxSent,
+			},
+		)
+	}
+
+	return txs, nil
+}
+
 func (a *arkClient) ping(
 	ctx context.Context, paymentID string,
 ) func() {
--- a/pkg/client-sdk/types.go
+++ b/pkg/client-sdk/types.go
@@ -129,3 +129,16 @@ type balanceRes struct {
 	offchainBalanceByExpiration map[int64]uint64
 	err                         error
 }
+
+const (
+	TxSent     TxType = "sent"
+	TxReceived TxType = "Received"
+)
+
+type TxType string
+
+type Transaction struct {
+	TxID   string
+	Amount uint64
+	Type   TxType
+}
--- a/pkg/client-sdk/ark_sdk.go
+++ b/pkg/client-sdk/ark_sdk.go
@@ -27,7 +27,8 @@ type ArkClient interface {
 	) (string, error)
 	SendAsync(ctx context.Context, withExpiryCoinselect bool, receivers []Receiver) (string, error)
 	ClaimAsync(ctx context.Context) (string, error)
-	ListVtxos(ctx context.Context) ([]client.Vtxo, []client.Vtxo, error)
+	ListVtxos(ctx context.Context) (spendable []client.Vtxo, spent []client.Vtxo, err error)
+	GetTransactionHistory(ctx context.Context) ([]Transaction, error)
 }
 
 type Receiver interface {


pkg/client-sdk/types.go
```diff
@@ -129,3 +129,16 @@ type balanceRes struct {
 	offchainBalanceByExpiration map[int64]uint64
 	err                         error
 }
+
+const (
+	TxSent     TxType = "sent"
+	TxReceived TxType = "Received"
```
louisinger:
```suggestion
	TxReceived TxType = "received"
```

bordalix (commented):
NACK.

1. If I understood the code correctly, this will not show pending transactions (i.e. not claimed) but Bob needs to know there's a pending transaction for him to claim.

2. Transactions need a timestamp

pkg/client-sdk/types.go
```diff
@@ -129,3 +129,16 @@ type balanceRes struct {
...
+)
+
+type TxType string
+
+type Transaction struct {
+	TxID   string
+	Amount uint64
+	Type   TxType
+}
```
altafan:
```suggestion
type Transaction struct {
	TxID   string
	Amount uint64
	Type   TxType
	Pending bool
	Timestamp int64
}
```

pkg/client-sdk/client.go
```diff
@@ -229,6 +229,49 @@ func (a *arkClient) ListVtxos(
 	return
 }
 
+func (a *arkClient) GetTransactionHistory(ctx context.Context) ([]Transaction, error) {
```
altafan: We can change the logic of this function as follows:
```
GetTransactionHistory {
   init txs map[string]Transaction // used to return the []Transaction at the end
   init help map[string]string // used to update a pending tx to claimed if needed
   
   loop spentVtxos {
      if vtxo is pending {
          help[vtxo.SpentBy] = vtxo.RedeemTx.Txid
          txs[vtxo.RedeemTx.Txid] = Transaction{...}
          continue
      }
      txs[vtxo.RoundTxid] = Transaction{...}
   }
   
   loop spendableVtxos {
       if vtxo is pending {
           txs[vtxo.RedeemTxid] = Transaction{...}
           continue
       }
       if help[vtxo.RoundTxid] {
           txs[help[vtxo.RoundTxid]].Pending = false
       } else {
           txs[vtxo.RoundTxid] = Transaction{...}
       }
   }
}
```

altafan: This way we prevent having multiple txs representing a single claim.
Not completely convinced of using just a Pending field to represent pending/claimed status of an async payment... maybe better to have in Transaction both Pending and Claimed bool field

pkg/client-sdk/covenant_client.go
```diff
@@ -509,6 +510,92 @@ func (a *covenantArkClient) Claim(ctx context.Context) (string, error) {
...
+func (a *covenantArkClient) GetTransactionHistory(ctx context.Context) ([]Transaction, error) {
+	spendableVtxos, spentVtxos, err := a.ListVtxos(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return vtxosToTxsCovenantless(spendableVtxos, spentVtxos)
+}
+
+func vtxosToTxsCovenantless(spendable, spent []client.Vtxo) ([]Transaction, error) {
```
altafan:
```suggestion
func vtxosToTxsCovenant(spendable, spent []client.Vtxo) ([]Transaction, error) {
```

pkg/client-sdk/covenant_client.go
```diff
@@ -509,6 +510,92 @@ func (a *covenantArkClient) Claim(ctx context.Context) (string, error) {
...
+	txsMap := make(map[string]Transaction)
+	pendingTxs := make(map[string]string)
+
+	// Process spent transactions
+	for _, vtxo := range spent {
+		txID := vtxo.RoundTxid
+		pending := false
+
+		if len(vtxo.UnconditionalForfeitTxs) > 0 && vtxo.RedeemTx != "" {
+			redeemPtx, err := psbt.NewFromRawBytes(strings.NewReader(vtxo.RedeemTx), true)
```
altafan: we must use psetv2 to decode the tx here

pkg/client-sdk/covenantless_client.go
```diff
@@ -633,6 +633,87 @@ func (a *covenantlessArkClient) Claim(ctx context.Context) (string, error) {
...
+func (a *covenantlessArkClient) GetTransactionHistory(ctx context.Context) ([]Transaction, error) {
+	spendableVtxos, spentVtxos, err := a.ListVtxos(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return vtxosToTxs(spendableVtxos, spentVtxos)
+}
+
+func vtxosToTxs(spendable, spent []client.Vtxo) ([]Transaction, error) {
```
altafan: we should call this like the following for consistency maybe?
```suggestion
func vtxosToTxsCovenantless(spendable, spent []client.Vtxo) ([]Transaction, error) {
```

pkg/client-sdk/covenantless_client.go
```diff
@@ -633,6 +633,87 @@ func (a *covenantlessArkClient) Claim(ctx context.Context) (string, error) {
...
+				Amount:  vtxo.Amount,
+				Type:    TxReceived,
+				Pending: true,
+			}
+		} else if pendingTxID, ok := pendingTxs[txID]; ok {
+			// Update the transaction if it's pending and now claimed
+			if tx, exists := txsMap[pendingTxID]; exists {
+				tx.Pending = false
+				tx.Claimed = true
+				tx.Type = TxReceived
```
altafan: Let's drop this line and keep the original type

bordalix:
@sekulicd still no timestamp on Transaction.

You can use (ExpiresAt - timeout) to figure it out when the VTXO/Tx was created.

pkg/client-sdk/client_test.go
```diff
@@ -0,0 +1,378 @@
...
+					Amount:  20000,
+					Type:    TxSent,
+					Pending: false,
+					Claimed: false,
+				},
+				{
+					TxID:    "94fa598302f17f00c8881e742ec0ce2f8c8d16f3d54fe6ba0fb7d13a493d84ad",
+					Amount:  19000,
+					Type:    TxReceived,
+					Pending: true,
```
altafan: let's add a TODO to remember that this will change once the ASP handles the change of an async payment as a normal vtxo:
```suggestion
					Pending: true, // TODO: expect false once the ASP handles the change properly
```

pkg/client-sdk/covenant_client.go
sekulicd: Done

pkg/client-sdk/covenant_client.go
sekulicd: Done

pkg/client-sdk/covenantless_client.go
sekulicd: Done

pkg/client-sdk/client_test.go
sekulicd: Done

sekulicd:
@altafan please verify PR now, for gosec thing i excluded 'G115 (CWE-190): integer overflow conversion uint64 -> int64' completely and few other errors i add ignore comment in code that i find its safe to ignore.

pkg/client-sdk/types.go
```diff
@@ -129,3 +130,19 @@ type balanceRes struct {
...
+
+const (
+	TxSent     TxType = "sent"
+	TxReceived TxType = "received"
+)
+
+type TxType string
+
+type Transaction struct {
+	TxID      string
```
louisinger: looking at the implementation it seems that the TxID member is either set to round ID or to redeem_tx ID (so the "vtxo ID"). 
I do think it can be misleading, why not adding Round TxID and Redeem TxID (or "vtxo txid") for all type of vtxos ? also maybe the vout of the vtxo could be useful ? 

```suggestion
	TxID      string // vtxo txid
	TxIndex uint32
	RoundTxID string 
```

sekulicd: Should i rename TxID to VtxoID and add VOut ? @louisinger

PR #303, opened 2024-09-10
fix: support http as asp url

bordalix:
Fix bug where `http://localhost:7070` for ASP url doesn't work but it should.

@altafan please review


tiero:
merge master (we removed codeQL for time being)

PR #311, opened 2024-09-15
Reversible vtxo in async payment

louisinger:
_This PR adds the necessary abstraction to process different types of VTXOs. It allows adding the `ReversibleVtxoScript` type used as output of async payments._

### Vtxo script type

* A vtxo script is defined as a taproot contract with at least 1 forfeit closure (`User && ASP`) and 1 exit closure (`A after t`). It may contain others closures implementing specific use cases.
* `common.VtxoScript` abstracts the taproot complexity of vtxo script. 
* a vtxo script is compiled, transferred and parsed using descriptor string.
* the default vtxo script is `tr(_,{ and(pk(USER), pk(ASP)), and(older(T), pk(USER)) })`
* the reversible vtxo script is `tr(_,{ { and(pk(SENDER), pk(ASP)), and(older(T), pk(SENDER)) }, { and(pk(RECEIVER), pk(ASP) } })`
* while spending a VTXO, the spender must provide the associated descriptor and the signer public key (in order to signal to the ASP the forfeit leaf to use)

### TL;DR

* [common] add `VtxoScript` interface
* [tx-builder] use `VtxoScript` to craft vtxo tree transactions
* [api-spec] add `descriptor` instead of pubkey while registering inputs
* [api-spec] add `signingPubKey` when spending a VTXO
* [api-spec] specify `descriptor` instead of offchain address while registering outputs
* [db] replace `Vtxo.Pubkey` by `Vtxo.Descriptor`

_PS: This PR does not add the necessary change to support the "Reverse" UX in the SDK even if it's technically possible today for the sender to spend (via reverse forfeit leaf) an async payment VTXO. It may need extra work to keep track of the reversible VTXOs and coin-select them._ 

@altafan @tiero @sekulicd @bordalix  please review


tiero:
May be worth to lay down the ground work of key path spending for cooperative cases? We may not need to migrate right now, but at least figuring out the abstraction of needs some more work in the future, better now

common/bitcointree/forfeit.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"github.com/ark-network/ark/common"
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
+)
+
+func MakeForfeitTxs(
```
altafan:
```suggestion
func BuildForfeitTxs(
```

common/bitcointree/type.go
```diff
@@ -1,6 +1,6 @@
 package bitcointree
 
 type Receiver struct {
-	Pubkey string
+	Script VtxoScript
```
altafan, line 4: why not using a more high-level `Descriptor string` and then we translate it to vtxo script by calling the parse func?

louisinger: This `bitcointree.Receiver` struct is used across a lot of functions, including the recursive tree builder, so the idea is to avoid parsing every time we need to use the script methods whereas the `domain.Receiver` (used by interface layer and DB) has a Descriptor string field

server/internal/infrastructure/tx-builder/covenant/builder_test.go
```diff
@@ -110,7 +109,7 @@ func TestBuildForfeitTxs(t *testing.T) {
 		t.Run("valid", func(t *testing.T) {
 			for _, f := range fixtures.Valid {
 				connectors, forfeitTxs, err := builder.BuildForfeitTxs(
-					pubkey, f.PoolTx, f.Payments,
+					pubkey, f.PoolTx, f.Payments, 3,
```
altafan: let's use a var explaining what that `3` is

server/test/e2e/covenantless/e2e_test.go
```diff
@@ -102,7 +102,7 @@ func TestSendOffchain(t *testing.T) {
 func TestUnilateralExit(t *testing.T) {
 	var receive utils.ArkReceive
 	receiveStr, err := runClarkCommand("receive")
-	require.NoError(t, err)
+	require.NoError(t, err, receiveStr)
```
altafan: can you revert?

altafan (approved):

PR #314, opened 2024-09-16
CLI init check

sekulicd:
Prevent execution without initialisation by checking cfgData and returning error if not initialised.

@tiero please review.


altafan:
@sekulicd is this still needed after the latest commits pushed on master? can you double check please?

sekulicd:
> @sekulicd is this still needed after the latest commits pushed on master? can you double check please?

Before this, if user runs any cmd code would panic as config data is nil, so this checks if config data is not nil before each cmd except init, so yes this is still needed.

tiero:
Can you merge master? we should already have the app.Version fix there

sekulicd:
> Can you merge master? we should already have the app.Version fix there

Done

client/main.go
```diff
@@ -388,6 +384,12 @@ func getArkSdkClient(ctx *cli.Context) (arksdk.ArkClient, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	commandName := ctx.Args().First()
```
tiero, line 388: Doing the check here, will then result in concatenating to other err messages ie. `error: error initializing ark sdk client: ` Is this what we really want? bit ugly. Is it possible to make this check more before others?

```
ark config
error: error initializing ark sdk client: CLI not initialized, run 'init' cmd to initialize
```

sekulicd: maybe we just fmt.Print and return without err?

client/main.go
```diff
@@ -388,6 +384,12 @@ func getArkSdkClient(ctx *cli.Context) (arksdk.ArkClient, error) {
...
+	if commandName != "init" && cfgData == nil {
+		return nil, errors.New("CLI not initialized, run 'init' cmd to initialize")
```
altafan:
```suggestion
		return nil, fmt.Errorf("CLI not initialized, run 'init' cmd to initialize")
```

PR #300, opened 2024-09-10
test adversarial scenarios 

louisinger:
_This PR adds integration tests for double-spend cheating cases. It also includes several fixes in broadcasting the connectors and forfeits transactions._

**(1) react to double-spend VTXO spent by classic payment**
* ASP broadcasts the next connector tx of the `spentBy` round.
* ASP broadcasts the associated forfeit tx, getting the funds back to the ASP before redemption.

**(2) react to double-spend VTXO spent by async payment**
* ASP broadcasts the redeem transaction.
* If VTXO(s) in the redeem tx have been claimed : do (1)

it closes #291 
it closes #289 

@altafan please review

PR #316, opened 2024-09-16
[SDK] Export redeemption utils

altafan:
Needed by #300.

Please @louisinger review


altafan:
not needed after #317

PR #319, opened 2024-09-17
Test Covenant GetTransactionHistory 

sekulicd:
Although, this could chain in near future, ATM, async(pending) payments are related to BTC chain(covenantless) while regular(NON-PENDING) payments are related to Liquid chain(covenant).

This adds test case, for covenant GetTransactionHistory, that check bellow scenario(bellow u can see expectations as well):
1. before onboarding => empty history
2. after sending sats to the boarding address => 1 PENDING boarding tx
3. after claiming the boarded funds => 1 NON-PENDING boarding tx
4. after sending N sats to bob => 1 NON-PENDING sent tx, 1 NON-PENDING boading tx

Test is failing cause current code seems that it only supports BTC chain as it ignores non pending payments, [here](https://github.com/sekulicd/ark/blob/189aedba7fd9726268f70207e3f1fead0216f39b/pkg/client-sdk/covenant_client.go#L1475).
For covenant GetTransactionHistory returns only boarding txs.

@bordalix @altafan this looks like a bug, WDYT?


altafan (changes requested):
Please provide fixtures in client_test.go like done for the covenantless case.

We will move them to a dedicated `fixtures.json` file later.

client/main.go
```diff
@@ -23,13 +23,13 @@ const (
 )
 
 var (
-	version      = "alpha"
+	Version      string
```
tiero, line 26: tought was fixed 😅

altafan: the branch is not synced with master

PR #320, opened 2024-09-17
SDK - Tx Storage & Ark SDK func update 

sekulicd:
This adds storage for transactions and vtxos into SDK sqlite storage.
This refactors ark sdk api by adding GetTransactionEventChannel which can be used to listen for new events, it replaces PaymentNotification that was discussed before, [this](https://github.com/ark-network/ark/pull/273) to be closed.
It also updates GetTransactionHistory to read txs from db.
Main idea is to add listenForVtxos function that polls the server every 1(or smtn diff) seconds for new VTXOs, 
Func listenToVtxoChan will listen for new vtxos and with help of current state in db will create tx history by using existing 
vtxosToTxsCovenant/vtxosToTxsCovenantless funcs.

I pushed changes to the same PR in which i emphasised potential bug in how we handle covenant transaction history, PSB:

```
Although, this could chain in near future, ATM, async(pending) payments are related to BTC chain(covenantless) 
while regular(NON-PENDING) payments are related to Liquid chain(covenant).

This adds test case, for covenant GetTransactionHistory, that check bellow scenario(bellow u can see 
expectations as well):
1. before onboarding => empty history
2. after sending sats to the boarding address => 1 PENDING boarding tx
3. after claiming the boarded funds => 1 NON-PENDING boarding tx
4. after sending N sats to bob => 1 NON-PENDING sent tx, 1 NON-PENDING boading tx
Test is failing cause current code seems that it only supports BTC chain as it ignores non pending payments, here.
For covenant GetTransactionHistory returns only boarding txs.
```
@bordalix let me know if u agree that this is bug.

@tiero @louisinger @bordalix @altafan please review.


tiero:
is the title correct? Only touches the covenant version? We should mention a new big feature like adding a SQLite storage. Also how that will impact WASM wrapper?

bordalix:
@sekulicd we made some changes to covenant's GetTransactionHistory() yesterday.
Please merge master into your branch and check if error persists.

altafan (changes requested):
Better to start with KV implementation of the repos (badger) instead of sqlite. We'll add it once the persistency layer of the SDK client is validated.

pkg/client-sdk/client.go
```diff
@@ -51,12 +52,23 @@ var (
...
-	store    store.ConfigStore
-	explorer explorer.Explorer
-	client   client.ASPClient
+	wallet       wallet.WalletService
+	store        store.ConfigStore
+	appDataStore store.AppDataStore
+	explorer     explorer.Explorer
+	client       client.ASPClient
+
+	vtxosChan            chan map[spent]client.Vtxo
```
altafan: can you explain why `map[spent]client.Vtxo`?

pkg/client-sdk/client.go
```diff
@@ -130,9 +142,77 @@ func (a *arkClient) InitWithWallet(
...
+		spendable, spent, err := a.client.ListVtxos(ctx, addr)
+		if err != nil {
+			return nil, nil, err
+		}
+		spendableVtxos = append(spendableVtxos, spendable...)
+		spentVtxos = append(spentVtxos, spent...)
+	}
+
+	return
+}
```
altafan: after syncing with master this must be dropped

PR #326, opened 2024-09-22
keep track of forfeit transaction funds

### LINKED ISSUE #301: Forfeit transaction should use the wallet interface to compute the receiving output script
`TxBuilder` is using the ASP public key to create an ASP output script while craft the forfeit transactions. However this script is maybe not the one used by the `WalletInterface` implementation. We should use `GetNextAddress` instead of using the constant ASP public key.

louisinger:
This PR ensures that the output script of forfeit transactions are related to the wallet main account. It adds `ForfeitAddress` fields in `GetInfoResponse` and use it instead of computing the output script from ASP public key.

It allows the funds to be selected to fund next rounds when a forfeit transaction appears onchain.

it closes #301 

@altafan  @tiero  please review


common/bitcointree/forfeit.go
```diff
@@ -15,9 +14,14 @@ func BuildForfeitTxs(
 	connectorAmount,
 	feeAmount uint64,
 	vtxoScript []byte,
-	aspPubKey *secp256k1.PublicKey,
+	aspAddress string,
```
tiero: It seems inconsistent to work directly with Addresses at this level: why not using script directly? AFAIK network is not extracted here, and is constant for the bitcointree instance. I would try to standardize also this elsewhere.

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -189,6 +189,7 @@ message GetInfoResponse {
   string network = 5;
   int64 dust = 6;
   string boarding_descriptor_template = 7;
+  string forfeit_address = 8;
```
tiero, line 192: Worth using descriptor also to be consistent? As we already have the network as 5th.

louisinger: I don't see any advantage to use descriptor, there is no "arguments" and the user doesn't have to generate new addresses from it. Also, `forfeit_address` could be any type of address, so it means supporting all type of descriptors etc... sounds a bit overkill to use descriptor only for a recipient address

louisinger: also, an address can be easily copy/pasted in any explorer to verify if ASP puts some forfeits transactions onchain

altafan: What if, instead of this and pubkey, we returned the descriptor that specifies how the forfeit address is  done and encodes also the pubkey?

louisinger: public key = key of the ASP to use in VTXO script
forfeit address = ASP script to use when building the forfeits

the forfeit address / script is just a way to agree on a common script while adding the ASP output in forfeits. it does not have to contain the ASP public key. In fact that's not the case now, forfeit address is the first main account address (a segwit v0 account, not the one used to compute ASP public key)

forfeit address is a wallet interface address. First, none of the wallets we are using are descriptor based.
 And if I see the advantages of descriptors in case of taproot VTXO scripts cause it contains arguments etc... I'm not sure why we should use a descriptor to encode a classic p2wpkh address (also because our descriptor package only supports taproot descriptor so it means extra work there to just communicate an output script)

altafan (approved):

PR #332, opened 2024-09-26
Fix btc wallet restore

louisinger:
_This PR is a draft, waiting for https://github.com/btcsuite/btcwallet/issues/957_

TL;DR

- remove unused Rescan call
- first account index set to 0 ("default" name)

It fixes the restore process for the main account, but the connectors account (with index = 1) is not restored (see btcwallet issue).


server/internal/infrastructure/wallet/btc-embedded/wallet.go
```diff
@@ -289,11 +292,11 @@ func (s *service) GenSeed(_ context.Context) (string, error) {
 }
 
 func (s *service) Create(_ context.Context, seed, password string) error {
-	return s.create(seed, password, 0)
+	return s.create(seed, password, 512)
```
altafan, line 295: Create is meant for creating a wallet from scratch, that's why the 0 time window. Please revert

altafan: Create is meant for creating a wallet from scratch, that's why the 0 time window. Please revert:

```suggestion
	return s.create(seed, password, 0)
```

tiero:
Reason to NOT mix the connectors scripts in the same main account script? We still know which is which by database no?

louisinger:
> Reason to NOT mix the connectors scripts in the same main account script? We still know which is which by database no? 

connectors must not be spent so having two different accounts makes the coin selection simpler. 
also for ASP "analysis", I do think it makes the available liquidity clearer if we don't mix connectors and main funds. 

another way to solve would be to use p2tr scope for connector instead of p2wpkh : this way the connector account index will be 0 and should be restored

tiero:
> p2tr scope 

Care to elaborate here?

> we don't mix connectors and main funds.

This is important, and assigning a bip44 account to each makes sense, but we could have those information in database worst case.

altafan (approved):

PR #333, opened 2024-09-26
Make embedded btc wallet produce less logs

### LINKED ISSUE #287: arkd: less important logs should be log level debug
for example neutrino stuff should not be log level Info. Also the `config.Params` like in make run

```sh
make run
Cleaning...
Running arkd with Bitcoin Core in regtest mode ...
INFO[0000] All 0 tables opened in 0s
INFO[0000] Discard stats nextEmptySlot: 0
INFO[0000] Set nextTxnTs to 0
config.Params:  regtest
config.Params:  regtest
```

altafan:
There's no way to change the level of the logs a service produces. Rather, we can set the the min log level for it and make those of the levels above disappear.

This makes the embedded btc wallet produce fewer logs by setting the log level of all components to warning.

NOTE: There's no way to set a log level for the sqlite db we use currently `modernc.org/sqlite`. Maybe we should move to something else? Is there a specific reason we chose this one? @louisinger @tiero

Closes #287

Please @louisinger @tiero review

### DIFF
--- a/server/internal/infrastructure/wallet/btc-embedded/wallet.go
+++ b/server/internal/infrastructure/wallet/btc-embedded/wallet.go
@@ -1299,5 +1299,7 @@ func fromOutputScript(script []byte, netParams *chaincfg.Params) (btcutil.Addres
 }
 
 func logger(subsystem string) btclog.Logger {
-	return btclog.NewBackend(log.StandardLogger().Writer()).Logger(subsystem)
+	logger := btclog.NewBackend(log.StandardLogger().Writer()).Logger(subsystem)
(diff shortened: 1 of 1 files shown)


tiero:
It's the only go native so no CGO

tiero:
Something even lower than warn? Like debug or trace?

altafan:
These are the [levels](https://github.com/btcsuite/btclog/blob/master/log.go#L83-L89). If we set _Trace_ log level, it means we want to see the logs of all levels, if we use _Debug_  we exclude only trace logs and so on. We use _Warn_ to see only warning errors and criticals and exclude all the other levels.

PR #336, opened 2024-09-27
fix: `createPoolTx` coin selection

louisinger:
This PR is a rework of the `createPoolTx` coin selection strategy. It fixes "dust" errors happening if the second fee estimation fails.

@altafan please review

### DIFF
--- a/server/internal/infrastructure/tx-builder/covenantless/builder.go
+++ b/server/internal/infrastructure/tx-builder/covenantless/builder.go
@@ -699,19 +699,24 @@ func (b *txBuilder) createPoolTx(
 		return nil, err
 	}
 
-	changeAddresses, err := b.wallet.DeriveAddresses(ctx, 1)
-	if err != nil {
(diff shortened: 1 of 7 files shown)


server/internal/infrastructure/tx-builder/covenantless/builder.go
```diff
@@ -699,30 +699,30 @@ func (b *txBuilder) createPoolTx(
...
+	}
 
-			aspScript, err := txscript.PayToAddrScript(addr)
-			if err != nil {
-				return nil, err
-			}
+	changeScript, err := txscript.PayToAddrScript(changeAddress)
+	if err != nil {
+		return nil, err
+	}
```
altafan: not that great to derive a change address in case the change is 0 or below the dust. Please revert

tiero:
createPoolTx > createRoundTx as we are at it

altafan (approved):

PR #338, opened 2024-09-27
SDK - Badger store, GetTransactionHistory & GetTransactionEventChannel

sekulicd:
This add new GetTransactionEventChannel RPC which adds option to the user to track transaction events.
This updates GetTransactionHistory so that it fetches txs from db in case GetTransactionEventChannel is invoked which start listening for Vtxo's and Boarding txs and recreated tx history, or in case GetTransactionEventChannel is not invoked it recreates full txs history on the fly.
This adds transaction cdm to the CLI that lists all transactions.

@tiero @altafan please review.


altafan:
merged #354 instead

PR #345, opened 2024-10-03
RPC GetPaymentsStream

sekulicd:
Add new RPC GetPaymentsStream to enable accurate payment history reconstruction

This commit introduces the GetPaymentsStream RPC to the ArkService API, allowing clients to subscribe to payment events and reliably recreate the payment history.

This change addresses issues identified in [PR #338](https://github.com/ark-network/ark/pull/338), where we found that it was not possible to precisely reconstruct all payment (transaction) events solely on the client side. Specifically, we faced challenges in recognizing when boarding transactions were claimed due to race conditions between scanning the boarding transaction's spent status via the explorer and receiving the same event in the round transaction.

By providing this new RPC, clients can now receive real-time updates on payment events, ensuring they can accurately and consistently track payment histories without relying on potentially inconsistent client-side event reconstruction.

@tiero @altafan @louisinger  @bordalix  please review.


api-spec/protobuf/ark/v1/service.proto
```diff
@@ -321,3 +326,24 @@ message PendingPayment {
...
+    RoundPayment round_payments = 1;
+    AsyncPayment async_payments = 2;
+  }
+}
+
+message RoundPayment {
+  string txid = 1;
+  repeated Outpoint spent_vtxos = 2;
+  repeated Vtxo spendable_vtxos = 3;
+  repeated Outpoint claimed_boarding_vtxos = 4;
```
altafan: These are the extra inputs of the round tx, ie the boarded utxos. Let's call them utxos, not vtxos:

```suggestion
  repeated Outpoint claimed_boarding_utxos = 4;
```

server/internal/core/application/covenantless.go
```diff
@@ -1097,6 +1111,15 @@ func (s *covenantlessService) finalizeRound() {
...
 	}
 
+	go func() {
+		s.paymentEventsCh <- RoundPaymentEvent{
+			RoundTxID:             round.Txid,
+			SpentVtxos:            getSpentVtxos(round.Payments),
+			SpendableVtxos:        s.getNewVtxos(round),
+			ClaimedBoardingInputs: boardingInputs,
+		}
+	}()
```
altafan: this should be done in [updateVtxoSet](https://github.com/ark-network/ark/blob/master/server/internal/core/application/covenantless.go#L1220-L1266) (same for covenant)

sekulicd: dont have there boardingInputs?

tiero:
We should not user Payments, but more general Transactions

api-spec/protobuf/ark/v1/service.proto
sekulicd: done

sekulicd:
> We should not user Payments, but more general Transactions

@tiero i am not sure about this, Round and Async payment are in fact transactions but inside there can be multiple "payments" where multiple users could send/receive multiple coins(vtxos), also we already name in/outs pairs in Round domain as Payments and adding additional name to the spec can be bit confusing.

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -92,6 +92,11 @@ service ArkService {
       get: "/v1/vtxos/{address}"
     };
   }
+  rpc GetPaymentsStream(GetPaymentsStreamRequest) returns (stream GetPaymentsStreamResponse) {
+    option (google.api.http) = {
+      get: "/v1/payments"
+    };
+  }
```
altafan:
```suggestion
  rpc GetTransactionsStream(GetTransactionsStreamRequest) returns (stream GetTransactionsStreamResponse) {
    option (google.api.http) = {
      get: "/v1/transactions"
    };
  }
```

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -321,3 +326,24 @@ message PendingPayment {
...
   repeated string unconditional_forfeit_txs =2;
 }
+
+message GetPaymentsStreamRequest {}
+message GetPaymentsStreamResponse {
+  oneof tx {
+    RoundPayment round_payments = 1;
+    AsyncPayment async_payments = 2;
+  }
+}
```
altafan:
```suggestion
message GetTransactionsStreamRequest {}
message GetTransactionsStreamResponse {
  oneof tx {
    RoundTransaction round = 1;
    RedeemTransaction redeem = 2;
  }
}
```

sekulicd: Done

api-spec/protobuf/ark/v1/service.proto
sekulicd: Done

altafan (approved):

PR #353, opened 2024-10-10
[SDK] Fixes to WASM and REST client

bordalix:
@altafan please review


pkg/client-sdk/covenant_client.go
```diff
@@ -1800,7 +1800,7 @@ func vtxosToTxsCovenant(
 			},
 			Amount:    uint64(math.Abs(float64(amount))),
 			Type:      txType,
-			CreatedAt: getCreatedAtFromExpiry(roundLifetime, *v.ExpiresAt),
+			CreatedAt: getCreatedAtFromExpiry(roundLifetime, v.ExpiresAt),
```
louisinger, line 1803: Not clear to me why we can't use v.CreatedAt here ?

pkg/client-sdk/covenantless_client.go
```diff
@@ -1520,14 +1526,32 @@ func (a *covenantlessArkClient) handleRoundStream(
 			if notify.Err != nil {
 				return "", notify.Err
 			}
+			if notify.Event == nil {
```
louisinger, line 1529: why not moving this in REST client ? It seems not useful in case of gRPC stream

pkg/client-sdk/covenantless_client.go
```diff
@@ -1047,34 +1050,32 @@ func (a *covenantlessArkClient) Settle(ctx context.Context) (string, error) {
...
+		return a.store.TransactionStore().GetAllTransactions(ctx)
+	}
+
 	spendableVtxos, spentVtxos, err := a.ListVtxos(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	boardingTxs, ignoreVtxos, err := a.getBoardingTxs(ctx)
+	boardingTxs, _, err := a.getBoardingTxs(ctx)
```
louisinger, line 1066:
```suggestion
	boardingTxs, err := a.getBoardingTxs(ctx)
```

the second return var is never used, let's remove it

louisinger (commented):
ACK

altafan (approved):

PR #356, opened 2024-10-17
New address encoding

louisinger:
_This PR adds the new ark address encoding, replacing bare user public key by vtxo taproot key. It implies change on the APIs and the SDK._

image

### Address encoding

An ark address is `network prefix + VTXO taproot pubkey + ASP pubkey` encoded with [bech32m](https://bitcoinops.org/en/topics/bech32/).
```
ark1x0lm8hhr2wc6n6lyemtyh9rz8rg2ftpkfun46aca56kjg3ws0tsztfpuanaquxc6faedvjk3tax0575y6perapg3e95654pk8r4fjecs5fyd2
```
* _network prefix_ identifies testnet, mainnet, regtest...
* _VTXO taproot pubkey_ is the P2TR key that should be used as scriptpubkey in VTXO leaf output of the tree.
* _ASP pubkey_ allows verifying the ASP of the address

### API & server changes

* remove descriptor from `RegisterOutputsForNextRound` RPC, only ark address is expected.
* expect descriptor to be revealed while spending VTXO = `RegisterInputsForNextRound` RPC.
* [db] VTXO table: drop descriptor field, replace by pubkey field.

image

@altafan @tiero please review


server/internal/infrastructure/db/sqlite/migration/20241015101830_remove-descriptor.down.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE vtxo DROP COLUMN addr;
```
altafan: let's merge all migration files into a single one, we don't really care now about doing this politely.

altafan:
> [db] VTXO table: drop descriptor field, replace by address field.

@louisinger let's replace the descriptor with the taproot pubkey instead of the address:
- we don't want to "depend" on network as address contains a network prefix
- we don't want to persist the same server pubkey for all records

kentaurse:
hello, @louisinger 
Bech32 offers a robust and error-resistant format for encoding new address types in Go, leveraging its strong checksum and simplicity, making it an ideal choice for handling public key combinations.

`
package main

import (
    "encoding/hex"
    "fmt"
    "github.com/btcsuite/btcutil/bech32"
)

// EncodeAddress encodes the new Ark address using Bech32.
func EncodeAddress(networkPrefix, vtxoPubkey, aspPubkey string) (string, error) {
    // Combine the public keys into a single byte slice
    combinedPubkey := vtxoPubkey + aspPubkey
    pubkeyBytes, err := hex.DecodeString(combinedPubkey)
    if err != nil {
        return "", err
    }

    // Encode using Bech32
    encodedAddress, err := bech32.Encode(networkPrefix, pubkeyBytes)
    if err != nil {
        return "", err
    }

    return encodedAddress, nil
}

func main() {
    networkPrefix := "ark1"
    vtxoPubkey := "your_vtxo_pubkey_here"
    aspPubkey := "your_asp_pubkey_here"

    newAddress, err := EncodeAddress(networkPrefix, vtxoPubkey, aspPubkey)
    if err != nil {
        fmt.Println("Error encoding address:", err)
        return
    }

    fmt.Println("Encoded Address:", newAddress)
}
`
`
package main

import (
    "encoding/json"
    "net/http"
)

func registerAddressHandler(w http.ResponseWriter, r *http.Request) {
    var data struct {
        NetworkPrefix string `json:"network_prefix"`
        VtxoPubkey    string `json:"vtxo_pubkey"`
        AspPubkey     string `json:"asp_pubkey"`
    }

    if err := json.NewDecoder(r.Body).Decode(&data); err != nil {
        http.Error(w, "Invalid request", http.StatusBadRequest)
        return
    }

    newAddress, err := EncodeAddress(data.NetworkPrefix, data.VtxoPubkey, data.AspPubkey)
    if err != nil {
        http.Error(w, "Error encoding address", http.StatusInternalServerError)
        return
    }

    json.NewEncoder(w).Encode(map[string]string{"address": newAddress})
}
`
`
-- SQL command to update the VTXO table schema
ALTER TABLE vtxo_table
DROP COLUMN descriptor,  -- Remove the old descriptor field
ADD COLUMN pubkey VARCHAR(255);  -- Add a new pubkey field to store the VTXO public key
`

altafan (approved):

PR #359, opened 2024-10-21
Out of round 

louisinger:
_This PR adds the necessary changes to migrate from "shortcut" scheme to "OOR" scheme_

TL;DR
- remove "reversible" vtxo script.
- allow sending async "pending" coins.
- remove `Vtxo` "pending" member in database, API, and domain.
- add "chain OOR transactions" e2e test case.

### SDK APIs changes

- add `CoinSelectOptions` parameters in every method using the coin selector. Allow more granular sendAsync, sendOffchain and redeem.
- remove "Claim" (duplicate of `SendOffChain(ctx, nil, nil)`).
- boarding utxos can be selected in `CollaborativeRedeem` and `SendOffChain` methods.

@altafan @tiero @bordalix @sekulicd please review


api-spec/protobuf/ark/v1/service.proto
```diff
@@ -318,10 +318,9 @@ message Vtxo {
   string spent_by = 4;
   int64 expire_at = 5;
   bool swept = 6;
-  bool pending = 7;
-  string redeem_tx = 8;
-  uint64 amount = 9;
-  string pubkey = 10;
+  uint64 amount = 7;
+  string pubkey = 8;
+  optional string redeem_tx = 9;
```
altafan: please revert these changes

pkg/client-sdk/example/covenantless/alice_to_bob.go
```diff
@@ -85,7 +85,7 @@ func main() {
 	log.Infof("alice offchain balance: %d", aliceBalance.OffchainBalance.Total)
 
 	log.Infof("alice claiming onboarding funds...")
-	txid, err := aliceArkClient.Claim(ctx)
+	txid, err := aliceArkClient.SendOffChain(ctx, nil, nil)
```
sekulicd: This is bit confusing, in CLI we use SendAsync for covenantless, why do we use SendOffChain in covenantless example? IMO there should be Send and Claim, SendOffChain with receivers=nil, for self send, doesnt seem like best UX.

client/main.go
```diff
@@ -275,7 +275,7 @@ func claim(ctx *cli.Context) error {
 		return err
 	}
 
-	txID, err := arkSdkClient.Claim(ctx.Context)
+	txID, err := arkSdkClient.SendOffChain(ctx.Context, nil, nil)
```
sekulicd: we kept CLI cmd name ‘claim’, although we we removed it from SDK and referenced it in cmd(SendOffchain).

pkg/client-sdk/example/covenantless/alice_to_bob.go
louisinger: I agree but in my mind the SDK interface should be as minimal as possible. Nothing prevents the developer from creating a more user friendly "Claim" function on top of sendOffchain

client/main.go
louisinger: right but do we really need that claim ? (self send ALL). Now with OOR there is no need to claim as soon as possible. Maybe this "claim" should be replaced by a more granular "settle" cmd ?

tiero: +1 for settle cmd

altafan (approved):

PR #363, opened 2024-10-31
[CLI] add `--rest` optional flag `init` command

louisinger:
This PR allows using the CLI with the REST client instead of gRPC

--rest optional flag will set arksdk.RestClient in init method.
```
ark init --rest .....
```

@altafan please review

### DIFF
--- a/client/main.go
+++ b/client/main.go
@@ -127,6 +127,12 @@ var (
 		Name:  "force",
 		Usage: "force redemption without collaboration",
 	}
+	restFlag = &cli.BoolFlag{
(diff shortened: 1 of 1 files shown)


tiero:
it makes sense to be global like `ark --rest <command>` no?

louisinger:
> it makes sense to be global like `ark --rest <command>` no?

yes but it needs more changes cause the client type is persisted in SDK state. This is just a quick fix to easy switch between grpc and rest

altafan (approved):

PR #365, opened 2024-11-01
[SDK] wallet init accepts nsec private key format

louisinger:
This PR allows the `singlekeywallet` init method to accept nsec-encoded secret keys. 
 
@tiero please review


tiero (commented):
NACK. Don't think should be part of the SDK rather just clients choice

PR #369, opened 2024-11-06
fix: make VTXO scripts miniscript compatible

bonomat:
miniscript uses `OP_CSV OP_VERIFY` instead of `OP_CSV OP_DROP`. This seems to be a miniscript specific rule. I couldn't find the exact reason for this. The only documentation was [this comment](https://arklabshq.slack.com/archives/C07TWFTS4P9/p1730886013553069?thread_ts=1730866846.191849&cid=C07TWFTS4P9). With this change we were able to make the asp compatible with the following miniscript on the client side: `tr(UNSPENDABLE_KEY,{and_v(v:pk(ASP),pk(USER_1)),and_v(v:older(TIMEOUT),pk(USER_0))})`

BREAKING CHANGE: `OP_VERIFY` is now used instead of `OP_DROP` in all VTXO scripts.

### DIFF
--- a/common/bitcointree/script.go
+++ b/common/bitcointree/script.go
@@ -103,7 +103,7 @@ func (d *CSVSigClosure) Leaf() (*txscript.TapLeaf, error) {
 
 func (d *CSVSigClosure) Decode(script []byte) (bool, error) {
 	csvIndex := bytes.Index(
-		script, []byte{txscript.OP_CHECKSEQUENCEVERIFY, txscript.OP_DROP},
+		script, []byte{txscript.OP_CHECKSEQUENCEVERIFY, txscript.OP_VERIFY},
 	)
 	if csvIndex == -1 || csvIndex == 0 {
 		return false, nil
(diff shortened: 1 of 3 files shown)


louisinger (approved):
ACK

lordlendalot:
Thank you for having a look at this. 
We had some more thoughts on this: 
miniscript might not be the best approach if the goal is to support arbitrary scripts in Ark. 

For example, if you were to build lightning "on Ark" you couldn't use miniscript because the locking script also makes use of `OP_DROP`: https://github.com/lightning/bolts/blob/master/03-transactions.md#to_local-output

bonomat:
Thank you for having a look at this. 
We had some more thoughts on this: 
miniscript might not be the best approach if the goal is to support arbitrary scripts in Ark. 

For example, if you were to build lightning "on Ark" you couldn't use miniscript because the locking script also makes use of `OP_DROP`: https://github.com/lightning/bolts/blob/master/03-transactions.md#to_local-output

altafan:
Even though the changes have been approved, seems better to close this instead of merging?

PR #371, opened 2024-11-07
Simulation framework

sekulicd:
This introduces a simulation framework designed to test the server by simulating multiple clients performing various actions over several rounds. The framework reads simulation configurations from YAML files, validates them against a predefined schema, and executes the simulation accordingly.
This contains code for local and remote(AWS) simulation, check README.

Few issues were fixed while testing (af492a79, 28116d05) @louisinger take a look.
Also there is separate [PR](https://github.com/ark-network/ark/pull/372) for handling panic recovery in go routines started before server startup.

Note that AWS setup doesnt work always for many users (100) as there is rate limit for creating many ECS tasks in parallel (for bigger setup AWS support needs to be contacted to increase rate limit).
This should be enhanced with Signet support.
Prev PR was closed cause it had logs of commits and resolving conflicts while rebasing was complicated.


sekulicd:
@tiero @altafan this now has support for signed, considering it contains two bug fixes i think it should be merged to master asap.

tiero:
>  two bug fixes

worth cherry-picking them in a dedicated PR, so we dont hold them for the simulation to be reviewed?

PR #372, opened 2024-11-07
Panic recovery

### LINKED ISSUE #360: Handle panic in Go routines
While testing simulation PR i noticed that test fails for 5 users claiming in same round.
In logs i could see error "missing list of signed forfeit txs” in finalizeFound. 
It was not obvious to find out root cause of this error as last log in startFinalization was "nonces collected for round”.
Was not able to figure it out how finalizeFound was invoked if startFinalization was not finished.
After some time by dumping go routines stack trace i saw that there was a panic in musig2.AggregateNonces which was ignored as there are no panic recovery for this go routine.
So code panicked, startFinalization exited and its defer func was called which continued with finalizeFound func invocation.

We have panic interceptor on server level but we need to handle panic in app service as this go routine is invoked before server is started.

```
func (s *service) start(withAppSvc bool) error {
	tlsConfig, err := s.config.tlsConfig()
	if err != nil {
		return err
	}

	if err := s.newServer(tlsConfig, withAppSvc); err != nil {
		return err
	}

	if withAppSvc {
		appSvc, _ := s.appConfig.AppService()
		if err := appSvc.Start(); err != nil {
			return fmt.Errorf("failed to start app service: %s", err)
		}
		log.Info("started app service")
	}

	if s.config.insecure() {
		// nolint:all
		go s.server.ListenAndServe()
	} else {
		// nolint:all
		go s.server.ListenAndServeTLS("", "")
	}
	log.Infof("started listening at %s", s.config.address())

	return nil
}
```

Check as well in SDK or elsewhere if there are go routines without panic recovery.

sekulicd:
This adds panic recovery in applicaition layer for methods that starts before grpc server, which therefore are not covered with gprc interceptor panic recovery middleware.

This closes #360


louisinger (changes requested):
you should add .idea folder into gitignore

louisinger (approved):

PR #374, opened 2024-11-07
[SDK] Improve REST JSON streaming

altafan:
This contains the following fixes:
- Update protos so that the rest endpoint of GetEventStream doesn't go in conflict with the one of GetRound: `/v1/round/events` > `/v1/events`
- Update protos to drop event message in ping response
- Do not close GetEventStream server side after round failed/finalized event is sent. Allows clients that start listening to an already ongoing round wait for the next one
- Make the SDK's REST client use the stream GetEventStream endpoint instead of polling Ping
- Fix SDK tx history api
- Renamings in SDK

Please @tiero @louisinger review this.


server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -528,7 +461,7 @@ func (h *handler) listenToEvents() {
 				},
 			}
 		case domain.RoundFinalized:
-			shouldClose = true
+			// shouldClose = true
```
tiero: remove commented code?

tiero:
CI fails

tiero:
Use a more detailed title

louisinger (approved):

PR #375, opened 2024-11-11
PaymentMap pop func input traversal & nonce encode/decode

sekulicd:
This contains fixes cherry picked from [PR](https://github.com/ark-network/ark/pull/371).


server/cmd/arkd/main.go
```diff
@@ -88,6 +88,8 @@ func mainAction(_ *cli.Context) error {
 		return err
 	}
 
+	log.Infof("ASP config: %+v", appConfig)
```
tiero:
```suggestion
	log.Infof("Ark Server config: %+v", appConfig)
```

sekulicd: Done

common/bitcointree/musig2.go
```diff
@@ -585,57 +586,61 @@ type readable interface {
...
+			cell := factory()
+			if err := cell.Decode(data); err != nil {
+				return nil, err
 			}
-			return nil, err
+			row[j] = cell
 		}
-		row = append(row, cell)
+
+		matrix[i] = row
```
altafan: Please, use append instead of assingment to populate slice

```suggestion
	matrix := make([][]T, 0, rowCount)

	// For each row, read its length and then its elements
	for i := uint32(0); i < rowCount; i++ {
		var colCount uint32
		// Read row length
		if err := binary.Read(data, binary.LittleEndian, &colCount); err != nil {
			return nil, err
		}

		// Initialize row
		row := make([]T, 0, colCount)

		// Read row data
		for j := uint32(0); j < colCount; j++ {
			cell := factory()
			if err := cell.Decode(data); err != nil {
				return nil, err
			}
			row = append(row, cell)
		}

		matrix = append(matrix, row)
```

sekulicd: done

tiero:
more eloquent PR title

sekulicd:
> more eloquent PR title

Done

PR #380, opened 2024-11-15
[Market Hour] Update GetInfo & Add UpdateMarketHour RPC

sekulicd:
This PR introduces initial changes to support market hour configuration that will help users determine optimal times for making payments with lower fees.

### Update GetInfo RPC
- Update GetInfo RPC with market hour config params

### Market Hour Admin RPC
- Add UpdateMarketHour RPC to AdminService

### Market Hour Persistence
- Add SQLite-based persistence for market hour configuration
- Load market hours from DB on service startup
- Fallback to config values if no DB entry exists

## Configuration Parameters
- `first_market_hour`: Unix timestamp for initial market hour
- `period`: Time interval between market hours (in seconds)
- `round_lifetime`: Round lifetime for market hours (optional, defaults to round lifetime)


api-spec/protobuf/ark/v1/admin.proto
```diff
@@ -63,4 +69,13 @@ message GetRoundsRequest {
...
 message GetRoundsResponse {
   repeated string rounds = 1;
-}
\ No newline at end of file
+}
+
+message UpdateMarketHourRequest {
+  int64 first_market_hour = 1;
+  int64 period = 2;
+  int64 round_lifetime = 3;
```
altafan, line 77:
```suggestion
  int64 round_interval = 3;
```

altafan: We may want to specify market hours longer or shorter than an hour so let's specify the start and end time of the first market hours, the period and the round interval (not filefitme) during the market hours

```suggestion
  // Starting timestamp of the first market hour.
  int64 start_time = 1;
  // Ending timestamp of the first market hour.
  int64 end_time = 2;
  // Period expressed in hours between market hours 
  int64 period = 3;
  // Round interval during the market hours, this overrides the default one.
  int64 round_interval = 4;
```

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -336,4 +337,10 @@ message RedeemTransaction {
   string txid = 1;
   repeated Outpoint spent_vtxos = 2;
   repeated Vtxo spendable_vtxos = 3;
+}
+
+message MarketHour {
+  int64 first_market_hour = 1;  // Unix time representing the first market hour
+  int64 period = 2;             // Period between market hours
+  int64 round_lifetime = 3;     // Round lifetime in seconds
```
altafan:
```suggestion
  // Starting timestamp of the next market hour.
  int64 next_market_hour_start = 1; 
  // Ending timestamp of the next market hour.
  int64 next_market_hour_end = 2;
  // Period between market hours expressed in hours.
  int64 period = 3;
  // Round interval in seconds during market hours.
  int64 round_interval = 4;
```

api-spec/protobuf/ark/v1/admin.proto
altafan: please reflect these changes also in env vars

api-spec/protobuf/ark/v1/service.proto
sekulicd: Based on [issue](https://github.com/ark-network/ark/issues/297) desc i understood ASP sets timestamp of first marked hour and period, so that user invokes GetInfo and can always calculate when next round happens.
With this suggestion it seems to me that user will need frequently to call GetInfo. @altafan

server/internal/core/domain/market_hour.go
```diff
@@ -0,0 +1,9 @@
+package domain
+
+type MarketHour struct {
+	ID              int64
```
altafan: not needed

server/internal/core/domain/market_hour.go
```diff
@@ -0,0 +1,9 @@
...
+	FirstMarketHour int64
+	Period          int64
+	RoundLifetime   int64
+	CreatedAt       int64
```
altafan: not needed

altafan: this domain is just for persisting some settings, we will update record in db when updating, we don't really need to keep track of the changes made to this domain.

api-spec/protobuf/ark/v1/service.proto
sekulicd: @altafan

altafan: mmmm, this is not what i have in mind exactly. When a client calls get info he knows when the next market hour is intea of the very first one (likely in the past), from that moment, he uses the `period` to know when the next one will be.

Let's assume the server changes config, the clients will receive a notification to fetch get info and update their data about market hours. Added this to complete the reasoning but this last thing is not required to be implemented in this pr.

server/internal/core/domain/market_hour.go
tiero: I think it's valuable for clients (maybe another domain) to know which was the `upatedAt` moment, to safely know when to update his own cached settings.

api-spec/protobuf/ark/v1/service.proto
sekulicd: @altafan in GetInfo should server calc next start_time (end_time) based on first_start_time, time.now() and round_interval?

server/internal/core/domain/market_hour.go
altafan: ok let's make this field become `UdatedAt` then?
```suggestion
	UpdatedAt       int64
```

api-spec/protobuf/ark/v1/service.proto
altafan: indeed, based on first start time, time.now() and period (not round interval, not needed)

server/internal/config/config.go
```diff
@@ -5,40 +5,44 @@ import (
...
+	BitcoindRpcPass         string
+	BitcoindRpcHost         string
+	TLSExtraIPs             []string
+	TLSExtraDomains         []string
+	UnlockerType            string
+	UnlockerFilePath        string
+	UnlockerPassword        string
+	FirstMarketHour         int64
+	MarketHourPeriod        int64
+	MarketHourRoundLifetime int64
```
tiero: What's this? Also, _RoundLifetime_ is not correct (where used elsewhere has to be replaced)

api-spec/protobuf/ark/v1/service.proto
sekulicd: i was thinking to use round interval to calculate end_time @altafan, ok?

sekulicd: should we have end_time at all considering start_time + round_interval is enough? @altafan

bordalix: Why not simplify to:
- next_market_hour_unixtime: for obvious reasons
- period_between_market_hours: to find out when will happen next market hours and how frequently they happen
- duration_of_market_hour: to know for how long will I need to be online to participate in this round

server/internal/core/domain/market_hour.go
sekulicd: Done

server/internal/core/domain/market_hour.go
sekulicd: Done

server/internal/config/config.go
sekulicd: Done

api-spec/protobuf/ark/v1/service.proto
sekulicd: isnt duration_of_market_hour same as round_interval? @bordalix

altafan: @sekulicd no, the round interval is the duration between rounds, if 10 for example, it means that 10s last between a round and the next one, ie. that's how much it takes a round to complete eventually.

server/internal/app-config/config.go
```diff
@@ -54,21 +54,24 @@ var (
...
+	Network                 common.Network
+	SchedulerType           string
+	TxBuilderType           string
+	WalletAddr              string
+	RoundLifetime           int64
+	UnilateralExitDelay     int64
+	BoardingExitDelay       int64
+	NostrDefaultRelays      []string
+	NoteUriPrefix           string
+	MarketStartTime         int64
```
altafan: Let's keep the same prefix for consitency:
```suggestion
	MarketHourStartTime         int64
```

server/internal/config/config.go
```diff
@@ -77,6 +81,9 @@ var (
 	UnlockerFilePath = "UNLOCKER_FILE_PATH"
 	UnlockerPassword = "UNLOCKER_PASSWORD"
 	NoteUriPrefix    = "NOTE_URI_PREFIX"
+	MarketStartTime         = "MARKET_START_TIME"
```
altafan: let's keep the same prefix for consitency:

```suggestion
	MarketHourStartTime         = "MARKET_HOUR_START_TIME"
```

server/internal/core/application/covenant.go
```diff
@@ -423,6 +447,11 @@ func (s *covenantService) GetInfo(ctx context.Context) (*ServiceInfo, error) {
 		return nil, err
 	}
 
+	marketNextStart := s.marketStartTime
+	if s.marketStartTime < time.Now().Unix() {
+		marketNextStart = s.marketStartTime + s.marketHourPeriod
+	}
```
altafan: as far as i understood, `s.marketStartTime` is the very first market hour's starting time so this should be put in a loop as now it returns always the second market hour, which is not always the next ofc

server/internal/core/application/covenant.go
```diff
@@ -30,12 +30,15 @@ var (
...
-	boardingExitDelay   int64
+	network                 common.Network
+	pubkey                  *secp256k1.PublicKey
+	roundLifetime           int64
+	roundInterval           int64
+	unilateralExitDelay     int64
+	boardingExitDelay       int64
+	marketStartTime         int64
+	marketHourPeriod        int64
+	marketHourRoundInterval int64
```
altafan: maybe better to keep in memory a `domain.MarketHour`?

server/internal/core/application/types.go
```diff
@@ -70,6 +71,13 @@ type ServiceInfo struct {
 	Dust                       uint64
 	BoardingDescriptorTemplate string
 	ForfeitAddress             string
+	MarketHour                 *NextMarketHour
```
altafan: this struct isn't really needed to me, we can use a `domain.MarketHour`?

server/internal/infrastructure/db/badger/market_hour_repo.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to open market hour store: %s", err)
+	}
+
+	return &marketHourRepository{store}, nil
+}
+
+func (r *marketHourRepository) Get(ctx context.Context) (*domain.MarketHour, error) {
+	var marketHour domain.MarketHour
+	err := r.store.Get("market_hour", &marketHour)
```
altafan, line 53: please put `"market_hour"` in a constant like `marketHourKey` or something like that

server/internal/core/application/types.go
sekulicd: MarketHour is domain what is stored in DB, NextMarketHour(contains NextStartTime) is calculated based on MarketHour, imo its logic is more clear with this.

server/internal/core/application/covenantless_test.go
```diff
@@ -0,0 +1,130 @@
...
+			now:           parseTime("2023-10-10 13:00:00"),
+			expectedStart: parseTime("2023-10-10 13:00:00"),
+			expectedEnd:   parseTime("2023-10-10 13:05:00"),
+			expectError:   false,
+			description:   "Now is exactly at the initial market hour start time",
+		},
+		{
+			now:           parseTime("2023-10-10 13:02:00"),
+			expectedStart: parseTime("2023-10-10 16:00:00"),
+			expectedEnd:   parseTime("2023-10-10 16:05:00"),
```
altafan: in this case we should return the current market hour not the next one

server/internal/core/application/covenantless_test.go
```diff
@@ -0,0 +1,130 @@
...
+			now:           parseTime("2023-10-10 22:01:00"),
+			expectedStart: parseTime("2023-10-11 01:00:00"),
+			expectedEnd:   parseTime("2023-10-11 01:05:00"),
+			expectError:   false,
+			description:   "Now is much later",
+		},
+		{
+			now:           parseTime("2023-10-10 16:01:00"),
+			expectedStart: parseTime("2023-10-10 19:00:00"),
+			expectedEnd:   parseTime("2023-10-10 19:05:00"),
```
altafan: same here

server/internal/core/application/covenantless_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Helper function to parse time strings in UTC
+func parseTime(value string) time.Time {
+	layout := "2006-01-02 15:04:05"
+	t, err := time.ParseInLocation(layout, value, time.UTC)
+	if err != nil {
+		panic(err)
+	}
+	return t
+}
```
altafan: let's not panic, instead let's do this:
```suggestion
func parseTime(value string) (t time.Time, err error) {
	layout := "2006-01-02 15:04:05"
	t, err = time.ParseInLocation(layout, value, time.UTC)
	return
}
```

then we do:
```
marketHourtStartTime, err := parseTime("...")
require.NoError(t, err)
```

Please use require instead of raising errors manually in all tests of this file

server/internal/infrastructure/db/service_test.go
```diff
@@ -505,6 +504,46 @@ func testEntityRepository(t *testing.T, svc ports.RepoManager) {
 	})
 }
 
+func testMarketHourRepository(t *testing.T, svc ports.RepoManager) {
+	t.Run("test_entity_repository", func(t *testing.T) {
```
altafan:
```suggestion
	t.Run("test_market_hour_repository", func(t *testing.T) {
```

server/internal/core/application/covenantless_test.go
sekulicd: Done

server/internal/core/application/covenantless_test.go
sekulicd: Are we sure about this? if user invokes GetInfo in the middle of MarketHour isnt it better to return next market hour?

altafan (approved):

PR #382, opened 2024-11-18
Rework forfeit transactions validation

louisinger:
_This PR changes the verification process of the forfeit transaction made by the server. It allows the client to compute the forfeit tx fee amount._

Instead of building every forfeit transactions before the round finalization, the server waits to receive the signed forfeit tx to verify it. The server verifies the fee amount is not lower than the min-relay-fee and do not exceed 105% of the min-relay-fee.  It has several advantages:

* do not "freeze" the fees amount, making the estimation more flexible. 
* make the fee estimation more accurate because the user knows the exact size of the tapscript control block.
* the server does not have to build every forfeit transactions before sending the vtxo tree to round participants. This makes round creation quicker.
* the server doesn't keep descriptors in memory during the round creation.

@tiero @altafan please review


server/internal/core/application/utils.go
```diff
@@ -250,73 +238,97 @@ func (m *paymentsMap) view(id string) (*domain.Payment, bool) {
...
+
+	// verify the signatures
+	for _, txs := range validTxs {
+		for _, tx := range txs {
+			valid, txid, err := m.builder.VerifyTapscriptPartialSigs(tx)
+			if err != nil {
+				return err
+			}
+			if !valid {
+				return fmt.Errorf("invalid forfeit tx signature (%s)", txid)
```
altafan: move this into `m.builder.VerifyForfeitTxs`

PR #384, opened 2024-11-19
Agnostic `VtxoScript` + migrate from descriptor to script array

louisinger:
### Replace descriptor by tapscripts array

Since descriptors don't yet support timelocks, we are currently using a custom descriptor lib maintained by us. This is cool but the package is basically just a list of supported descriptor templates, not something that is general-purpose and ready-to-use and requires us to update this list for any custom script created by the client, while we designed the Server to support any kind of script.

Therefore, instead of using our custom `descriptor` package to represent a taproot tree, we now use a list of hex-encoded raw taproot tree leaves, allowing any custom script client-side without having to touch anything server-side.

### Agnostic `VtxoScript`

* `DefaultVtxoScript` has been replaced by `TapscriptsVtxoScript` accepting a set of taproot scripts instead of hardcoding the script types.
* `MultisigClosure` and `CSVSigClosure` now accept a list of keys instead of a hardcoded number of keys. It allows building custom forfeit and exit closures.
* The server will consider any VTXO script as valid if and only if : (1) all forfeits closures contain at least the Server public key and (2) there is no exit closure where the exit delay is smaller than `ARK_EXIT_DELAY`.

@altafan @tiero please review


api-spec/protobuf/ark/v1/service.proto
```diff
@@ -123,17 +123,15 @@ message GetInfoResponse {
   int64 round_interval = 4;
   string network = 5;
   int64 dust = 6;
-  string boarding_descriptor_template = 7;
-  repeated string vtxo_descriptor_templates = 8;
-  string forfeit_address = 9;
+  string forfeit_address = 7;
```
altafan: let's revert these changes please

common/vtxo.go
```diff
@@ -30,14 +31,15 @@ It may also contain others closures implementing specific use cases.
 
 VtxoScript abstracts the taproot complexity behind vtxo contracts.
 it is compiled, transferred and parsed using descriptor string.
-
-default vtxo script = tr(_,{ and(pk(USER), pk(ASP)), and(older(T), pk(USER)) })
-reversible vtxo script = tr(_,{ { and(pk(SENDER), pk(ASP)), and(older(T), pk(SENDER)) }, { and(pk(RECEIVER), pk(ASP) } })
 */
-type VtxoScript[T TaprootTree] interface {
+type VtxoScript[T TaprootTree, F interface{}, E interface{}] interface {
+	Validate(asp *secp256k1.PublicKey, minExitDelay uint) error
```
tiero:
```suggestion
	Validate(server *secp256k1.PublicKey, minExitDelay uint) error
```

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -123,17 +123,15 @@ message GetInfoResponse {
...
 }
 
 message GetBoardingAddressRequest {
   string pubkey = 1;
 }
 message GetBoardingAddressResponse {
   string address = 1;
-  string descriptor = 2;
+  repeated string tapscripts = 2;
```
tiero: I would rather add a filed so we retain possibility in the future to also return descriptors

altafan: indeed

altafan (approved):

PR #388, opened 2024-11-22
Move OOR transaction creation client-side

louisinger:
This PR removes `CreateAsyncPayment` API and replace by `BuildOORTransaction` in the SDK. 

Now the clients craft the psbt client-side, then sign it and sent it to the server using `CompletePayment` API. If valid, the server signs and send the signed tx back the client.

@altafan please review


server/internal/core/application/covenantless.go
```diff
@@ -170,116 +169,192 @@ func (s *covenantlessService) Stop() {
...
-			if len(input.TaprootScriptSpendSig) == 0 {
-				return fmt.Errorf("missing tapscript spend sig")
-			}
+		vtxoOutpoint := ptx.UnsignedTx.TxIn[inputIndex].PreviousOutPoint
 
-			vtxoOutpoint := ptx.UnsignedTx.TxIn[inputIndex].PreviousOutPoint
+		// verify that the vtxo is spendable
 
-			// verify that the vtxo is spendable
+		vtxos, err := vtxoRepo.GetVtxos(ctx, []domain.VtxoKey{{Txid: vtxoOutpoint.Hash.String(), VOut: vtxoOutpoint.Index}})
```
altafan: this call should be done only once, for all vtxos we have to fetch. It's a waste of resources making a single call to the db for each tx input.

PR #389, opened 2024-11-27
Support forfeit with `CHECKLOCKTIMEVERIFY`

louisinger:
_This PR adds the `CLTVMultisigClosure` that can be used as a valid forfeit path in VtxoScript._

A CLTV forfeit is **invalid until an absolute locktime** (block height or block time). The server won't accept any forfeit or redeem transaction using a locked CLTV path.  Tested by `TestSendToCLTVMultisigClosure`.

@tiero @altafan please review


server/internal/infrastructure/wallet/liquid-standalone/service.go
```diff
@@ -168,6 +173,35 @@ func (s *service) VerifyMessageSignature(ctx context.Context, message, signature
 	return false, errors.New("not implemented")
 }
 
+func (s *service) GetCurrentBlockTime(ctx context.Context) (*ports.BlockTimestamp, error) {
+	tipURL, err := url.JoinPath(s.esploraURL, "blocks/tip/height")
```
altafan: call `block/tips/hash` to get the hash of the latest block, then call `block/:hash` to get the object that includes height and timestamp (counts for block timestamp)

altafan (approved):

PR #391, opened 2024-12-02
Consolidate Duplicate Server appConfig and Config Structure

### LINKED ISSUE #390: Merge `app-config` and `config` packages
The [app-config](https://github.com/ark-network/ark/blob/master/server/internal/app-config/config.go) pkg and the [config](https://github.com/ark-network/ark/blob/master/server/internal/config/config.go) pkg are using a duplicated "Config" struct, making it annoying to update the config variables. 

Let's drop `app-config` and merge it into `config` package.

aruokhai:
### Objective
A duplicate configuration struct in the Server directory complicates coordination. By consolidating it into a single Config structure, the configuration flow becomes more streamlined and efficient.

### Scope
Non-critical, as it only affects Config parsing and usage.

closes #390


common/bip68.go
```diff
@@ -32,6 +32,14 @@ type Locktime struct {
 	Value uint32
 }
 
+func FromConfig(locktime int64) Locktime {
```
louisinger: I'd move this function in `config` pkg as it is only used there. No need to be exported too.

altafan: as per louis suggestion, let's drop this new function and move it to config

altafan: ping @aruokhai

louisinger (approved):

server/internal/config/config.go
```diff
@@ -216,3 +272,343 @@ func getNetwork() (common.Network, error) {
 		return common.Network{}, fmt.Errorf("unknown network %s", viper.GetString(Network))
 	}
 }
+
+func DetermineLocktimeType(locktime int64) common.Locktime {
+	if locktime >= 512 {
```
bordalix:
```suggestion
	if locktime >= 500000000 {
```
@louisinger ?

server/internal/config/config.go
```diff
@@ -216,3 +272,343 @@ func getNetwork() (common.Network, error) {
...
+				minAllowedSequence, c.RoundLifetime,
+			)
+		}
+	}
+
+	if c.UnilateralExitDelay.Type == common.LocktimeTypeBlock {
+		return fmt.Errorf(
+			"invalid unilateral exit delay, must at least %d", minAllowedSequence,
+		)
+	}
```
bordalix, line 348: I don't understand the purpose of this check.

server/internal/config/config.go
```diff
@@ -216,3 +272,343 @@ func getNetwork() (common.Network, error) {
...
+
+	if c.BoardingExitDelay.Type == common.LocktimeTypeBlock {
+		return fmt.Errorf(
+			"invalid boarding exit delay, must at least %d", minAllowedSequence,
+		)
+	}
```
bordalix, line 354: same here

altafan (approved):

server/internal/config/config.go
altafan: the check is inherited from master.. atm, the locktime type is determined by its value: if it's below the min locktime in seconds (512), it's considered a "block" one

server/internal/config/config.go
altafan: these are all checks we make in master, and yes, they look a bit weird. I'll review them and eventually open a dedicated pr.

server/internal/config/config.go
```diff
@@ -216,3 +272,343 @@ func getNetwork() (common.Network, error) {
...
+		if c.RoundLifetime.Value%minAllowedSequence != 0 {
+			c.RoundLifetime.Value -= c.RoundLifetime.Value % minAllowedSequence
+			log.Infof(
+				"round lifetime must be a multiple of %d, rounded to %d",
+				minAllowedSequence, c.RoundLifetime,
+			)
+		}
+	}
+
+	if c.UnilateralExitDelay.Type == common.LocktimeTypeBlock {
```
bordalix, line 344: Sorry guys, but I must insist on this.

If you look at code, and you pass to config a UnilateralExitDelay of 666, from `func DetermineLocktimeType()` the `c.UnilateralExitDelay.Type` will be `common.LocktimeTypeBlock` so `Validate()` will always throw here.

This means we will never be able to pass a value >= 512 to UnilateralExitDelay.

Am I wrong?

server/internal/config/config.go
```diff
@@ -216,3 +272,343 @@ func getNetwork() (common.Network, error) {
...
+		return fmt.Errorf(
+			"invalid unilateral exit delay, must at least %d", minAllowedSequence,
+		)
+	}
+
+	if c.BoardingExitDelay.Type == common.LocktimeTypeBlock {
```
bordalix, line 350: Same here, but with BoardingExitDelay

server/internal/config/config.go
louisinger: no if you pass 666, it will be read as `common.LocktimeTypeSecond`.

That's the expected behaviour. We don't have a flag to signal what type of lockime we want to use, so we are using the value itself:
* if `< 512` = we admit it's a "block number" locktime
* if `> 512` = we admit it's a "second" locktime

server/internal/config/config.go
```diff
@@ -216,3 +272,343 @@ func getNetwork() (common.Network, error) {
 		return common.Network{}, fmt.Errorf("unknown network %s", viper.GetString(Network))
 	}
 }
+
+func DetermineLocktimeType(locktime int64) common.Locktime {
```
louisinger:
```suggestion
func determineLocktimeType(locktime int64) common.Locktime {
```

no need to be exported, right ? cc @altafan

louisinger (approved):

server/internal/config/config.go
```diff
@@ -216,3 +272,343 @@ func getNetwork() (common.Network, error) {
 		return common.Network{}, fmt.Errorf("unknown network %s", viper.GetString(Network))
 	}
 }
+
+func DetermineLocktimeType(locktime int64) common.RelativeLocktime {
```
louisinger: Should not be exported :
```suggestion
func determineLocktimeType(locktime int64) common.RelativeLocktime {
```

louisinger (changes requested):

server/internal/config/config.go
louisinger: ok I've re-read, now got it. "block number" locktime type is only for **lifetime**, we disallow exit and boarding delays to be set as block. That's why the "minimum allowed value" is 512 for boarding and exit.

We should maybe allow block number to be set also for boarding and exit but I confirm, the checks are OK and fit the ones in master.

altafan:
@aruokhai can you please sync with master and include the changes we made to config in #417 to yours?

louisinger (approved):

altafan:
@aruokhai please sync again with master and include the changes made to config/app-config 🙏

PR #397, opened 2024-12-09
[arksdk] add `ArkClient.SignTransaction`

louisinger:
This PR adds a wrapper of `WalletService.SignTransaction` in client interface. 

@altafan please review


tiero:
Worth exposing it to wasm binary also?

altafan (approved):

PR #398, opened 2024-12-09
sqlite: Split `SelectSweepableRounds` into several smaller queries

louisinger:
Rework the `SelectSweepableRounds` SQL query. The current implementation was using a lot of LEFT JOIN making `GetSweepableRounds` stuck the main thread.

@altafan please review

### DIFF
--- a/server/internal/infrastructure/db/sqlite/round_repo.go
+++ b/server/internal/infrastructure/db/sqlite/round_repo.go
@@ -263,34 +263,76 @@ func (r *roundRepository) GetRoundWithTxid(ctx context.Context, txid string) (*d
 }
 
 func (r *roundRepository) GetSweepableRounds(ctx context.Context) ([]domain.Round, error) {
-	rows, err := r.querier.SelectSweepableRounds(ctx)
+	// First get the sweepable round IDs
+	ids, err := r.querier.SelectSweepableRoundIds(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	rvs := make([]combinedRow, 0, len(rows))
-	for _, row := range rows {
-		rvs = append(rvs, combinedRow{
-			round:    row.Round,
-			request:  row.RoundRequestVw,
-			tx:       row.RoundTxVw,
-			receiver: row.RequestReceiverVw,
-			vtxo:     row.RequestVtxoVw,
-		})
+	var rounds []domain.Round
+	for _, id := range ids {
+		// Get basic round data
+		round, err := r.querier.SelectRoundDataById(ctx, id)
+		if err != nil {
+			return nil, err
+		}
+
+		// Get requests if needed
+		requests, err := r.querier.SelectRoundRequestsById(ctx, sql.NullString{String: id, Valid: true})
+		if err != nil {
+			return nil, err
+		}
+
+		// Get transactions if needed
+		txs, err := r.querier.SelectRoundTxsById(ctx, sql.NullString{String: id, Valid: true})
+		if err != nil {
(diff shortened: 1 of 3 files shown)


sekulicd:
Although this probably improves current situation I would say NACK to it cause we need to revamp db layer a bit.
Here is a problem: 
Our domain Round is aggregating whole DB:
```
type Round struct {
	Id                string
	StartingTimestamp int64
	EndingTimestamp   int64
	Stage             Stage
	TxRequests        map[string]TxRequest
	Txid              string
	UnsignedTx        string
	ForfeitTxs        []string
	VtxoTree          tree.VtxoTree
	Connectors        []string
	ConnectorAddress  string
	DustAmount        uint64
	Version           uint
	Swept             bool // true if all the vtxos are vtxo.Swept or vtxo.Redeemed
	changes           []RoundEvent
}
```

Bellow is one of problematic queries and rows count in comment:
```
SELECT count(*)
FROM round
         LEFT OUTER JOIN round_request_vw ON round.id=round_request_vw.round_id
         LEFT OUTER JOIN round_tx_vw ON round.id=round_tx_vw.round_id
         LEFT OUTER JOIN request_receiver_vw ON round_request_vw.id=request_receiver_vw.request_id
         LEFT OUTER JOIN request_vtxo_vw ON round_request_vw.id=request_vtxo_vw.request_id
WHERE round.swept = false AND round.ended = true AND round.failed = false; --11 010 322

select count(*) from round; --147
select count(*) from round_request_vw; -- 2111
select count(*) from round_tx_vw; --88258
select count(*) from request_receiver_vw; --4125
select count(*) from request_vtxo_vw; --3249
```

SelectSweptRounds results in massive result set that takes more than 40 GB of RAM.

We need to be careful with what we fetch for db, and we should never fetch data to populate all Rounds with all its properties(basically whole DB).

For eg. bellow code needs sweepable tx ids:

```
func (s *covenantlessService) restoreWatchingVtxos() error {
	sweepableRounds, err := s.repoManager.Rounds().GetSweepableRounds(context.Background())
	if err != nil {
		return err
	}

	vtxos := make([]domain.Vtxo, 0)

	for _, round := range sweepableRounds {
		fromRound, err := s.repoManager.Vtxos().GetVtxosForRound(
			context.Background(), round.Txid,
		)
		if err != nil {
			log.WithError(err).Warnf("failed to retrieve vtxos for round %s", round.Txid)
			continue
		}
		for _, v := range fromRound {
			if !v.Swept && !v.Redeemed {
				vtxos = append(vtxos, v)
			}
		}
	}

	if len(vtxos) <= 0 {
		return nil
	}

	if err := s.startWatchingVtxos(vtxos); err != nil {
		return err
	}

	log.Debugf("restored watching %d vtxos", len(vtxos))
	return nil
}
```

We could use bellow query:
```
select count(*)
from round r inner join tx t on r.id = t.round_id 
WHERE r.swept = false AND r.ended = true AND r.failed = false; — 83507
```

louisinger:
100% we should rework the interface to return more granular data.
but this must be done for all methods I would say and it will require change in the application layer, this is an hotfix at least to make the server work before a better solution.

louisinger:
also [the method is also used to get VtxoTree](https://github.com/ark-network/ark/blob/master/server%2Finternal%2Fcore%2Fapplication%2Fsweeper.go) and I think the trees are what make the dataset so big. We can probably gain by only selecting the trees but I'd still split the queries into smaller ones instead of a lot of JOIN.

altafan:
closed in favor of #399

PR #400, opened 2024-12-12
FIx client build on windows

Kukks:
For some reason, it doesn't compile on windows without converting to int explicitly


bordalix:
I didn't know about this and I find it unbelievable:

In go, syscall.Stdin is a `int` in any other platform but a `handle` in Windows

https://github.com/golang/go/issues/66083

bordalix (approved):

altafan:
cc @Kukks.

This is going to be merged as soon as it's synced with master.

PR #403, opened 2024-12-16
Bump server btcd

Kukks:
Nigiri now pulls bitcoin core 28.0, which has an rpc breaking change on `getnetworkinfo`, causing arkd to fail on startup due to a mismatch on json with btcd. (didn't bump everywhere else since I'm not sure if that fits in your workflow)


altafan:
please @Kukks, `cd` to the root folder of the project and run `go work sync`

PR #405, opened 2024-12-17
Replace Manual Script Decoding and Verification with Tokenizer

### LINKED ISSUE #394: Re-implement `tree.DecodeClosure` using `txscript.Tokenizer`
The current implementation of `DecodeClosure` is using `bytes.LastIndex` to search some opcodes in the script to decode. However, in case this opcode byte is included in public key, it could lead to unexpected errors.

We should rework the decode algorithm using `txscript.NewTokenizer`

aruokhai:
### Objective
Decoding scripts manually using a cursor can be error-prone and inefficient. A more reliable approach is to use the **Btcd's** `script tokenizer` utility, which automatically splits OPCODEs into their respective segments

### Scope Of Change
Critical, as it involves the decoding of scripts, which are crucial to the operations of Ark .

closes #394

### DIFF
--- a/common/tree/script.go
+++ b/common/tree/script.go
@@ -44,7 +44,7 @@ var forbiddenOpcodes = []byte{
 
 type Closure interface {
 	Script() ([]byte, error)
-	Decode(script []byte) (bool, error)
+	Decode(script txscript.ScriptTokenizer) (bool, error)
 	// WitnessSize returns the size of the witness excluding the script and control block
 	// extraWitnessSize is here to count the condition witness size
 	// or any other witness size that can't be computed from the script
@@ -121,7 +121,7 @@ func DecodeClosure(script []byte) (Closure, error) {
 	for _, t := range types {
 		scriptCopy := make([]byte, len(script))
 		copy(scriptCopy, script)
-		valid, err := t.closure.Decode(scriptCopy)
+		valid, err := t.closure.Decode(txscript.MakeScriptTokenizer(0, scriptCopy))
 		if err != nil {
 			decodeErr = append(decodeErr, fmt.Sprintf("%s: %v", t.name, err))
 			continue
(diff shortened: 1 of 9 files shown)


common/tree/script.go
```diff
@@ -44,7 +44,7 @@ var forbiddenOpcodes = []byte{
 
 type Closure interface {
 	Script() ([]byte, error)
-	Decode(script []byte) (bool, error)
```
louisinger: I would keep the interface unchanged and instantiate ScriptTokenizer in the function

louisinger (changes requested):

common/tree/script.go
aruokhai: That would make it difficult to use the `cursor` of `Tokenizer` type, which is quite splendid if a Decode type calls another Decode Type.

common/tree/script.go
```diff
@@ -175,146 +175,152 @@ func (f *MultisigClosure) Script() ([]byte, error) {
 }
 
 func (f *MultisigClosure) Decode(script []byte) (bool, error) {
-	if len(script) == 0 {
```
louisinger, line 178: Can you revert this check ? to have a meaningful error in case someone pass a `nil` value as argument

common/tree/script.go
```diff
@@ -175,146 +175,152 @@ func (f *MultisigClosure) Script() ([]byte, error) {
...
-		return false, fmt.Errorf("empty script")
+	tokenizer := txscript.MakeScriptTokenizer(0, script)
+
+	pubkeys := make([]*secp256k1.PublicKey, 0)
```
louisinger: I understand you're using this array to parse the first public key and then use the next opcode to parse either a CHECKSIGADD or CHECKSIG script type.

I like the idea, that's a nice improvement ! But it seems we can simplify this by only checking the opcode at `script[1+32]` ?
*  if it's CHECKSIGADD -> use `completeDecodeChecksigAdd`
* if it's CHECKSIGVERIFY or CHECKSIG -> use `completeDecodeChecksig`
* else `return false, nil`

in other words, not sure if we need the tokenizer here. We could make a static check on the 33th byte of the script and move `pubkeys` and `tokenizer` complexity in the private functions.

Maybe I'm missing something, open to suggestions, what do you think ?

common/tree/script.go
```diff
@@ -414,30 +406,39 @@ func (d *CSVMultisigClosure) Script() ([]byte, error) {
...
-	)
-	if csvIndex == -1 || csvIndex == 0 {
+	if tokenizer.Err() != nil {
 		return false, nil
 	}
 
-	sequence := script[:csvIndex]
-	if len(sequence) > 5 {
-		return false, fmt.Errorf("invalid sequence length")
+	_ = tokenizer.Next()
```
louisinger: Why do we "skip" a token here ? Maybe worth to put a small comment explaining this

common/tree/script.go
```diff
@@ -780,22 +781,27 @@ func (d *CLTVMultisigClosure) Script() ([]byte, error) {
...
+	if len(tokenizer.Script()) == 0 {
 		return false, fmt.Errorf("empty script")
 	}
 
-	cltvIndex := bytes.Index(
-		script, []byte{txscript.OP_CHECKLOCKTIMEVERIFY, txscript.OP_DROP},
-	)
-	if cltvIndex == -1 || cltvIndex == 0 {
-		return false, nil
+	_ = tokenizer.Next()
```
louisinger: same here, maybe a comment ?

common/tree/script.go
```diff
@@ -1024,30 +1026,28 @@ func (f *ConditionCSVMultisigClosure) Script() ([]byte, error) {
...
-	if len(script) == 0 {
+	tokenizer := txscript.MakeScriptTokenizer(0, script)
+
+	if len(tokenizer.Script()) == 0 {
 		return false, fmt.Errorf("empty script")
 	}
 
-	tokenizer := txscript.MakeScriptTokenizer(0, script)
-	verifyPositions := []int{}
+	condition := []byte{}
```
louisinger:
```suggestion
	condition := make([]byte, 0)
```

common/tree/script_test.go
```diff
@@ -461,7 +461,7 @@ func TestMultisigClosure(t *testing.T) {
 	t.Run("invalid empty script", func(t *testing.T) {
 		closure := &tree.MultisigClosure{}
 		valid, err := closure.Decode([]byte{})
-		require.Error(t, err)
```
louisinger: Any reason to change this test ?

louisinger (changes requested):

common/tree/script.go
aruokhai: seams about right :+1:

aruokhai: unfortunately, that would not work, because the next OPCODE after the data OPCODE in a `multichecksigadd` is OP_CHECKSIG, and not OP_CHECKSIG_ADD

aruokhai:
``` golang
case MultisigTypeChecksigAdd:
		for i, pubkey := range f.PubKeys {
			scriptBuilder.AddData(schnorr.SerializePubKey(pubkey))
			if i == 0 {
				scriptBuilder.AddOp(txscript.OP_CHECKSIG)
				continue
			}
			scriptBuilder.AddOp(txscript.OP_CHECKSIGADD)
		}
		scriptBuilder.AddInt64(int64(len(f.PubKeys)))
		scriptBuilder.AddOp(txscript.OP_NUMEQUAL)
	}
```

louisinger: We can still check if the second one is CHECKSIGADD (should be the 67th byte in the script). Or even better, we can use the last byte: 
* if last byte is a NUMEQUAL -> try decode checksigadd
* else -> try decode checksig

aruokhai: I replaced it with the initial logic, try decode with `decodecheckmultisig`, if fail `decodewithchecksigadd`

louisinger (approved):
tACK, thanks @aruokhai 
waiting for @altafan to double-check

louisinger:
@aruokhai can you merge with master to run the unit test in CI ?

aruokhai:
@louisinger done

common/tree/script.go
```diff
@@ -176,76 +176,75 @@ func (f *MultisigClosure) Script() ([]byte, error) {
...
 	pubkeys := make([]*secp256k1.PublicKey, 0)
 
-	// Keep track of position in script
-	pos := 0
+	for tokenizer.Next() {
 
-	for pos < len(script) {
-		// Check for 33-byte data push (32 bytes for pubkey + 1 byte for OP_DATA)
-		if pos+33 > len(script) {
+		if txscript.IsSmallInt(tokenizer.Opcode()) {
```
altafan, line 214: Let's put a comment over this line to explain why we break

common/tree/script.go
```diff
@@ -418,26 +407,37 @@ func (d *CSVMultisigClosure) Decode(script []byte) (bool, error) {
...
-	sequence = sequence[1:]
 
 	locktime, err := common.BIP68DecodeSequence(sequence)
 	if err != nil {
 		return false, err
 	}
 
 	multisigClosure := &MultisigClosure{}
-	valid, err := multisigClosure.Decode(script[csvIndex+2:])
+	valid, err := multisigClosure.Decode(tokenizer.Script()[tokenizer.ByteIndex():])
```
altafan: would be better to put `tokenizer.Script()[tokenizer.ByteIndex():]` into a variable like:
```suggestion
	buf := tokenizer.Script()[tokenizer.ByteIndex():]
	valid, err := multisigClosure.Decode(buf)
```

common/tree/script.go
```diff
@@ -806,7 +816,7 @@ func (d *CLTVMultisigClosure) Decode(script []byte) (bool, error) {
 	}
 
 	multisigClosure := &MultisigClosure{}
-	valid, err := multisigClosure.Decode(script[cltvIndex+2:])
+	valid, err := multisigClosure.Decode(tokenizer.Script()[tokenizer.ByteIndex():])
```
altafan: would be better to put `tokenizer.Script()[tokenizer.ByteIndex():]` into a variable like:
```suggestion
	buf := tokenizer.Script()[tokenizer.ByteIndex():]
	valid, err := multisigClosure.Decode(buf)
```

aruokhai:
Hey @altafan

PR #402, opened 2024-12-15
test(common): add more script-decode test cases

sljuka:
Add test cases that go through supported script closures: multisigClosure, csvMultisigClosure, cltvMultisigClosure and conditionMultisig.

If testing/code convention or anything regarding my code can be improved, please let me know, it would be helpful since I am starting out with Golang.

This should solve https://github.com/ark-network/ark/issues/395 (plan-b-network assignment)

PR #410, opened 2024-12-23
[TEST] Changes in the TestDecodeClosure Func

juarez009:


altafan:
Thanks for the contributon, but this is a duplicate of the already merged #402

PR #421, opened 2025-01-15
Split protos

### LINKED ISSUE #409: move  explorer-like RPCs to dedicate gRPC service
this would allow to cleanly register at startup those APIs, some operators may decide to not want to serve data to clients, making this a config option.

altafan:
This moves the explorer-like RPCs to a dedicated service and all message types to a dedicated proto file.

It also updates the SDK and the server's interface layer accordingly.

NOTE: This also makes use of simple int64 to represent any timestamp/duration in protos.

Closes #409.

Please @sekulicd @louisinger review


server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -21,14 +18,19 @@ import (
...
+
 type handler struct {
 	svc application.Service
 
 	eventsListenerHandler       *listenerHanlder[*arkv1.GetEventStreamResponse]
 	transactionsListenerHandler *listenerHanlder[*arkv1.GetTransactionsStreamResponse]
 }
 
-func NewHandler(service application.Service) arkv1.ArkServiceServer {
+func NewHandler(service application.Service) service {
```
sekulicd, line 33: Shall we also split handlers?

sekulicd:
> NOTE: This also makes use of simple int64 to represent any timestamp/duration in protos.

Considering market hours config(time/duration) could come via env vars and RPC, IMO best way is to use google.protobuf.Timestamp/google.protobuf.Duration so we dont have issues with time/duration conversion via code base.

altafan:
> Considering market hours config(time/duration) could come via env vars and RPC, IMO best way is to use google.protobuf.Timestamp/google.protobuf.Duration so we dont have issues with time/duration conversion via code base.

The standard way of representing a timestamp is with `int64`.
`time.Time` exposes a `Unix()` method that safely translates a timestamp to `int64`, while `time.Duration` is simply a `int64` type with some custom methods.
Lastly, we already represent other timestamps and durations with `int64` in the protos, so this was made mainly for consistency

PR #422, opened 2025-01-15
Allow the user to use custom signing scheme and multiple keys in vtxo tree signatures

louisinger:
_This PR adds a way for the user to sign only its branch during the vtxo tree signing process._

now `RegisterOutputsForNextRound` RPC expects the following request:

```proto
message Musig2 {
  repeated string cosigners_public_keys = 1;
  uint32 signing_type = 2;
}

message RegisterOutputsForNextRoundRequest {
  string request_id = 1;
  // List of receivers for to convert to leaves in the next VTXO tree. 
  repeated Output outputs = 2;
  optional Musig2 musig2 = 3;
}
```

* `ephemeral_pubkey` has been replaced by `musig2.cosigners_public_keys`. It allows setting multiple signing keys for the same request. Useful for collaborative VTXO contracts.
* `signing_type` is a flag signalling to the server the type of signing scheme the user wants to use, 2 are supported:
    * `SignAll = 0` is the existing scheme where one signs all branches of the tree.
    * `SignBranch = 1` is the new scheme where one signs only the branch(es) related to his VTXO(s) in the tree.

**BREAKING CHANGES:**

With this, the tree's PSBTs don't embed the sweep tapscript leaf and internal taproot key as unknown fields anymore. They are replaced instead with the batch lifetime, as clients know all other info to reconstruct the sweep leaf on their own. 

@tiero @altafan @sekulicd please review


common/bitcointree/musig2.go
```diff
@@ -109,110 +125,163 @@ func AggregateKeys(
...
-			nonce, err := musig2.GenNonces(
-				musig2.WithPublicKey(t.secretKey.PubKey()),
-			)
-			if err != nil {
-				return err
+	// Pre-allocate the result matrix
+	myNonces := make([][]*musig2.Nonces, len(t.txs))
+	for i := range myNonces {
+		myNonces[i] = make([]*musig2.Nonces, len(t.txs[i]))
+	}
```
altafan:
```suggestion
	myNonces := make([][]*musig2.Nonces, 0, len(t.txs))
	for range t.txs {
		myNonces = append(myNonces, make([]*musig2.Nonces, len(t.txs[i])))
	}
```

common/bitcointree/musig2.go
```diff
@@ -235,75 +309,72 @@ func (t *treeSignerSession) GetNonces() (TreeNonces, error) {
...
-
 	if t.aggregateNonces == nil {
 		return nil, errors.New("nonces not set")
 	}
 
-	sigs := make(TreePartialSigs, 0)
+	sigs := make(TreePartialSigs, len(t.txs))
+	for i := range sigs {
+		sigs[i] = make([]*musig2.PartialSignature, len(t.txs[i]))
+	}
```
altafan: same change here

common/bitcointree/musig2.go
```diff
@@ -235,75 +309,72 @@ func (t *treeSignerSession) GetNonces() (TreeNonces, error) {
...
 
 	return sigs, nil
 }
 
-func (t *treeSignerSession) signPartial(partialTx *psbt.Packet, posx int, posy int, seckey *btcec.PrivateKey) (*musig2.PartialSignature, error) {
+func (t *treeSignerSession) signPartial(
+	partialTx *psbt.Packet,
+	posx int,
+	posy int,
+	seckey *btcec.PrivateKey,
```
altafan: can be removed?

common/bitcointree/musig2.go
```diff
@@ -403,122 +449,154 @@ func (t *treeCoordinatorSession) AggregateNonces() (TreeNonces, error) {
 		}
 	}
 
-	aggregatedNonces := make(TreeNonces, 0)
+	aggregatedNonces := make(TreeNonces, len(t.txs))
+	for i := range aggregatedNonces {
+		aggregatedNonces[i] = make([]*Musig2Nonce, len(t.txs[i]))
+	}
```
altafan: same change here

common/bitcointree/builder.go
```diff
@@ -7,54 +7,48 @@ import (
...
-	"github.com/btcsuite/btcd/btcec/v2/schnorr/musig2"
 	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
 
-// CraftSharedOutput returns the taproot script and the amount of the initial root output
+// CraftSharedOutput returns the taproot script and the amount of the initial root shared output of a vtxo tree
```
altafan:
```suggestion
// CraftSharedOutput returns the taproot script and the amount of the root shared output of a vtxo tree
```

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -140,8 +140,9 @@ message GetBoardingAddressResponse {
 
 message RegisterInputsForNextRoundRequest {
   repeated Input inputs = 1;
-  optional string ephemeral_pubkey = 2;
+  repeated string signer_pubkeys = 2;
```
altafan: aggreagate `signer_pubkeys` and `signing_type` within a `musig2` field

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -246,9 +247,8 @@ message RoundFailed {
 
 message RoundSigningEvent {
   string id = 1;
-  repeated string cosigners_pubkeys = 2;
-  Tree unsigned_vtxo_tree = 3;
-  string unsigned_round_tx = 4;
+  Tree unsigned_vtxo_tree = 2;
```
altafan: let's keep the list of cosigners, this way it's easier for clients to understand if they are participating this round or have to wait for the next one.

common/bitcointree/musig2_test.go
```diff
@@ -26,110 +27,107 @@ var lifetime = common.RelativeLocktime{Type: common.LocktimeTypeBlock, Value: 14
...
+	t.Parallel()
 	fixtures := parseFixtures(t)
-	for _, f := range fixtures.Valid {
-		// Generate 20 cosigners
-		cosigners := make([]*secp256k1.PrivateKey, 20)
-		cosignerPubkeys := make([]*btcec.PublicKey, 20)
-		for i := 0; i < 20; i++ {
-			prvkey, err := secp256k1.GeneratePrivateKey()
+	for i, f := range fixtures.Valid {
+		t.Run(fmt.Sprintf("test %d", i), func(t *testing.T) {
```
altafan: let's give each test vector a name instead of calling them `test_0`, `test_1`, ...

common/bitcointree/musig2_test.go
```diff
@@ -26,110 +27,107 @@ var lifetime = common.RelativeLocktime{Type: common.LocktimeTypeBlock, Value: 14
 var testTxid, _ = chainhash.NewHashFromStr("49f8664acc899be91902f8ade781b7eeb9cbe22bdd9efbc36e56195de21bcd12")
 
 func TestRoundTripSignTree(t *testing.T) {
```
altafan: The name of the test is misleading, wdyt of `TestBuildAndSignVtxoTree`? Also the file should be renamed, but i'd do that once we refactor the whole common pkg

common/bitcointree/validation.go
```diff
@@ -206,38 +197,21 @@ func validateNodeTransaction(node tree.Node, tree tree.VtxoTree, tapTreeRoot []b
...
 
-		aggregatedKey, err := AggregateKeys(cosigners, tapTreeRoot)
-		if err != nil {
-			return fmt.Errorf("unable to aggregate keys: %w", err)
-		}
-
-		if !bytes.Equal(inputData.TaprootInternalKey, schnorr.SerializePubKey(aggregatedKey.PreTweakedKey)) {
-			return ErrInternalKey
+		if len(cosigners) == 0 {
+			return ErrInvalidTaprootScript
```
altafan: why this error message?

common/locktime.go
```diff
@@ -69,6 +78,13 @@ func (l RelativeLocktime) LessThan(other RelativeLocktime) bool {
 	return l.Compare(other) < 0
 }
 
+func (l RelativeLocktime) Bytes() []byte {
+	b := make([]byte, 5)
+	b[0] = byte(l.Type)
+	binary.BigEndian.PutUint32(b[1:], l.Value)
```
altafan: Isn't it encoded in little endian within a bitcoin script?

common/tree/type.go
```diff
@@ -4,9 +4,20 @@ import (
...
+
 type TreeFactory func(outpoint psetv2.InputArgs) (VtxoTree, error)
 
 type VtxoLeaf struct {
-	PubKey string
-	Amount uint64
+	PubKey            string
+	Amount            uint64
+	Type              SigningType
+	SignersPublicKeys []string
```
altafan: like for the protos, let's put these fields in a `Musig2` struct. Also rename them `Type` > `SigningType` and `SignersPublicKeys` > `CosignerPubkeys`

pkg/client-sdk/client/client.go
```diff
@@ -25,10 +25,10 @@ type RoundEvent interface {
 type TransportClient interface {
 	GetInfo(ctx context.Context) (*Info, error)
 	RegisterInputsForNextRound(
-		ctx context.Context, inputs []Input, ephemeralKey string,
+		ctx context.Context, inputs []Input, signerPubKeys []string, signingType tree.SigningType,
```
altafan:
```suggestion
		ctx context.Context, inputs []Input, cosignerPubkeys []string, signingType tree.SigningType,
```

pkg/client-sdk/client/client.go
```diff
@@ -25,10 +25,10 @@ type RoundEvent interface {
...
 	) (string, error)
 	RegisterNotesForNextRound(
-		ctx context.Context, notes []string, ephemeralKey string,
+		ctx context.Context, notes []string, signerPubKeys []string, signingType tree.SigningType,
```
altafan:
```suggestion
		ctx context.Context, notes []string, cosignerPubkeys []string, signingType tree.SigningType,
```

pkg/client-sdk/client/client.go
```diff
@@ -191,10 +191,9 @@ type RoundFailedEvent struct {
 func (e RoundFailedEvent) isRoundEvent() {}
 
 type RoundSigningStartedEvent struct {
-	ID               string
-	UnsignedTree     tree.VtxoTree
-	CosignersPubKeys []*secp256k1.PublicKey
-	UnsignedRoundTx  string
+	ID              string
```
altafan: revert changes

pkg/client-sdk/client/client.go
altafan: change anywhere else in the code pls

pkg/client-sdk/covenantless_client.go
```diff
@@ -34,6 +34,42 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// SettleOptions are only available for covenantless clients
+// it allows to customize the vtxo signing process
+type SettleOptions struct {
```
altafan:
```suggestion
type SignOptions struct {
```

maybe better?

pkg/client-sdk/covenantless_client.go
```diff
@@ -34,6 +34,42 @@ import (
...
+		return fmt.Errorf("invalid options type")
+	}
+
+	t := tree.SignAll
+	opts.SigningType = &t
+	return nil
+}
+
+// WithCustomTreeSigner allows to use a set of custom signer for the vtxo tree signing process
+func WithCustomTreeSigner(privKeys []*secp256k1.PrivateKey) Option {
```
altafan: I took a look at handleOptions and i think this should be come WithExtraSigner to be useful for the user.. this way he just need to add extra keys, without taking care of creating one for himself, that should always be done by the SDK.

```suggestion
func WithExtraSigner(privKeys []*secp256k1.PrivateKey) Option {
```

pkg/client-sdk/covenantless_client.go
```diff
@@ -2702,3 +2771,33 @@ func buildRedeemTx(
...
+
+	var signingType tree.SigningType
+	if options.SigningType != nil {
+		signingType = *options.SigningType
+	} else {
+		signingType = tree.SignBranch
+	}
+
+	// if no custom signer priv keys are provided, we generate a new ephemeral key
+	if len(signerPubKeys) == 0 {
```
altafan: in light of [this comment](https://github.com/ark-network/ark/pull/422/files#r1918586259), we should always generate the keypair for the sdk wallet

pkg/client-sdk/covenantless_client.go
```diff
@@ -1723,52 +1784,60 @@ func (a *covenantlessArkClient) handleRoundSigningStarted(
...
 
-	signerSession = bitcointree.NewTreeSignerSession(
-		ephemeralKey, sharedOutputValue, event.UnsignedTree, root.CloneBytes(),
-	)
+	signerSessions := make([]bitcointree.SignerSession, 0, len(signerPrivKeys))
 
-	if err = signerSession.SetKeys(event.CosignersPubKeys); err != nil {
-		return
-	}
+	for _, privKey := range signerPrivKeys {
```
altafan: should be done in parallel?

pkg/client-sdk/covenantless_client.go
```diff
@@ -1723,52 +1784,60 @@ func (a *covenantlessArkClient) handleRoundSigningStarted(
...
 ) error {
-	if signerSession == nil {
+	if len(signerSessions) <= 0 {
 		return fmt.Errorf("tree signer session not set")
 	}
 
-	if err := signerSession.SetAggregatedNonces(event.Nonces); err != nil {
-		return err
-	}
+	for _, session := range signerSessions {
```
altafan: same here?

pkg/client-sdk/covenantless_client.go
```diff
@@ -2702,3 +2771,33 @@ func buildRedeemTx(
...
+	} else {
+		signingType = tree.SignBranch
+	}
+
+	// if no custom signer priv keys are provided, we generate a new ephemeral key
+	if len(signerPubKeys) == 0 {
+		ephemeralKey, err := secp256k1.GeneratePrivateKey()
+		if err != nil {
+			return nil, nil, tree.SignBranch, err
+		}
```
altafan: maybe we should use an ephemeral key only when signing ALL, otherwise we should use a wallet key to be sure we can re-sign the branch if needed?

server/internal/core/application/covenantless.go
```diff
@@ -528,14 +538,20 @@ func (s *covenantlessService) SpendNotes(ctx context.Context, notes []note.Note)
 		return "", fmt.Errorf("failed to create tx request: %s", err)
 	}
 
+	// add the server pubkey as a signer
+	signerPubkeys = append(signerPubkeys, hex.EncodeToString(s.serverSigningPubKey.SerializeCompressed()))
+
+	request.AddSignerPubKeys(signerPubkeys)
+	request.AddSigningType(signingType)
+
```
altafan: why do we do this here? the server key could be added only once, when the requests are popped from the queue and the batching is processed.

This way instead we would end up duplicating the server key in memory for every note redeemed in the same round, wouldn't we?

server/internal/core/application/covenantless.go
```diff
@@ -635,6 +651,16 @@ func (s *covenantlessService) SpendVtxos(ctx context.Context, inputs []ports.Inp
...
 	}
+
+	if len(signerPubkeys) <= 0 {
+		return "", fmt.Errorf("signer pubkeys are required")
+	}
+
+	signerPubkeys = append(signerPubkeys, hex.EncodeToString(s.serverSigningPubKey.SerializeCompressed()))
+
+	request.AddSignerPubKeys(signerPubkeys)
+	request.AddSigningType(signingType)
```
altafan: same here?

server/internal/core/application/covenantless.go
```diff
@@ -1058,17 +1064,9 @@ func (s *covenantlessService) startFinalization() {
...
-	ephemeralKey, err := secp256k1.GeneratePrivateKey()
-	if err != nil {
-		round.Fail(fmt.Errorf("failed to generate ephemeral key: %s", err))
-		log.WithError(err).Warn("failed to generate ephemeral key")
-		return
-	}
-
-	cosigners = append(cosigners, ephemeralKey.PubKey())
-
+	log.Debugf("building round tx for round %s", round.Id)
```
altafan:
```suggestion
	log.Debugf("building tx for round %s", round.Id)
```

server/internal/core/application/covenantless.go
```diff
@@ -1080,16 +1078,24 @@ func (s *covenantlessService) startFinalization() {
...
 
-		signingSession := newMusigSigningSession(len(cosigners))
+		uniqueSignerPubkeys := make(map[string]struct{})
+		for _, request := range requests {
+			for _, pubkey := range request.SignerPubKeys {
+				uniqueSignerPubkeys[pubkey] = struct{}{}
+			}
+		}
+
+		nbOfCosigners := len(uniqueSignerPubkeys)
```
altafan, line 1089: this number should be wrong as all requests contain the server's key, shouldn't it?

server/internal/core/application/covenantless.go
```diff
@@ -1080,16 +1078,24 @@ func (s *covenantlessService) startFinalization() {
 	s.forfeitTxs.init(connectors, requests)
 
 	if len(vtxoTree) > 0 {
-		log.Debugf("signing vtxo tree for round %s", round.Id)
 
-		signingSession := newMusigSigningSession(len(cosigners))
+		uniqueSignerPubkeys := make(map[string]struct{})
```
altafan: if we added the server key only after popping the requests, we could just count the number of cosigners instead of using this map

louisinger: No, cause you don't know if 2 different TxRequest are using the same cosigner

api-spec/protobuf/ark/v1/service.proto
altafan: anyway, let's make the user send the cosigner pubkeys when registering the outputs, not the inputs

pkg/client-sdk/covenantless_client.go
louisinger: Personal preference for Settle cause "Sign" term may also be associated with redeem tx (signing forfeit/redeem) 🤔 . `VtxoTreeOptions` ?

pkg/client-sdk/covenantless_client.go
louisinger: in theory, sign ALL doesn't prevent the server to use the sweep-early feature so I'd keep wallet's key whatever the type

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -120,17 +120,22 @@ message GetBoardingAddressResponse {
...
-  repeated string notes = 3;
+  repeated string notes = 2;
 }
 message RegisterInputsForNextRoundResponse {
   string request_id = 1;
 }
 
+message Musig2 {
+  repeated string cosigners_public_keys = 1;
+  uint32 signing_type = 2;
```
altafan: Since we don't really plan to add other signing type, whay if we make it a bool field that self describes the sign ALL option?

```suggestion
  bool signing_all = 2;
```

altafan:
Tried to make 2 rest client settle at the same time and they crashed because of the encoding of the nonces. Seems like some elems of the matrix are nil when signing a branch.

common/bitcointree/builder.go
```diff
@@ -228,104 +219,121 @@ func getTx(
...
 	}
 
-	nodes := make([]node, 0, len(receivers))
+	cosignersALL := make([]*secp256k1.PublicKey, 0)
 	for _, r := range receivers {
-		pubkeyBytes, err := hex.DecodeString(r.PubKey)
-		if err != nil {
-			return nil, err
+		if r.Musig2Data == nil {
+			return nil, fmt.Errorf("musig2 data is nil for %s", r.PubKey)
```
altafan:
```suggestion
			return nil, fmt.Errorf("missing musig2 data for receiver %s", r.PubKey)
```

common/bitcointree/builder.go
```diff
@@ -228,104 +219,121 @@ func getTx(
...
-			return nil, err
+		if r.Musig2Data.SigningType != tree.SignAll {
+			continue
 		}
 
-		leafNode := &leaf{
-			amount: int64(r.Amount),
-			pubkey: pubkey,
+		if len(r.Musig2Data.CosignersPublicKeys) == 0 {
+			return nil, fmt.Errorf("cosigners public keys for %s are empty", r.PubKey)
```
altafan:
```suggestion
			return nil, fmt.Errorf("missing cosigners public keys for receiver %s", r.PubKey)
```

common/bitcointree/musig2_test.go
```diff
@@ -21,152 +22,205 @@ const (
...
-	fixtures := parseFixtures(t)
-	for _, f := range fixtures.Valid {
-		// Generate 20 cosigners
-		cosigners := make([]*secp256k1.PrivateKey, 20)
-		cosignerPubkeys := make([]*btcec.PublicKey, 20)
-		for i := 0; i < 20; i++ {
-			prvkey, err := secp256k1.GeneratePrivateKey()
+	for _, tc := range generateTestCases(t) {
+		t.Run(tc.name, func(t *testing.T) {
+			_, sharedOutputAmount, err := bitcointree.CraftSharedOutput(
```
altafan: let's not ignore any return value and also ensure they are valid (non nil/zero/empty)

altafan: please apply this change to the rest of the test too.

common/bitcointree/musig2_test.go
```diff
@@ -21,152 +22,205 @@ const (
...
-		// Get signatures from all signers
-		for i, session := range signerSessions {
-			sig, err := session.Sign()
-			require.NoError(t, err)
-			err = serverCoordinator.AddSig(cosignerPubkeys[i], sig)
+			// Cceate signer sessions for each receivers
+			signerSessions := make(map[*btcec.PublicKey]bitcointree.SignerSession)
+			for _, prvkey := range tc.privKeys {
+				session := bitcointree.NewTreeSignerSession(prvkey)
+				require.NoError(t, session.Init(sweepRoot[:], sharedOutputAmount, vtxoTree))
```
altafan: please, do not "hide" the api call into the require, move it out

altafan: apply the change to the rest of the test

altafan (approved):

PR #434, opened 2025-01-27
Add txid to SubmitRedeemTx response

altafan:
This adds a txid field to the response message of SubmitRedeemTx to facilitate clients retrieving the hash of the offchain tx.


tiero:
Is there a reason to keep the signed PSBT in this flow? Just curious, it won't break clients, so is good.

altafan:
It's "internal": the transport client returns both signed tx and txid, while the SDK API returns only the txid.

louisinger (approved):

PR #449, opened 2025-02-05
[SDK] Fix closing event stream in rest client

altafan:
This makes the rest client of the client SDK properly handle the close of the event stream so that clients don't left connections opened unintentionally.

BONUS: This also makes the server able to force the close of such connections in case of shutdown.

Please @louisinger @bordalix @sekulicd review.


bordalix (approved):

louisinger (changes requested):
cherry-pick https://github.com/ark-network/ark/pull/450/commits/b0981991e0f5c15dc9902316038bb981394041b7 to fix the CI

PR #450, opened 2025-02-06
[SDK] add `WithEventsCh` settle option

louisinger:
_This PR adds a way to inspect the settlement process by replaying the`RoundEvent` received from the server to a custom channel created by the client. It also allow to do the same with JS using a callback function._

**Golang**

```go
eventsCh := make(chan client.RoundEvent)
defer close(eventsCh)
go func() {
	for event := range eventsCh {
		fmt.Println(event)
	}
}()

resp, err := arkSdkClient.Settle(context.Background(), arksdk.WithEventsCh(eventsCh))
if err != nil {
	return nil, err
}
```

**WASM JS wrapper**

```js
await settle((e) => console.log(JSON.parse(e)));
```

@tiero @bordalix @altafan @sekulicd  please review

### DIFF
--- a/nak.Dockerfile
+++ b/nak.Dockerfile
@@ -1,5 +1,5 @@
 # Use official Go image as base
-FROM golang:1.23.1-alpine
+FROM golang:1.23.3-alpine
 
 # Install git (needed for go install)
 RUN apk add --no-cache git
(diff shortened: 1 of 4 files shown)


sekulicd:
@louisinger which use cases this helps with?

bordalix (approved):

tiero:
> @louisinger which use cases this helps with?

clients can monitor settlement events for better UX instead now `await settle()` dont' really communicate the progression of the settlement process.

PR #452, opened 2025-02-06
rename round to batch in log messages

bordalix:
@altafan please review


tiero:
Would this be redundant with [other work](https://github.com/ark-network/ark/pull/450) if we actually expose proper callbacks so then you can handle the "settlement events" as you please? (ideally a visual progress bar instead of textual messages).

PR #453, opened 2025-02-07
[Server] Optmization of embedded wallet and forfeit tx validation

altafan:
The purpose of this is to optimize the flow of finalizing a round. In particular:
- Embedded bitcoin wallet: cache forfeit address at creation/unlocking instead of doing so the very first time the relative API is called. Contains also some renaming for consistency.
- SubmitForfeitTxs API: move validation of txs signatures at round finalization to prevent clients to have wait for their calls to succeed (if a lot of clients join the same round)

### DIFF
--- a/server/internal/core/application/covenant.go
+++ b/server/internal/core/application/covenant.go
@@ -707,6 +707,21 @@ func (s *covenantService) finalizeRound() {
 		return
 	}
 
+	// TODO: make this concurrent and ensure all forfeits are verified before returning error
+	for _, tx := range forfeitTxs {
+		ok, txid, err := s.builder.VerifyTapscriptPartialSigs(tx)
+		if err != nil {
+			changes = round.Fail(fmt.Errorf("failed to validate forfeit tx %s: %s", txid, err))
+			log.WithError(err).Warnf("failed to validate forfeit tx %s: %s", txid, err)
+			return
+		}
+		if !ok {
+			changes = round.Fail(fmt.Errorf("invalid signature for forfeit tx %s", txid))
+			log.Warnf("invalid signature for forfeit tx %s", txid)
+			return
+		}
+	}
+
 	log.Debugf("signing round transaction %s\n", round.Id)
 
 	boardingInputs := make([]int, 0)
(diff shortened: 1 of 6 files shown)


server/internal/core/application/covenantless.go
```diff
@@ -1401,6 +1401,21 @@ func (s *covenantlessService) finalizeRound(notes []note.Note) {
 		return
 	}
 
+	// TODO: make this concurrent and ensure all forfeits are verified before returning error
```
louisinger: `verifyTapscriptPartialSigsMap(txs map[string]*psbt.Packet)` let's reuse that function ?

sekulicd (approved):

server/internal/core/application/covenantless.go
altafan: done

louisinger (approved):

server/internal/core/application/covenant.go
```diff
@@ -707,6 +708,12 @@ func (s *covenantService) finalizeRound() {
 		return
 	}
 
+	if err := s.verifyForfeitTxsSigs(forfeitTxs); err != nil {
+		changes = round.Fail(err)
+		log.WithError(err).Warn("failed to validate forfeit txs")
```
tiero, line 713: here we simply log and dont halt execution right?

server/internal/core/application/covenantless.go
```diff
@@ -750,6 +750,11 @@ func (s *covenantlessService) UpdateTxRequestStatus(_ context.Context, id string
 }
 
 func (s *covenantlessService) SignVtxos(ctx context.Context, forfeitTxs []string) error {
+	start := time.Now()
+	defer func() {
+		end := time.Since(start)
+		fmt.Printf("SUBMIt FORFEIT TXS TOOK: %dµs", end.Microseconds())
```
tiero: is fmt.Printf what we want? ie. is then controlled by the logrus log level? Also `SUBMIt` > `SUBMIT` if you want to keep upper case.

tiero: it shoudl be a log level Trace, not even debug imho

altafan: it's been removed

PR #467, opened 2025-02-18
Wait for boarding input signatures

### LINKED ISSUE #466: `finalizeRound` should stop sleeping in case boarding utxos are all signed
Like we did for [forfeits transactions](https://github.com/ark-network/ark/pull/462), we should also have a "doneCh" for boarding utxos signatures.

sekulicd:
This adds waiting for boarding inputs signatures.

This closes #466 

@altafan @louisinger please review.

### DIFF
--- a/server/internal/core/application/covenantless.go
+++ b/server/internal/core/application/covenantless.go
@@ -1534,10 +1534,11 @@ func (s *covenantlessService) waitForForfeitsAndBoardingSigs(
 				}
 
 				numOfInputsSigned := 0
-				for i := range roundTx.UnsignedTx.TxIn {
-					in := &roundTx.Inputs[i]
-					if len(in.FinalScriptWitness) > 0 {
-						numOfInputsSigned++
+				for _, v := range roundTx.Inputs {
+					if len(v.TaprootScriptSpendSig) > 0 {
+						if len(v.TaprootScriptSpendSig[0].Signature) > 0 {
+							numOfInputsSigned++
+						}
 					}
 				}
 
--- a/server/internal/core/application/covenantless.go
+++ b/server/internal/core/application/covenantless.go
@@ -1137,6 +1137,7 @@ func (s *covenantlessService) startFinalization(roundEndTime time.Time) {
 
 	var notes []note.Note
 	var roundAborted bool
+	numOfBoardingInputs := 0
 	defer func() {
 		delete(s.treeSigningSessions, round.Id)
 		if roundAborted {
@@ -1153,7 +1154,7 @@ func (s *covenantlessService) startFinalization(roundEndTime time.Time) {
 			return
 		}
 
-		s.finalizeRound(notes, roundEndTime)
+		s.finalizeRound(notes, roundEndTime, numOfBoardingInputs)
 	}()
 
 	if round.IsFailed() {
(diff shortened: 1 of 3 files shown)


server/internal/core/application/covenantless.go
```diff
@@ -1501,6 +1502,74 @@ func (s *covenantlessService) finalizeRound(notes []note.Note, roundEndTime time
...
+	doneCh := make(chan struct{}, 1)
+	errCh := make(chan error, 1)
+
+	go func() {
+		ticker := time.NewTicker(time.Second / 3)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-waitCtx.Done():
```
altafan, line 1522: we have 2 functions listining to the same channel.. this means only one of them will receive the done message from waitCtx. We can't control whom, so this can lead to unexpected behaviors.

In general i'm not sure about this approach, considering that clients submit sigs for their boarding inputs by calling SubmitSignedForfeitTxs.

I'd like to stick with the existing one and fixi where we signal that forfeit txs collection ended. At the moment we likely wait only for forfeits, we should wait also for boarding inputs.

@louisinger can you give more insights please?

sekulicd:
> we have 2 functions listining to the same channel.. this means only one of them will receive the done message from waitCtx. We can't control whom, so this can lead to unexpected behaviors.

ctx done can be used to shutdown multiple go routines, tested it with bellow code, also found resources [online](https://stackoverflow.com/questions/61007385/golang-pattern-to-kill-multiple-goroutines-at-once)

```
package main

import (
	"context"
	"fmt"
	"time"
)

func main() {
	waitCtx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancel()

	go func() {
		ticker := time.NewTicker(time.Second / 3)
		defer ticker.Stop()

		for {
			select {
			case <-waitCtx.Done():
				fmt.Print("done go routine 1")
				return

			case <-ticker.C:

			}
		}
	}()

	go func() {
		ticker := time.NewTicker(time.Second / 3)
		defer ticker.Stop()

		for {
			select {
			case <-waitCtx.Done():
				fmt.Println("done go routine 2")
				return

			case <-ticker.C:

			}
		}
	}()

	go func() {
		ticker := time.NewTicker(time.Second / 3)
		defer ticker.Stop()

		for {
			select {
			case <-waitCtx.Done():
				fmt.Println("done go routine 3")
				return

			case <-ticker.C:

			}
		}
	}()

	select {
	case <-waitCtx.Done():
		fmt.Println("done main routine")
	}

	time.Sleep(5 * time.Second)
}
```

Regarding approach, dont see better option, as we have two global vars, forfeitMap and currentRound, that can be accessed asynchronous by multiple users, we need to poll and check if all necessary data are sent from all clients.

louisinger:
@sekulicd an alternative could be to check if the job is done (all forfeits & all boarding inputs signed) at the end of `SignVtxos` and `SignRoundTx` methods. This way we don't need polling, what do u think ?

I don't like this polling cause it locks the currentRound var every Second / 3, it may slow down the access to that var for people signing the round tx no ?

sekulicd:
> @sekulicd an alternative could be to check if the job is done (all forfeits & all boarding inputs signed) at the end of `SignVtxos` and `SignRoundTx` methods. This way we don't need polling, what do u think ?
> 
> I don't like this polling cause it locks the currentRound var every Second / 3, it may slow down the access to that var for people signing the round tx no ?

we need to lock it either way no? @louisinger

louisinger:
@sekulicd deleted the comment by mistake but I saw it:

yes `SignVtxos` & `SignRoundTx` locks the mutex, but maybe it's more efficient to do the "isDone" check at the same time instead of locking 3 times / second ?

PR #471, opened 2025-02-20
handle collaborative exit without vtxo change

louisinger:
_This PR fixes collaborative exit payment where there is no change (exit all)._

it includes 2 different fixes:
* server : handle `nil` musig2data.
* client : if onchain outputs only, skip musig2 signing steps.

@altafan please review


tiero:
Is WASM client also aware of this change?

altafan (approved):

altafan:
Yes the change affects the WASM, it's enough to update it client-side once this is merged

PR #472, opened 2025-02-20
[application] delete outdated tx requests

louisinger:
This PR cleans the requests map from outdated request. A request is outdated if the user didn't ping for at least 1 minute. It prevents `duplicated input` errors in case the client stops pinging and then retry with the same inputs.

@altafan please review

### DIFF
--- a/server/internal/core/application/utils.go
+++ b/server/internal/core/application/utils.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ark-network/ark/server/internal/core/ports"
 	"github.com/nbd-wtf/go-nostr"
 	"github.com/nbd-wtf/go-nostr/nip19"
+	log "github.com/sirupsen/logrus"
 )
 
 type timedTxRequest struct {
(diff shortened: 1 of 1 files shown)


server/internal/core/application/utils.go
```diff
@@ -118,6 +119,8 @@ func (m *txRequestsQueue) pop(num int64) ([]domain.TxRequest, []ports.BoardingIn
 		}
 		// Skip tx requests for which users didn't notify to be online in the last minute.
 		if p.pingTimestamp.IsZero() || time.Since(p.pingTimestamp).Minutes() > 1 {
+			log.Debugf("delete tx request %s because it didn't ping in the last minute", p.Id)
+			delete(m.requests, p.Id) // cleanup the request from the map
```
altafan: let's add an if and delete form the queue only if the last ping was longer than 5 minutes ago. Also, add a TODO to move this to a dedicated func. It's ok for now to have a side effect but we don't really want to keep it like this.

```suggestion
			// TODO: move to dedicated function.
			if p.pingTimestamp.IsZero() || time.Since(p.pingTimestamp).Minutes() > 1 {
			         log.Debugf("delete tx request %s because it didn't ping in the last minute", p.Id)
			         delete(m.requests, p.Id) // cleanup the request from the map
			}
```

altafan (approved):

PR #477, opened 2025-02-20
[SDK] Add GetVtxoEventChannel API

altafan:
This adds a new GetVtxoEventChannel API to the sdk client so that consumers can get notified whenever a list of vtxos is either added to the vtxo set or updated to spent.

This also adds changes so that GetTransactionEventChannel returns events for transactions that are added to the history, or updated to confirmed (onchain txs) or settled (offchain txs).

Please @bordalix @louisinger review this.

### DIFF
--- a/pkg/client-sdk/covenantless_client.go
+++ b/pkg/client-sdk/covenantless_client.go
@@ -2621,6 +2621,26 @@ func (a *covenantlessArkClient) handleRoundTx(
 				Settled:   true,
 				CreatedAt: time.Now(),
 			})
+		} else {
+			vtxosToAddAmount := uint64(0)
+			for _, v := range vtxosToAdd {
+				vtxosToAddAmount += v.Amount
+			}
+			settledBoardingAmount := uint64(0)
+			for _, tx := range pendingBoardingTxs {
+				settledBoardingAmount += tx.Amount
+			}
+			if vtxosToAddAmount > 0 && vtxosToAddAmount < settledBoardingAmount {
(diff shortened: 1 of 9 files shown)


pkg/client-sdk/covenantless_client.go
```diff
@@ -2653,6 +2667,11 @@ func (a *covenantlessArkClient) handleRoundTx(
 		}
 	}
 
+	fmt.Println("TXS TO ADD", len(txsToAdd))
+	fmt.Println("TXS TO SETTLE", len(txsToSettle))
+	fmt.Println("VTXOS TO ADD", len(vtxosToAdd))
+	fmt.Println("VTXOS TO SPEND", len(vtxosToSpend))
```
tiero:
```suggestion
```

pkg/client-sdk/types/types.go
```diff
@@ -96,6 +115,27 @@ func (t Transaction) String() string {
...
+type TxEventType int
+
+const (
+	TxsAdded TxEventType = iota
+	TxsSettled
+	TxsConfirmed
+)
+
+func (e TxEventType) String() string {
+	return map[TxEventType]string{
```
tiero, line 127: shall we have also an even for preconf? or thats the default state of TxsAdded

louisinger (approved):

pkg/client-sdk/types/types.go
altafan: we don't handle preconf state yet

PR #485, opened 2025-02-26
Add BIP0322 Proof of Funds & implement `RegisterIntent`

### LINKED ISSUE #483: BIP322 as ownership proof
Using [BIP322](https://bips.xyz/322) as an ownership proof allows us to use any spending path of a VTXO to prove ownership, which allows complex scripts to be forfeit without a full counterparty interaction if possible.

When registering as an intent, we should commit to a timestamp that fits consistently with the server time with a small variance (~5-10s) allowed. When registering for a round, we can use the round id.

### LINKED ISSUE #484: Notes simplification
The current system for swept/expired VTXOS to Notes is tightly coupled to Nostr: a user needs to register an npub upfront, and if the vtxos expire/get swept, we send a Nostr DM with a note.

However, this is very redundant. When a user loads their wallet, they can ping the server to see if any of their vtxos are no longer valid. The user may then use an ownership proof (#483) to reclaim their lost vtxo in a new batch. This completely removes the need for a note system for expired scenarios. 

We still want notes for other functions though, potentially for #479  and #482

louisinger:
update: https://github.com/ark-network/ark/pull/485#issuecomment-2690963762

_This PR deletes the nostr notifications system and replace it by a simple `GetNote` API. `OwnershipProof` is also replaced by [BIP0322](https://bips.xyz/322) "Proof of Funds"._

### GetNote

GetNote RPC expects `bip322_signature` and the associated `message`. If the signature is valid, the server returns a note where amount = sum of bip322_signature outpoints. The server also mark the VTXOs as spent in DB to prevent cheating.

### BIP0322 Proof of funds

`common.bip322` package is standalone and follows the BIP0322 specifications **⚠️  except that it doesn't verify the inputs sequence in order to allow the proof to be used for offchain outpoints.**

The `message` must be the current Unix timestamp. It is considered valid by the server if and only if it is close to current timestamp.

### CLI and SDK

`sdk.RecoverSweptVtxos` or `ark recover` CLI command automatically fetch swept and unspent VTXOs, create a BIP0322 signature, fetch the associated note and redeem it.

it closes #484
it closes #483 

@altafan @Kukks please review


common/bip322/proof.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, input := range inputs {
+		if err := input.validate(); err != nil {
+			return nil, err
+		}
+	}
+
+	firstInput := inputs[0]
+	toSpend := craftToSpendTx(message, firstInput.WitnessUtxo.PkScript)
+
+	outpoints := make([]*wire.OutPoint, 0, len(inputs)+1)
```
altafan: move this code to `craftToSignTx()` util function

common/bip322/signature_test.go
```diff
@@ -0,0 +1,71 @@
...
+		index, err := strconv.ParseUint(splitted[1], 10, 32)
+		require.NoError(t, err)
+
+		txHash, err := chainhash.NewHashFromStr(splitted[0])
+		require.NoError(t, err)
+
+		outpoint := wire.OutPoint{
+			Hash:  *txHash,
+			Index: uint32(index),
+		}
```
altafan: use wire.NewOutPointFromString

server/internal/core/application/covenantless.go
```diff
@@ -1107,41 +1109,129 @@ func (s *covenantlessService) RegisterCosignerSignatures(
...
+		return nil, fmt.Errorf("failed to parse message: %s", err)
 	}
 
-	if err := validateProofs(ctx, s.repoManager.Vtxos(), signedVtxoOutpoints); err != nil {
-		return err
+	now := time.Now()
+	messageTimestamp := time.Unix(parsedMessage, 0)
+
+	// parsedMessage should be between now - 15 seconds and now + 15 seconds
+	timeDiff := messageTimestamp.Sub(now)
```
altafan: swap, now - messageTimestamp

Kukks:
Couldn't one simply use the VTXO ownership proof in itself as a note? The api to exchange a proof for a vtxo  is redundant:

This flow:
* VTXO expires/gets swept
* Wallet goes online, detected expired vtxo
* Wallet creates proof, send to GetNote
* Server returns note
* Wallet redeems note for new VTXO

Proposed flow:
* VTXO expires/gets swept
* Wallet goes online, detected expired vtxo
* Wallet creates proof and registers it as input for next batch

louisinger:
> Proposed flow:
> 
> * VTXO expires/gets swept
> * Wallet goes online, detected expired vtxo
> * Wallet creates proof and registers it as input for next batch

I agree, but if we do that, we should be able to register also the not expired VTXOs in the same proof no ? If I have 1 VTXO not swept + 1 expired VTXO, I create a proof containing both outpoints and use it as input ? 

something like:
```proto
message RegisterInputsForNextRoundRequest {
  repeated Input inputs = 1; // not swept vtxos and their taproot scripts
  repeated string notes = 2;
  string bip322_signature = 3; // proof of funds for "inputs" array + swept vtxos you want to recover
}
```

bordalix:
> To register inputs, users must provide a valid BIP322 signature signing a timestamp message. The server accepts the signature if now - timestamp < 15 seconds.

Fragile:
1. It should be abs(now - timestamp) < 15 seconds
2. What are the limitations for this period?
3. Why only 15 seconds?

louisinger:
> 1. It should be abs(now - timestamp) < 15 seconds

not clear why ? timestamp should always be smaller than "now". 


> 2. What are the limitations for this period?
> 3. Why only 15 seconds?

It can be anything, the idea was to set a small enough delta so the proof can't be reused if intercepted. In fact maybe this delta should be the signer's choice / cc @Kukks ?

louisinger:
Converted to draft, I need to update the BIP322 proof with :
- [ ] allow outputs to be specified in the proof tx
- [ ] convert to PSBT
- [ ] specify taprootTree in the proof
- [ ] specify onchain output in the proof

Kukks:
> Converted to draft, I need to update the BIP322 proof with :
> 
> * [ ]  allow outputs to be specified in the proof tx
> * [ ]  convert to PSBT
> * [ ]  specify taprootTree in the proof
> * [ ]  specify onchain output in the proof

After chat, we will not use psbt after all and stay compatible with BIP322.  We will insert the data as a json inside the challenge message, something like
```
{
 "onchainOutputs": [0, 3...], (indexes)
 "inputTapTrees": [hex_encoded_tree, ...]
 (optional) "createdAt": timestamp
 (optional) "expiry": timestamp
```

louisinger:
Following the latest discussion, this PR has been updated to implement `RegisterIntent` allowing to register outputs within the BIP322 proof of ownership. 

the proof is signing the following message :

```go

type Message struct {
	// InputTapTrees is the list of taproot trees associated with the spent inputs
	// the index of the taproot tree in the list corresponds to the index of the input - 1
	// (we ignore the first bip322 input, as it is duplicate of the second one)
	InputTapTrees []string `json:"input_tap_trees"`
	// OnchainOutputIndexes specifies what are the outputs in the proof tx
	// that should be considered as onchain by the Ark operator
	OnchainOutputIndexes []int `json:"onchain_output_indexes"`
	// ValidAt is the timestamp (in seconds) at which the proof should be considered valid
	// if set to 0, the proof will be considered valid indefinitely or until ExpireAt is reached
	ValidAt int64 `json:"valid_at"`
	// ExpireAt is the timestamp (in seconds) at which the proof should be considered invalid
	// if set to 0, the proof will be considered valid indefinitely
	ExpireAt int64 `json:"expire_at"`
	// Musig2Data contains the related information about the vtxo tree signing
	// if the outputs are not registered in the proof or all the outputs are onchain, this field is not required
	Musig2Data *tree.Musig2 `json:"musig2_data"`
}
```

The idea is to sign all metadata required by the operator to include the intent in a batch.

cc @Kukks @altafan

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -119,11 +109,13 @@ message GetBoardingAddressResponse {
 
 /* In-Round Transaction API messages */
 
-message RegisterInputsForNextRoundRequest {
-  repeated Input inputs = 1;
+message RegisterIntentRequest {
+  // BIP322 signature embeds the outpoints and the proof of funds
+  Bip322Signature bip322_signature = 1;
   repeated string notes = 2;
```
Kukks, line 122:
```suggestion
  // TODO: Consolidate into bip322 as well
  repeated string notes = 2;
```

client/main.go
```diff
@@ -444,9 +440,7 @@ func redeem(ctx *cli.Context) error {
 	})
 }
 
-func registerNostrProfile(ctx *cli.Context) error {
-	profile := ctx.String(nostrProfileFlag.Name)
-
+func recoverVtxos(ctx *cli.Context) error {
```
Kukks, line 444: Is this needed? Would rather we just use the settle and have it include expired ones

common/bip322/message.go
```diff
@@ -0,0 +1,38 @@
+package bip322
+
+import (
+	"encoding/json"
+
+	"github.com/ark-network/ark/common/tree"
+)
+
+type Message struct {
```
Kukks: Maybe shouldnt be inside bip322, and be called IntentMessage?

client/main.go
```diff
@@ -270,18 +266,18 @@ func config(ctx *cli.Context) error {
...
-		"unilateral_exit_delay": cfgData.UnilateralExitDelay,
-		"dust":                  cfgData.Dust,
-		"boarding_exit_delay":   cfgData.BoardingExitDelay,
-		"explorer_url":          cfgData.ExplorerURL,
-		"forfeit_address":       cfgData.ForfeitAddress,
-		"with_transaction_feed": cfgData.WithTransactionFeed,
+		"server_url":                   cfgData.ServerUrl,
+		"server_pubkey":                hex.EncodeToString(cfgData.ServerPubKey.SerializeCompressed()),
+		"wallet_type":                  cfgData.WalletType,
+		"client_tyep":                  cfgData.ClientType,
```
Kukks: typo 👀 but existing code

client/main.go
Kukks: In fact, we should align with the js sdk and allow granular coin control in the sdk

pkg/client-sdk/wallet/singlekey/bitcoin_wallet.go
```diff
@@ -392,7 +392,7 @@ func (w *bitcoinWallet) getAddress(
 		data.ServerPubKey,
 		common.RelativeLocktime{
 			Type:  data.UnilateralExitDelay.Type,
-			Value: data.BoardingExitDelay.Value,
+			Value: data.UnilateralExitDelay.Value * 2,
```
Kukks: Confused wit h this change? Why  not use the new BoardingDescriptorTemplate?

pkg/client-sdk/wasm/browser/wrappers.go
```diff
@@ -520,15 +520,17 @@ func RedeemNotesWrapper() js.Func {
 	})
 }
 
-func SetNostrNotificationRecipientWrapper() js.Func {
+func RecoverSweptVtxosWrapper() js.Func {
```
Kukks, line 523: would remove as mentioned elsewhere

common/bip322/signature.go
```diff
@@ -0,0 +1,141 @@
...
+func (s *Signature) GetOutpoints() []wire.OutPoint {
+	outpoints := make([]wire.OutPoint, 0, len(s.TxIn)-1)
+	for _, input := range s.TxIn[1:] {
+		outpoints = append(outpoints, input.PreviousOutPoint)
+	}
+	return outpoints
+}
+
+func (s *Signature) ContainsOutputs() bool {
+	if len(s.TxOut) > 0 {
```
Kukks: if no outputs= false
if 1 outputs and out[0] ==opReturnPkScript = false
else true

Kukks (changes requested):

Kukks (commented):
We should 
* remove dedicated recover sdk/client and just use the same Settle method, but introduce granular coin control if need be
* use notes as fake inputs in bip322, to converge into a singular source of truth in the intent
* registerOutputs, potentially remove for now?

api-spec/protobuf/ark/v1/service.proto
louisinger: the "problem" is BIP322 expecting at least one "real" utxo to be signed. If the intent contains only notes, the bip322 proof will be an unsigned, is it a problem ?

client/main.go
louisinger: fixed

pkg/client-sdk/wallet/singlekey/bitcoin_wallet.go
louisinger: I messed with merge, I've reverted to `next-version` branch file

api-spec/protobuf/ark/v1/service.proto
Kukks: just the existence of knowing the note should be enough? Plus, we can technically issue notes that are locked to a script with this proposal?

Kukks: Technically a note is a preimage, we can migrate existing notes to a script that uses a hashlock

common/bip322/message.go
louisinger: done

pkg/client-sdk/covenantless_client.go
```diff
@@ -1433,11 +1299,81 @@ func (a *covenantlessArkClient) completeUnilateralExit(
 	return ptx.B64Encode()
 }
 
+func (a *covenantlessArkClient) Select(
```
altafan: Doesn't need to be exported.
```suggestion
func (a *covenantlessArkClient) selectFunds(
```

server/internal/core/application/covenantless.go
```diff
@@ -30,7 +31,10 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const marketHourDelta = 5 * time.Minute
+const (
+	marketHourDelta   = 5 * time.Minute
+	proofOfFundsDelta = 15 * time.Second
```
altafan: not needed, revert the changes to this line

server/internal/core/application/covenantless.go
```diff
@@ -717,26 +757,31 @@ func (s *covenantlessService) SpendVtxos(ctx context.Context, inputs []ports.Inp
 				}
 
 				// if the exit path is available, forbid registering the boarding utxo
-				if blocktime+exitDelay.Seconds() < now {
-					return "", fmt.Errorf("tx %s expired", input.Txid)
+				if blocktime+exitDelay.Seconds() < time.Now().Unix() {
```
altafan: USE TIME

server/internal/core/application/utils.go
```diff
@@ -108,12 +109,22 @@ func (m *txRequestsQueue) push(
 		}
 	}
 
+	for _, vtxo := range recoveredVtxos {
+		for _, request := range m.requests {
+			for _, pVtxo := range request.recoveredVtxos {
+				if vtxo == pVtxo {
```
altafan: Check only the outpoint

```suggestion
				if vtxo.VtxoKey == pVtxo.VtxoKey {
```

server/internal/interface/grpc/permissions/permissions.go
```diff
@@ -178,6 +170,10 @@ func Whitelist() map[string][]bakery.Op {
 			Entity: EntityExplorer,
 			Action: "read",
 		}},
+		fmt.Sprintf("/%s/GetNote", arkv1.ArkService_ServiceDesc.ServiceName): {{
+			Entity: EntityArk,
+			Action: "write",
+		}},
```
altafan: dunno why but this admin RPC must not be added to whitelist!

api-spec/protobuf/ark/v1/service.proto
altafan: we do it in another pr, this became too big already

client/main.go
louisinger: https://github.com/ark-network/ark/issues/541

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -123,11 +109,13 @@ message GetBoardingAddressResponse {
 
 /* In-Round Transaction API messages */
 
-message RegisterInputsForNextRoundRequest {
-  repeated Input inputs = 1;
+message RegisterIntentRequest {
```
tiero, line 119: lets not replace it, and only add a new RPC

Kukks: > lets not replace it, and only add a new RPC

Ideally we do remove it in the medium term, it's a redundant rpc  and keeps duplicate logic around input proofs

altafan (approved):

PR #493, opened 2025-03-06
Telemetry - Go runtime metrics

sekulicd:
This adds another ‘telemetry’ layer that will help us have better observability of arkd process it self.
We already have next metrics/dashboards: hostmetrics which gathers metrics for whole machine(cpu, RAM/Disc memory, Network/Disc I/O), we have arkd GPRC latency, count, request size, recently i added Cadvisor that tracks stats for each docker container running in machine separately and now I added metrics that will give us better insight into arkd go process it self. Check [here](https://github.com/sekulicd/ark/blob/7108217930ffe837cedf262d1fa3ad161747cc62/server/internal/interface/grpc/otel.go#L21) metrics i decided to track after investigation, goal is to inspect possible memory, go routines leaks, does gc behaves problematic, there is also metric tracking lock time for mutexes and num of cgo calls that could maybe indicate some places for future optimizations.

@altafan @tiero please take a look.

### DIFF
--- a/server/internal/interface/grpc/otel.go
+++ b/server/internal/interface/grpc/otel.go
@@ -19,8 +19,6 @@ import (
 )
 
 var arkRuntimeMetrics = []string{
-	//Counts how many calls Go made into C code (via cgo). Useful for diagnosing heavy cgo overhead
-	"/cgo/go-to-c-calls:calls",
 	//CPU time in user Go code (not GC or idle). Compare with GC CPU to see if GC is dominating.
 	"/cpu/classes/user:cpu-seconds",
 	//CPU time spent in garbage collection. Helps detect if GC overhead is large.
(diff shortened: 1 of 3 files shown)


server/internal/interface/grpc/otel.go
```diff
@@ -0,0 +1,252 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+	"runtime/metrics"
+	"strings"
+	"sync"
+	"time"
+)
+
+var arkRuntimeMetrics = []string{
+	//Counts how many calls Go made into C code (via cgo). Useful for diagnosing heavy cgo overhead
+	"/cgo/go-to-c-calls:calls",
```
tiero: dont think we use cgo

sekulicd: i am in go-elements world still :)

sekulicd: Removed

altafan (approved):

PR #495, opened 2025-03-13
[SDK] Detect rbf txs while listening for boarding utxos

altafan:
This makes the SDK client recognize rbf txs when used in stateful mode (`WithTransactionFeed: true`).

Before this, the SDK would have kept adding new records to the tx history incorrectly instead of replacing an existing one with the rbf-ed tx.

Checking if a tx has been rbf-ed is straight-forward when using [mempool apis](https://mempool.space/signet/docs/api/rest#get-mempool-rbf).

Doing the same with esplora is more challenging (e.g. in regtest): the key point is that once a tx is replaced and evicted from the mempool, esplora returns 404 when querying it.
[Here](https://github.com/ark-network/ark/blob/txhistory-handle-rbf/pkg/client-sdk/explorer/explorer.go#L386-L418) we leverage this to understand if a tx has been replaced, and we use the [outspends endpoint](https://github.com/Blockstream/esplora/blob/master/API.md#get-txtxidoutspends) to know the txid of the new tx spending the inputs of the old one.
But that's the catch: we can't ask the explorer for the tx hex of a replaced tx as it returns 404.
For this reason, it's been necessary to make the SDK client persist also the hex of a tx and to make this easier, the server now includes the hex of a round/redeem tx when sending a message on the transaction stream.

SIDE CHANGES:
* In the server, the hex of a round or redeem tx is now included in the message sent over transaction stream
* In the SDK, changed default explorer URLs to mempool instead of esplora (used only in regtest)
* In the SDK, fixed the order of unconfirmed txs in the history from last to first positions in the list

Please @bordalix @louisinger @sekulicd  review.


pkg/client-sdk/client.go
```diff
@@ -45,10 +45,10 @@ var (
 		common.Liquid.Name:         "https://blockstream.info/liquid/api",
 		common.LiquidTestNet.Name:  "https://blockstream.info/liquidtestnet/api",
 		common.LiquidRegTest.Name:  "http://localhost:3001",
-		common.Bitcoin.Name:        "https://blockstream.info/api",
-		common.BitcoinTestNet.Name: "https://blockstream.info/testnet/api",
+		common.Bitcoin.Name:        "https://mempool.space/api",
+		common.BitcoinTestNet.Name: "https://mempool.space/testnet/api",
 		//common.BitcoinTestNet4.Name: "https://mempool.space/testnet4/api", //TODO uncomment once supported
-		common.BitcoinSigNet.Name:    "https://blockstream.info/signet/api",
+		common.BitcoinSigNet.Name:    "https://mempool.space/signet/api",
```
louisinger, line 51: `common.BitcoinMutinyNet.Name` mutinyNet does not need the mempool APIs ?

altafan: it's already there.. look the whole file :)

louisinger (approved):

PR #496, opened 2025-03-14
chore: make function comment match function name

avoidalone:
make function comment match function name


tiero (commented):
The kvdb and macaroons package are copy pasted from lnd AFAIK @altafan

avoidalone:
> The kvdb and macaroons package are copy pasted from lnd AFAIK @altafan



> The kvdb and macaroons package are copy pasted from lnd AFAIK @altafan

Should I roll back my changes to this library?

altafan (approved):

tiero:
Apparently we also exported stuff on our own, so your changes are legitimate

PR #497, opened 2025-03-17
Skip waiting for signed boarding ins/forfeit txs if none is required

### LINKED ISSUE #492: Do not update round tx if no signed boarding inputs have been submitted
During round execution, if a client calls SubmitSignedRoundTx and passes the unsigned round tx as it is, causes the server to update the round tx and also to check if the stage is completed.

We MUST prevent this by simply adding a check to ensure that boarding ins are actually signed before [this](https://github.com/ark-network/ark/blob/master/server/internal/core/application/covenantless.go#L785-L792) as it's cause of weird and unpredictable behaviors

altafan:
This includes server-side optimization for the execution of the round:
- Do not wait for the execution timeout to expire when a round doesn't contain boarding inputs or forfeit txs to be signed: this is a very edge case in which a round includes only users that are redeeming notes. In this case, users just need to sign the VTXO tree to complete the round execution. Before this, in such case, the server would have waited for the timeout to expire because clients wouldn't submit any signed boarding input or forfeit tx.
- Add safe checks to [`SignRoundTx`](https://github.com/ark-network/ark/blob/master/server/internal/core/application/covenantless.go#L776) to prevent overriding a round tx if the user didn't submit any signed boarding input. Before this, a malicious client could have submitted an unsigned round tx to make the server loose all previous collected signed boarding inputs.

Closes #492.

### DIFF
--- a/server/internal/core/application/covenantless.go
+++ b/server/internal/core/application/covenantless.go
@@ -778,18 +778,21 @@ func (s *covenantlessService) SignVtxos(ctx context.Context, forfeitTxs []string
 }
 
 func (s *covenantlessService) SignRoundTx(ctx context.Context, signedRoundTx string) error {
+	numSignedInputs, err := s.builder.CountSignedTaprootInputs(signedRoundTx)
+	if err != nil {
+		return fmt.Errorf("failed to count number of signed boarding inputs: %s", err)
+	}
+	if numSignedInputs == 0 {
+		return nil
+	}
+
 	s.currentRoundLock.Lock()
 	defer s.currentRoundLock.Unlock()
 	currentRound := s.currentRound
 
-	numSignedInputs, combined, err := s.builder.VerifyAndCombinePartialTx(
-		currentRound.UnsignedTx, signedRoundTx,
-	)
+	combined, err := s.builder.VerifyAndCombinePartialTx(currentRound.UnsignedTx, signedRoundTx)
 	if err != nil {
 		return fmt.Errorf("failed to verify and combine partial tx: %s", err)
-	}
-	if numSignedInputs == 0 {
-		return nil
 	}
 
 	s.currentRound.UnsignedTx = combined
@@ -1487,7 +1490,8 @@ func (s *covenantlessService) finalizeRound(notes []note.Note, roundEndTime time
 	}
 	includesBoardingInputs := false
 	for _, in := range roundTx.Inputs {
-		if len(in.TaprootLeafScript) > 0 {
+		scriptType := txscript.GetScriptClass(in.WitnessUtxo.PkScript)
+		if scriptType == txscript.WitnessV1TaprootTy {
 			includesBoardingInputs = true
 			break
 		}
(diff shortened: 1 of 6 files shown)


server/internal/core/application/covenantless.go
```diff
@@ -778,10 +782,15 @@ func (s *covenantlessService) SignRoundTx(ctx context.Context, signedRoundTx str
 	defer s.currentRoundLock.Unlock()
 	currentRound := s.currentRound
 
-	combined, err := s.builder.VerifyAndCombinePartialTx(currentRound.UnsignedTx, signedRoundTx)
+	numSignedInputs, combined, err := s.builder.VerifyAndCombinePartialTx(
```
louisinger: Suggestion:
can we count the numSignedInputs from the psbt sent by the client BEFORE locking the current round mutex? This way we prevent a user spamming unsigned tx to "block" the others

altafan: not sure it's possible with the given design of `VerifyAndCombinePartialTx` as it requires a source and destination tx

altafan: solved by adding a new api to the tx builder

server/internal/infrastructure/tx-builder/covenant/builder.go
```diff
@@ -1071,7 +1087,11 @@ func (b *txBuilder) VerifyAndCombinePartialTx(dest string, src string) (string,
 		}
 	}
 
-	return roundSigner.Pset.ToBase64()
+	b64, err := roundSigner.Pset.ToBase64()
```
louisinger:
```suggestion
	return roundSigner.Pset.ToBase64()
```
what was the problem with this ?

server/internal/infrastructure/tx-builder/covenantless/builder.go
```diff
@@ -1103,7 +1118,11 @@ func (b *txBuilder) VerifyAndCombinePartialTx(dest string, src string) (string,
 		}
 	}
 
-	return roundTx.B64Encode()
+	b64, err := roundTx.B64Encode()
```
louisinger:
```suggestion
	return roundTx.B64Encode()
```

louisinger (approved):

server/internal/core/application/covenantless.go
```diff
@@ -1469,68 +1481,92 @@ func (s *covenantlessService) finalizeRound(notes []note.Note, roundEndTime time
...
 	if err != nil {
 		log.Debugf("failed to parse round tx: %s", round.UnsignedTx)
 		changes = round.Fail(fmt.Errorf("failed to parse round tx: %s", err))
 		log.WithError(err).Warn("failed to parse round tx")
 		return
 	}
+	includesBoardingInputs := false
+	for _, in := range roundTx.Inputs {
+		scriptType := txscript.GetScriptClass(in.WitnessUtxo.PkScript)
+		if scriptType == txscript.WitnessV1TaprootTy {
```
louisinger, line 1497: nit: this is not enough if the server is using taproot utxos in the future, maybe worth to add a comment?

PR #502, opened 2025-03-21
SDK - refresh tx db

sekulicd:
This updates listenForArkTxs so that before connecting to stream TX db is refreshed.
This should prevent losing of txs in case wallet using SDK, not online while there are TX activities.

@altafan please review.

### DIFF
--- a/pkg/client-sdk/covenantless_client.go
+++ b/pkg/client-sdk/covenantless_client.go
@@ -188,7 +188,7 @@ func LoadCovenantlessClient(sdkStore types.Store) (ArkClient, error) {
 	if cfgData.WithTransactionFeed {
 		txStreamCtx, txStreamCtxCancel := context.WithCancel(context.Background())
 		covenantlessClient.txStreamCtxCancel = txStreamCtxCancel
-		if err := covenantlessClient.refreshTxDb(context.Background()); err != nil {
+		if err := covenantlessClient.refreshDb(context.Background()); err != nil {
 			return nil, err
 		}
 		go covenantlessClient.listenForArkTxs(txStreamCtx)
@@ -242,7 +242,7 @@ func LoadCovenantlessClientWithWallet(
 	if cfgData.WithTransactionFeed {
 		txStreamCtx, txStreamCtxCancel := context.WithCancel(context.Background())
 		covenantlessClient.txStreamCtxCancel = txStreamCtxCancel
-		if err := covenantlessClient.refreshTxDb(context.Background()); err != nil {
+		if err := covenantlessClient.refreshDb(context.Background()); err != nil {
 			return nil, err
 		}
 		go covenantlessClient.listenForArkTxs(txStreamCtx)
@@ -260,7 +260,7 @@ func (a *covenantlessArkClient) Init(ctx context.Context, args InitArgs) error {
 	if args.WithTransactionFeed {
 		txStreamCtx, txStreamCtxCancel := context.WithCancel(context.Background())
 		a.txStreamCtxCancel = txStreamCtxCancel
-		if err := a.refreshTxDb(context.Background()); err != nil {
+		if err := a.refreshDb(context.Background()); err != nil {
 			return err
 		}
 		go a.listenForArkTxs(txStreamCtx)
@@ -278,7 +278,7 @@ func (a *covenantlessArkClient) InitWithWallet(ctx context.Context, args InitWit
 	if a.WithTransactionFeed {
 		txStreamCtx, txStreamCtxCancel := context.WithCancel(context.Background())
 		a.txStreamCtxCancel = txStreamCtxCancel
-		if err := a.refreshTxDb(context.Background()); err != nil {
+		if err := a.refreshDb(context.Background()); err != nil {
 			return err
 		}
 		go a.listenForArkTxs(txStreamCtx)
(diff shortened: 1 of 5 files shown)


altafan (changes requested):
deleting the whole history and re-add it might be an heavy task if it's a very long one. Better to update or add txs to the transaction store. Also, are missing to update the vtxo store

pkg/client-sdk/covenantless_client.go
```diff
@@ -329,6 +334,155 @@ func (a *covenantlessArkClient) listenForArkTxs(ctx context.Context) {
...
+	oldTxsMap := make(map[string]types.Transaction, len(oldTxs))
+	for _, tx := range oldTxs {
+		oldTxsMap[tx.TransactionKey.String()] = tx
+	}
+
+	// add new boarding, update existing that are not confirmed
+	boardingTxsToAdd := make([]types.Transaction, 0)
+	boardingTxsToConfirm := make([]types.Transaction, 0)
+	for _, tx := range boardingTxs {
+		oldTx, exists := oldTxsMap[tx.Hex]
```
altafan:
```suggestion
		oldTx, exists := oldTxsMap[tx.TransactionKey.String()]
```

pkg/client-sdk/covenantless_client.go
```diff
@@ -329,6 +334,155 @@ func (a *covenantlessArkClient) listenForArkTxs(ctx context.Context) {
...
+		if !exists {
+			boardingTxsToAdd = append(boardingTxsToAdd, tx)
+		} else {
+			if oldTx.CreatedAt.IsZero() && !tx.CreatedAt.IsZero() {
+				boardingTxsToConfirm = append(boardingTxsToConfirm, tx)
+			}
```
altafan: we also need to keep track of the txs to settle: if an old tx is not settled and the new one is, we should add it to the list `boardingTxsToSettle`.

There's an edge case here: we may need to take care of the case where a tx is unconfirmed in db, while the one "fetched form the explorer" is settled. Such txs needs to have updated both the CreatedAt and Settled fields, the event produced for this is ok to be tx_settled.

pkg/client-sdk/covenantless_client.go
```diff
@@ -329,6 +334,155 @@ func (a *covenantlessArkClient) listenForArkTxs(ctx context.Context) {
...
+		if _, err := a.store.TransactionStore().ConfirmTransactions(
+			ctx, []string{tx.TransactionKey.String()}, tx.CreatedAt,
+		); err != nil {
+			return err
+		}
+	}
+
+	if _, err := a.store.TransactionStore().SettleTransactions(ctx, txsToSettle); err != nil {
+		return err
+	}
```
altafan: nest all these ifs into others like:
```
if len(allTxsToAdd) > 0 { ... }
if len(txsToSettle) > 0 { ... }
```

to prevent making useless db calls

pkg/client-sdk/covenantless_client.go
```diff
@@ -277,6 +277,11 @@ func (a *covenantlessArkClient) InitWithWallet(ctx context.Context, args InitWit
 }
 
 func (a *covenantlessArkClient) listenForArkTxs(ctx context.Context) {
+	if err := a.refreshTxDb(ctx); err != nil {
+		log.WithError(err).Error("failed to refreshTxDb")
+		return
+	}
```
altafan: `listenForArkTxs` runs as a go routine, and this way we would log an error and stop its execution, but the SDK client would still be returned as usable to the user. Better to move this outside, this way if this process fails we return error to the user.

pkg/client-sdk/covenantless_client.go
```diff
@@ -329,6 +334,155 @@ func (a *covenantlessArkClient) listenForArkTxs(ctx context.Context) {
...
+		); err != nil {
+			return err
+		}
+	}
+
+	if _, err := a.store.TransactionStore().SettleTransactions(ctx, txsToSettle); err != nil {
+		return err
+	}
+
+	oldSpendableVtxos, _, err := a.store.VtxoStore().GetAllVtxos(ctx)
```
altafan: the logic below sounds ok but i think it missed to handle a case: the vtxos fetched from db miss a SPENT vtxo fetched from the explorer. This vtxo would never be added to the db, right?

pkg/client-sdk/covenantless_client.go
sekulicd: Done

pkg/client-sdk/covenantless_client.go
sekulicd: Done

pkg/client-sdk/covenantless_client.go
sekulicd: Done

pkg/client-sdk/covenantless_client.go
sekulicd: Done

pkg/client-sdk/covenantless_client.go
sekulicd: > There's an edge case here: we may need to take care of the case where a tx is unconfirmed in db, while the one "fetched form the explorer" is settled. Such txs needs to have updated both the CreatedAt and Settled fields, the event produced for this is ok to be tx_settled.

we dont have repo method to update both fileds, shuold we create it? Smtn like UpdateTx ?

altafan (approved):

PR #504, opened 2025-03-24
show market hour on sdk config

bordalix:
In order for clients to calculate the best market hour for future claims, its helpful to know when the server is expecting to run a happy hour. The server already gives that info on /v1/info. This PR makes that information visible in the SDK.

@altafan please review


altafan (changes requested):
Updates required in pkg/client-sdk/store/file

altafan (approved):
tACK 19ed54242be335468fc28f2a94f58007d2d063e2

PR #505, opened 2025-03-25
VtxoScript validation when spending a vtxo

louisinger:
_This PR add robust checks on `SubmitRedeemTx` and `RegisterInputsForNextRound` RPCs. The idea is to validate the revealed tapscripts of every spent VTXO in order to prevent a malicious user to inject a "backdoor" tapscript._

### Settlement

The current `RegisterInputsForNextRoundRequest` has `Tapscripts` field on inputs. This PR only validates the scripts.

### Redeem tx

This PR introduces a **breaking change** for client, because now redeem PSBTs must reveal the inputs tapscripts using an unknown PSBT field "tapscripts".

_Why unknown field instead of TaprootLeafScript PSBT field ?_

We are using this field to signal the leaf used to sign, like most wallets do. Revealing the taproot tree using this field implies a lot of change on critical signing code (both clients and servers). Unknown field seems more adapted because the scripts are only used for validation. Also, it reduces the size of the PSBT.

@Kukks @altafan please review

### DIFF
--- a/common/bitcointree/psbt.go
+++ b/common/bitcointree/psbt.go
@@ -8,6 +8,7 @@ import (
 	"github.com/ark-network/ark/common"
 	"github.com/ark-network/ark/common/tree"
 	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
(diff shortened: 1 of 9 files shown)


Kukks (commented):
Should we use the format described in PSBT_OUT_TAP_TREE in bip371?

common/bitcointree/psbt.go
```diff
@@ -126,3 +214,57 @@ func cosignerPrefixedKey(index int) []byte {
...
 }
+
+// readCompactSizeUint reads a compact size uint from the reader
+func readCompactSizeUint(r *bytes.Reader) (uint64, error) {
+	firstByte, err := r.ReadByte()
+	if err != nil {
+		return 0, err
+	}
+
+	switch firstByte {
```
altafan, line 228: can you explain why this switch is needed? can't we just always do?
```
var val uint64
if err := binary.Read(r, binary.LittleEndian, &val); err != nil {
	return 0, err
}
return val, nil
```

PR #507, opened 2025-03-28
Add SubscribeForAddress explorer API

### LINKED ISSUE #506: Electrum-like endpoints
To lower the impact that client have on ListVtxos, let's add apis to the explorer for (un)subscribing to addresses and get notifications

altafan:
This adds a new explorer API to subscribe for an address and get notified about new or spent vtxos in real-time.

Closes #506 

Please @louisinger @bordalix review.


tiero:
LGTM🛸

api-spec/protobuf/ark/v1/explorer.proto
```diff
@@ -21,6 +21,11 @@ service ExplorerService {
       get: "/v1/vtxos/{address}"
     };
   };
+  rpc SubscribeForAddress(SubscribeForAddressRequest) returns (stream SubscribeForAddressResponse) {
```
louisinger, line 24: It means 1 stream / address, should we allow the client to subscribe several addresses in the same RPC ?

api-spec/protobuf/ark/v1/explorer.proto
```diff
@@ -44,3 +49,11 @@ message ListVtxosResponse {
   repeated Vtxo spendable_vtxos = 1;
   repeated Vtxo spent_vtxos = 2;
 }
+
+message SubscribeForAddressRequest {
+  string address = 1;
+}
+message SubscribeForAddressResponse {
```
louisinger, line 56: Should we also return the following events: 
* when a vtxo is swept ?
* when a vtxo is unilaterally put onchain ?

pkg/client-sdk/client.go
```diff
@@ -142,6 +143,35 @@ func (a *arkClient) ListVtxos(
...
+func (a *arkClient) NotifyIncomingFunds(
+	ctx context.Context, addr string,
+) ([]types.Vtxo, error) {
+	eventCh, closeFn, err := a.client.SubscribeForAddress(ctx, addr)
+	if err != nil {
+		return nil, err
+	}
+
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
```
louisinger, line 155: This `wg` is weird:
```
wg.Add(1)
go func() ... 
wg.Done()
```

it seems `Wait` is never called and `Done` is called just after starting the routine, sounds we don't need it

api-spec/protobuf/ark/v1/explorer.proto
tiero: Currently we support single key wallets in our clients, don't think is necessary at the moment?

louisinger: Yes, but in theory fulmine has also to call this method for VHTLC?

tiero: They are commit-reveal type of contract, you subscribe once, wait for being "spent" and then you unsubscribe. 
Rarely you fund multiple contracts at same time. I'm not against it, just wondering if is worth as we don't have strong need for now

Kukks: We can always extend via a new rpc later `SubscribeForAddresses`

PR #512, opened 2025-03-31
[Server] Move db write ops to background

altafan:
This is about server-side optmization.

Specifically, this makes all the APIs that write to the db do such an operation in the background and not affect the response returned to the user.

In the background, if the write to the db fails we wait 100 milliseconds and rewrite. We basically assume the writes to return error only because of conflicts because we performed read operations upfront, and the domains muted successfully, so we are 100% sure the writes must succeed.

Please @louisinger @sekulicd @tiero review.

### DIFF
--- a/server/internal/core/application/covenantless.go
+++ b/server/internal/core/application/covenantless.go
@@ -524,54 +524,58 @@ func (s *covenantlessService) SubmitRedeemTx(
 		return "", "", fmt.Errorf("failed to sign redeem tx: %s", err)
 	}
 
-	// Create new vtxos, update spent vtxos state
-	newVtxos := make([]domain.Vtxo, 0, len(ptx.UnsignedTx.TxOut))
-	for outIndex, out := range outputs {
-		vtxoTapKey, err := schnorr.ParsePubKey(out.PkScript[2:])
-		if err != nil {
-			return "", "", fmt.Errorf("failed to parse vtxo taproot key: %s", err)
+	go func(ptx *psbt.Packet, signedRedeemTx, redeemTxid string) {
+		// Create new vtxos, update spent vtxos state
+		newVtxos := make([]domain.Vtxo, 0, len(ptx.UnsignedTx.TxOut))
+		for outIndex, out := range outputs {
+			//notlint:all
+			vtxoPubkey := hex.EncodeToString(out.PkScript[2:])
+
+			newVtxos = append(newVtxos, domain.Vtxo{
+				VtxoKey: domain.VtxoKey{
+					Txid: redeemTxid,
+					VOut: uint32(outIndex),
+				},
+				PubKey:    vtxoPubkey,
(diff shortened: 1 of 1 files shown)


altafan:
I need to slow done all the tests because the client now needs to give the server the time to finalize the db writing operation done in the background. That's the reason for all failing tests

server/internal/core/application/covenantless.go
```diff
@@ -1267,7 +1281,29 @@ func (s *covenantlessService) DeleteNostrRecipient(ctx context.Context, signedVt
...
+		}
+		fetchedKeys = append(fetchedKeys, vtxoKey)
+	}
+	if len(fetchedKeys) <= 0 {
+		return nil
+	}
+
+	go func(vtxoKeys []domain.VtxoKey) {
+		err := s.repoManager.Entities().Delete(ctx, vtxoKeys)
+		for err != nil {
```
sekulicd: dont like this way of handling error, if we really want to do this optimization, we should have proper queue mechanism with finite retry attempts in repo layer, also assumes that error will ‘disappear’ after x num of retries, even if 'We basically assume the writes to return error only because of conflicts’ true, code hides this logic, adding comment would help but still dont feel this is the way.

sekulicd (commented):
You did changes in:
SubmitRedeemTx -> offloading looping through outs, AddVtxos and SpendVtxos in go routine, we shouldnt have big delay, its just few outs, and two fast update/insert query
RegisterCosignerNonces -> seems like idea is to remove delay introduced by mutex, but this as well shoulnd be so big delay, also RegisterCosignerNonces belongs to round flow where user anyways needs to wait for server to send next event, so even if client spares some time it will hang doing nothing until receives new event
RegisterCosigenrSignatures -> same as for RegisterCosigenrNonces
DeleteNostrReceipient -> similar to SubmitRedeemTx

server/internal/core/application/covenantless.go
altafan: Agreed. I'm gonna move the retry logic at repo level

pkg/client-sdk/test/wasm/wasm_test.go
```diff
@@ -350,6 +368,18 @@ func sendOffChain(page playwright.Page, addr string, amount int) error {
 	return err
 }
 
+func waitForPayment(page playwright.Page, addr string) error {
+	_, err := page.Evaluate(fmt.Sprintf(`async () => {
+			try {
+            return await notifyIncomingFunds("%s");
```
louisinger, line 374:
```suggestion
            return notifyIncomingFunds("%s");
```

louisinger: nit

server/internal/core/application/covenantless.go
altafan: done

PR #518, opened 2025-04-03
Implements CleanData interface

bordalix:
This PR implements CleanData() interface on tx and vtxo DBs.
Uses this to clean tx and vtxo db on wallet init.
@altafan please review


bordalix:
ACK

PR #524, opened 2025-04-08
Indexer

sekulicd:
This implements rpc’s from indexer.proto and adds those methods to sdk client(grpc/rest) 

@altafan please review.


server/internal/core/application/indexer.go
```diff
@@ -0,0 +1,553 @@
...
+	return &indexerService{
+		pubkey:      pubkey,
+		repoManager: repoManager,
+	}
+}
+
+func (i *indexerService) GetCommitmentTxInfo(
+	ctx context.Context, txid string,
+) (*CommitmentTxResp, error) {
+	round, err := i.repoManager.Rounds().GetRoundWithTxid(ctx, txid)
```
altafan: This API can take time if the database and the round are big, we should use an ad hoc api that basically returns the data as defined in the proto.

server/internal/core/application/indexer.go
```diff
@@ -0,0 +1,553 @@
...
+	nodes, pageResp := paginate(flattenNodes(vtxoTree), req.Page, maxPageSizeVtxoTree)
+
+	return &VtxoTreeResp{
+		Nodes: nodes,
+		Page:  pageResp,
+	}, nil
+}
+
+func (i *indexerService) GetForfeitTxs(ctx context.Context, req ForfeitTxsReq) (*ForfeitTxsResp, error) {
+	round, err := i.repoManager.Rounds().GetRoundWithTxid(ctx, req.BatchOutpoint.Txid) //TODO batch thing
```
altafan: let's make use of an ad hoc API `GetRoundForfeitTxs` that returns only the list of forfeit txids for a given round txid

server/internal/core/application/indexer.go
```diff
@@ -0,0 +1,553 @@
...
+
+	return &ForfeitTxsResp{
+		Txs:  forfeitTxs,
+		Page: pageResp,
+	}, nil
+
+}
+
+func (i *indexerService) GetConnectors(ctx context.Context, req ConnectorsReq) (*ConnectorResp, error) {
+	round, err := i.repoManager.Rounds().GetRoundWithTxid(ctx, req.BatchOutpoint.Txid) //TODO batch thing
```
altafan: same here, we can make use of an ad hoc `GetRoundConnectorTree`

server/internal/core/application/indexer.go
```diff
@@ -0,0 +1,553 @@
...
+	spendableVtxosPaged, pageResp := paginate(spendableVtxos, req.Page, maxPageSizeSpendableVtxos)
+
+	return &SpendableVtxosResp{
+		Vtxos: spendableVtxosPaged,
+		Page:  pageResp,
+	}, nil
+}
+
+func (i *indexerService) GetTransactionHistory(ctx context.Context, req TxHistoryReq) (*TxHistoryResp, error) {
+	allVtxos, err := i.repoManager.Vtxos().GetAll(ctx)
```
altafan: let's make use of an ad hoc API `GetVtxosWithPubKey` instead of getting the whole vtxo set and then filter by pubkey. The code to extract the pubkey from the address can be taken from the func above

server/internal/core/application/indexer.go
```diff
@@ -0,0 +1,553 @@
...
+		if err := i.buildChain(ctx, parentOutpoint, chain, false); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (i *indexerService) GetVirtualTxs(ctx context.Context, req VirtualTxsReq) (*VirtualTxsResp, error) {
+	vtxos, err := i.repoManager.Vtxos().GetAll(ctx)
```
altafan, line 313: let's make use of an ad hoc API `GetTxsWithTxids`. This API is easy to implement int he sqlite repo, harder for the badger one.. for the latter we can even return an error "not implemented" as it's meant only for testing purposes

server/internal/core/application/types.go
```diff
@@ -140,3 +140,142 @@ type TxRequestInfo struct {
 	Cosigners      []string
 	LastPing       time.Time
 }
+
+type VtxoTreeReq struct {
```
altafan: can we please drop all these `*Req` structs? I'm ok with the resp ones, but anyway, this indexer will stay here only for a bit hopefully... let's have the less possible impact in terms of changes to existing files.

server/internal/core/application/indexer.go
altafan: let's make use of the newly defined API `GetTxsWithTxids`.

server/internal/core/application/indexer.go
sekulicd: Done

server/internal/core/application/indexer.go
sekulicd: Done

server/internal/core/application/indexer.go
sekulicd: Done

server/internal/core/application/indexer.go
sekulicd: Done

server/internal/core/application/types.go
sekulicd: Done

server/internal/core/application/indexer.go
sekulicd: Done

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -0,0 +1,222 @@
...
+    option (google.api.http) = {
+      get: "/v1/batch/{batch_outpoint.txid}/{batch_outpoint.vout}/forfeitTxs"
+    };
+  };
+  rpc GetConnectors(GetConnectorsRequest) returns (GetConnectorsResponse) {
+    option (google.api.http) = {
+      get: "/v1/batch/{batch_outpoint.txid}/{batch_outpoint.vout}/connectors"
+    };
+  };
+  rpc GetSpendableVtxos(GetSpendableVtxosRequest) returns (GetSpendableVtxosResponse) {
```
Kukks, line 28: I personally find this call a bit limiting. Why not be `GetVtxos` and `GetVtxosRequest`  would have extra filters ?

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -0,0 +1,222 @@
...
+  bool is_spent = 8;
+  string spent_by = 9;
+}
+
+message IndexerChain {
+  string txid = 1;
+  IndexerChainedTxType type = 2;
+}
+
+message IndexerTxHistoryRecord {
```
Kukks, line 188: This message confuses me quite a bit:

* Why would the key ever be a commitment tx?
* What is amount specifically here? Is it a "balance delta"?
*  Unsure how the `type` works: a transactions that spends a vtxo from `addressA` and sends some funds to someone, and change back to `addressA` would mean the type of this tx is both INDEXER_TX_TYPE_RECEIVED and INDEXER_TX_TYPE_SENT. No clue what INDEXER_TX_TYPE_SWEEP means 
* `created_at` is actually a cosigning request timestamp right?
* `confirmed_at` relates to its parent commitment tx(s)?
* `is_settled` Unsure what it means

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -0,0 +1,222 @@
...
+
+message IndexerTransactions {
+  repeated IndexerChain txs = 1;
+  int64 expires_at = 2;
+}
+
+message IndexerBatch {
+  uint64 total_batch_amount = 1;
+  uint64 total_forfeit_amount = 2;
+  int32 total_input_vtxos = 3;
```
Kukks, line 153: I thought we were removing this type (inputs) of data from the indexer? I would also move them from being batch specific to CommitmentTx level

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -0,0 +1,222 @@
...
+  IndexerOutpoint batch_outpoint = 1;
+  IndexerPageRequest page = 2;
+}
+message GetVtxoTreeResponse {
+  repeated IndexerNode vtxo_tree = 1;
+  IndexerPageResponse page = 2;
+}
+
+message GetForfeitTxsRequest {
+  IndexerOutpoint batch_outpoint = 1;
```
Kukks, line 74: Would do on a commitment tx level

api-spec/protobuf/ark/v1/indexer.proto
Kukks: I would introduce a way to fetch all effected vtxos with each record, something similar to the `SubscribeForAddressesResponse` message. I guess fits with my suggestion to change `GetSpendableVtxos` to `GetVtxos`

pkg/client-sdk/client/client.go
```diff
@@ -263,3 +274,125 @@ type AddressEvent struct {
...
+type IndexerVtxo struct {
+	Outpoint  Outpoint
+	CreatedAt int64
+	ExpiresAt int64
+	Amount    uint64
+	Script    string
+	IsLeaf    bool
+	IsSwept   bool
+	IsSpent   bool
+	SpentBy   string
```
Kukks, line 337: Right now, this propoerty being set is only exposed through `SubscribeForAddressesResponse`

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -0,0 +1,222 @@
...
+message GetSweptCommitmentTxResponse {
+  repeated string swept_by = 1;
+}
+
+message SubscribeForAddressesRequest {
+  repeated string addresses = 1;
+}
+message SubscribeForAddressesResponse {
+  string address = 1;
+  repeated IndexerVtxo new_vtxos = 2;
```
Kukks, line 141: maybe a repeated IndexerOutpoint or even just a txid and let the user fetch through API to keep the streams light?

Kukks (changes requested):

api-spec/protobuf/ark/v1/indexer.proto
altafan: we would have 2 lists in the response that would make the pagination tricky/impossible

api-spec/protobuf/ark/v1/indexer.proto
altafan: `total_input_vtxos` here refers to the count of vtxos forfeited during the batch process

altafan: i prefer redundancy here and have `total_input_vtxos` also in GetCommitmentTx as a cumulative sum

api-spec/protobuf/ark/v1/indexer.proto
altafan: yeah in pure electrum-like style, this endpoint should also accept script hashes instead of addresses

api-spec/protobuf/ark/v1/indexer.proto
Kukks: why  2 lists?

Kukks: the IndexerVtxo holds enough data that you can return a flat list

api-spec/protobuf/ark/v1/indexer.proto
Kukks: Let's say a multi batch tx happens and:
a user has a VTXO of 0.1 BTC but wants to get 2 VTXOs of 0.5 each.

These 0.5 VTXOs potentially end up in 2 different batches. There is only one forfeit.  Only a total_input_vtxos  on GetCommitmentTx  would make sense. 

When a round is being done, intents should be on the round, batches only dictate vtxo and tree signing, but the forfeiture is on a commitment tx scope?

PR #533, opened 2025-04-11
[SDK] Disable onchain features if server doesn't allow them

altafan:
In case the server returns `UtxoMaxAmount = 0` in GetInfo, this disables:
- generating boarding address
- watching for boarding tx records for history
- collaborative exit

Please @tiero review.

### DIFF
--- a/pkg/client-sdk/covenantless_client.go
+++ b/pkg/client-sdk/covenantless_client.go
@@ -293,6 +293,10 @@ func (a *covenantlessArkClient) InitWithWallet(ctx context.Context, args InitWit
 func (a *covenantlessArkClient) Balance(
 	ctx context.Context, computeVtxoExpiration bool,
 ) (*Balance, error) {
+	if err := a.safeCheck(); err != nil {
+		return nil, err
+	}
+
 	offchainAddrs, boardingAddrs, redeemAddrs, err := a.wallet.GetAddresses(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/client-sdk/covenantless_client.go
```diff
@@ -1002,10 +1013,16 @@ func (a *covenantlessArkClient) refreshDb(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	boardingTxs, roundsToIgnore, err := a.getBoardingTxs(ctx)
-	if err != nil {
-		return err
+
+	boardingTxs := make([]types.Transaction, 0)
+	roundsToIgnore := make(map[string]struct{})
+	if a.Config.UtxoMaxAmount != 0 {
```
bordalix: @altafan imagine the following scenario:
- user is using Fulmine for a few weeks, made a couple of boarding txs
- yesterday the server decided to stop allowing UTXO tx, so UtxoMaxAmount is now 0
- does this code means that today the user will stop seeing the past boarding txs in his history?

bordalix: Actually, this is what happens:
- I received 4 boarding txs of 21, 32, 44 and 50 k sats
- I received 2 offchain txs of 4 and 5 k sats
- Then I put UTXO_MAX_AMOUNT=0 on server and restore Fulmine wallet
- Now I see JUST ONE boarding tx with the total value of 147 k sats (the value is correct btw)

Screenshot 2025-04-21 at 11 03 25

altafan: Is it possible that you settled all 4 boardings?

Btw, the problem you pointed out is worth to be taken in consideration, but only for the restore: in this case we should ignore what the server says and restore all pending boarding funds.. the only possibility for the user will be to claim back those funds once the locktime expires if the server didn't re-allow boardings till that moment.

If the user sent funds to the boarding addresss when it was allowed and now the server disallowed them, he will receive an error like `operation not allowed`, and, again, he will only claim back the funds if the server didn't re-allow boardings by the time the loctime expires.

bordalix (approved):
tACK

PR #537, opened 2025-04-15
fix: prevent collision between in round and redeem vtxo

sekulicd:
## Problem
Previously, there was a potential race condition where a vtxo could be both redeemed via [SubmitRedeemTx](cci:1://file:///Users/dusansekulic/code/go/ark/server/internal/core/application/covenantless.go:223:0-606:1) and simultaneously registered for the next round via [SpendVtxos](cci:1://file:///Users/dusansekulic/code/go/ark/server/internal/core/application/covenantless.go:674:0-809:1). This could lead to double-spending attempts and inconsistent state.

## Solution
This PR implements bidirectional checks to prevent collisions between vtxos being redeemed and those being registered for the next round:

1. Added a new `redeemTxRequests` structure to track vtxos that are currently being redeemed
2. Enhanced [SubmitRedeemTx](cci:1://file:///Users/dusansekulic/code/go/ark/server/internal/core/application/covenantless.go:223:0-606:1) to:
   - Check if any vtxos are already registered for the next round
   - Track vtxos being redeemed in the `redeemTxRequests` structure
   - Remove vtxos from tracking after successful processing or in case of errors
3. Enhanced SpendVtxos to check if any vtxos are currently being redeemed before allowing them to be registered

@altafan please review.

### DIFF
--- a/server/internal/core/application/covenantless.go
+++ b/server/internal/core/application/covenantless.go
@@ -1384,6 +1384,7 @@ func (s *covenantlessService) startFinalization(roundEndTime time.Time) {
 
 	var notes []note.Note
 	var roundAborted bool
+	var vtxoKeys []domain.VtxoKey
 	defer func() {
 		delete(s.treeSigningSessions, round.Id)
 		if roundAborted {
@@ -1396,6 +1397,7 @@ func (s *covenantlessService) startFinalization(roundEndTime time.Time) {
 		}
 
 		if round.IsFailed() {
+			s.roundInputs.remove(vtxoKeys)
 			s.startRound()
 			return
 		}
@@ -1424,13 +1426,11 @@ func (s *covenantlessService) startFinalization(roundEndTime time.Time) {
 	}
 	requests, boardingInputs, redeeemedNotes, musig2data := s.txRequests.pop(num)
 	notes = redeeemedNotes
-	vtxoKeys := make([]domain.VtxoKey, 0)
 	for _, req := range requests {
 		for _, in := range req.Inputs {
 			vtxoKeys = append(vtxoKeys, in.VtxoKey)
 		}
 	}
-	s.roundInputs.remove(vtxoKeys)
 	s.numOfBoardingInputsMtx.Lock()
 	s.numOfBoardingInputs = len(boardingInputs)
 	s.numOfBoardingInputsMtx.Unlock()
@@ -1667,6 +1667,17 @@ func (s *covenantlessService) finalizeRound(notes []note.Note, roundEndTime time
 	s.currentRoundLock.Lock()
 	round := s.currentRound
 	s.currentRoundLock.Unlock()
+
+	defer func() {
+		vtxoKeys := make([]domain.VtxoKey, 0)
+		for _, req := range round.TxRequests {
+			for _, in := range req.Inputs {
+				vtxoKeys = append(vtxoKeys, in.VtxoKey)
+			}
+		}
+		s.roundInputs.remove(vtxoKeys)
+	}()
+
 	if round.IsFailed() {
 		return
 	}


server/internal/core/application/covenantless.go
```diff
@@ -252,6 +254,22 @@ func (s *covenantlessService) SubmitRedeemTx(
...
 
+	for _, vtxoKey := range spentVtxoKeys {
+		if s.txRequests.isVtxoRegisteredForNextRound(vtxoKey) {
+			return "", "", fmt.Errorf("vtxo %s is already registered for the next round", vtxoKey.String())
+		}
+	}
+
+	for _, vtxoKey := range spentVtxoKeys {
+		s.redeemTxRequests.addVtxo(vtxoKey)
+	}
```
bordalix:
```suggestion
	for _, vtxoKey := range spentVtxoKeys {
		if s.txRequests.isVtxoRegisteredForNextRound(vtxoKey) {
			return "", "", fmt.Errorf("vtxo %s is already registered for the next round", vtxoKey.String())
		}
		s.redeemTxRequests.addVtxo(vtxoKey)
	}
```

server/internal/core/application/utils.go
```diff
@@ -270,6 +270,21 @@ func (m *txRequestsQueue) view(id string) (*domain.TxRequest, bool) {
 	}, true
 }
 
+func (m *txRequestsQueue) isVtxoRegisteredForNextRound(vtxoKey domain.VtxoKey) bool {
```
altafan: this can be expensive if there are a lot of tx requests and end up slowing down the SubmitRedeemTx execution. I think it's better to use another map like `redeemTxRequests` to make this check O(1) instead of O(n)

server/internal/core/application/utils.go
```diff
@@ -411,6 +426,37 @@ func (m *forfeitTxsMap) allSigned() bool {
...
+	defer r.lock.Unlock()
+	delete(r.vtxos, vtxoKey.String())
+}
+
+func (r *redeemTxRequests) isVtxoRedeemed(vtxoKey domain.VtxoKey) bool {
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+	_, exists := r.vtxos[vtxoKey.String()]
+	return exists
+}
```
altafan: i'd make these methods accept list of vtxo keys this way we hide any for loop inside them

server/internal/core/application/utils.go
```diff
@@ -411,6 +426,37 @@ func (m *forfeitTxsMap) allSigned() bool {
 	return true
 }
 
+type redeemTxRequests struct {
```
altafan: given https://github.com/ark-network/ark/pull/537/files#r2044873760, i'd rename this util struct to a generic `outpointMap` that can be reused for both tx request and redeem tx inputs.

altafan (commented):
Also please, in `ark.trivy.yaml`, update `runs-on` to `ubuntu-24.04`

server/internal/core/application/covenantless.go
sekulicd: Done

server/internal/core/application/utils.go
sekulicd: Done

sekulicd:
> Let's add a test that ensures the double spend attempt won't succeed.

This cant be tested with consistent outcome as it is matter of timing.
We cant control SDK invocation timing for Settle and SendOffchain.

sekulicd:
> Also please, in `ark.trivy.yaml`, update `runs-on` to `ubuntu-24.04`

Done

server/internal/core/application/utils.go
sekulicd: Done

server/internal/core/application/utils.go
sekulicd: Done

server/internal/core/application/utils.go
```diff
@@ -32,14 +32,16 @@ type timedTxRequest struct {
 }
 
 type txRequestsQueue struct {
-	lock     *sync.RWMutex
-	requests map[string]*timedTxRequest
+	lock      *sync.RWMutex
+	requests  map[string]*timedTxRequest
+	vtxoIndex map[domain.VtxoKey]struct{}
```
altafan: can we move this directly to the app svc and make it become of type redeemTxRequests (that's why i suggested the renaming)? This way we not make this double-check operation dependant from the txRequestQueue that is meant for handling in-round swaps

server/internal/core/application/covenantless.go
```diff
@@ -1404,6 +1424,13 @@ func (s *covenantlessService) startFinalization(roundEndTime time.Time) {
 	}
 	requests, boardingInputs, redeeemedNotes, musig2data := s.txRequests.pop(num)
 	notes = redeeemedNotes
+	vtxoKeys := make([]domain.VtxoKey, 0)
+	for _, req := range requests {
+		for _, in := range req.Inputs {
+			vtxoKeys = append(vtxoKeys, in.VtxoKey)
+		}
+	}
+	s.roundInputs.remove(vtxoKeys)
```
louisinger: Why removing the vtxo keys before signing the tree ? we should also prevent anybody to submit a redeem tx during the signing process (the keys should be removed when the round is done or failed) ?

altafan: should be fixed

PR #563, opened 2025-05-15
RPC Delete intent

### LINKED ISSUE #557: Add RPC to delete intent
Right now, a client that registers an intent can only complete the process and cannot abort. Let's make it possible to delete a registered intent by providing any of the outpoints included in the previously signed bip322 proof.

sekulicd:
This adds DeleteIntent RPC.

This closes #557 

@altafan please review.


altafan (commented):
Let's add `DeleteIntet` to the `pkg/client-sdk/client` service interface and related grpc and rest impls

sekulicd:
> Let's also add the second way of deleting an intent: if the user forgot the intent id, they can provide a new bip322 proof spending any of the inputs spent by the registered intent to prove the ownership and the right to execute the operation.

@altafan please review.

server/internal/core/application/service.go
```diff
@@ -1404,10 +1404,100 @@ func (s *covenantlessService) DeleteTxRequests(
...
+
+	if err := sig.Verify(encodedMessage, prevoutFetcher); err != nil {
+		return fmt.Errorf("failed to verify signature: %s", err)
+	}
+
+	allRequests, err := s.txRequests.viewAll(nil)
+	if err != nil {
+		return err
+	}
+	var idsToDelete []string
```
altafan: should be a map to prevent duplicates

server/internal/core/application/service.go
```diff
@@ -1404,10 +1404,100 @@ func (s *covenantlessService) DeleteTxRequests(
...
+	for _, req := range allRequests {
+		for _, in := range req.Inputs {
+			for _, op := range outpoints {
+				if in.Txid == op.Hash.String() && in.VOut == op.Index {
+					idsToDelete = append(idsToDelete, req.Id)
+				}
```
altafan: the idea is to allow deleting only one intent at a time, not many. So, shall we break once we found the id to delete?

sekulicd: not sure about this, as user can craft proof with vtxos belonging to diff intents, maybe delete all? @altafan

server/internal/core/application/service.go
sekulicd: Done

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -119,6 +125,12 @@ message RegisterIntentResponse {
   string request_id = 1;
 }
 
+message DeleteIntentRequest {
+  string request_id = 1;
```
Kukks, line 129: these should be oneOf. EIther you provide the intentid (which is a user secret) , or you provide an input (a single input is enough) in a bip322. The reasoning is that intent id is trivial and straightforward, but in the case of a wallet recovery where you only have a private key, the bip322 option works.

server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -156,6 +156,45 @@ func (h *handler) RegisterIntent(
 	}, nil
 }
 
+func (h *handler) DeleteIntent(
+	ctx context.Context, req *arkv1.DeleteIntentRequest,
+) (*arkv1.DeleteIntentResponse, error) {
+	requestID := req.GetRequestId()
+	bip322Signature := req.GetBip322Signature()
+
+	if requestID != "" {
```
Kukks, line 165: AH i see it's already in the code as I described.  Just needs the proto to use the oneof style?

Kukks (changes requested):
small change for the proto

Kukks (changes requested):
Can we also please get a test added?

altafan (approved):

server/internal/core/application/service.go
Kukks: yeah

PR #565, opened 2025-05-19
Refactor

altafan:
This PR contains a little internal refactor to facilitate the upcoming internal renaming that follows the announcement of the litepaper.

* walletd - Tidy up code, add changes to fix arkd startup and tier down issues:
   - Removed unnecessary password from `Lock` rpc
   - Add `GetNetwork` rpc to drop arkd env var
   - Replace `GetSyncedUpdate` rpc with `GetReadyUpdate`, this fixes the startup of arkd which was receiving 2 times the synced update that was causing the underlying app svc to be started twice
   - Split data structs of app layer into multiple files
   - Fix interceptors at interface layer
   - Fix dev envs
* arkd - Refactor at domain and app levels that will facilitate the development of upcoming features like the checkpoint system:
   - Source network from wallet
   - Fix regression issue at startup that was causing the app svc to be started twice, generating 2 rounds at a time (introduced when the wallet had been detached)
   - Fix regression issue at tear down that was causing arkd to log errors when closing the connection with the underlying wallet (introduced when the wallet had been detached)
   - Refactor domain:
      * Split repos in dedicated files
      * Add topic to events, register handlers by topic, allow to register multiple handlers (even with same topic)
      * Add `round.VtxoTreeExpiration`, drop `round.DustAmount`, rename `round.UnsignedTx` > `round.CommitmentTx`
      * Make sure to store final signed  commitment tx instead of unsigned only
      * Add OffchainTx + events that represent an offchain spending with support for checkpoint system already
   - Refactor application:
      * Make app svc only write to the event db, the update of the projections is moved into the repo manager service (this allows the app svc to register handlers for watch/schedule/event propagation operations only)
      * Make app svc generate evens at SubmitRedeemTx instead of writing to projection stores
      * Get rid of `custodialVtxos` as it was a not needed categorization of swept or note vtxos
      * Remove boarding input info from round tx event
      * Lock the wallet service at tier down
      * Fix dev envs

Please @louisinger review


server/internal/core/application/service.go
```diff
@@ -172,20 +174,113 @@ func NewService(
...
 			go func() {
 				defer func() {
 					if r := recover(); r != nil {
-						log.Errorf("recovered from panic in updateVtxoSet and scheduleSweepVtxosForRound: %v", r)
+						log.Errorf("recovered from panic in StartWatchingVtxos: %v", r)
+					}
+				}()
+
+				// nolint
+				svc.startWatchingVtxos(newVtxos)
```
louisinger, line 201: nit: this function is parsing vtxo scripts. in case of error, can't we log at least to not have "silent" errors if for some reason the vtxos are invalid ?

server/internal/core/application/service.go
```diff
@@ -212,6 +307,9 @@ func (s *covenantlessService) Stop() {
 		s.stopWatchingVtxos(vtxos)
 	}
 
+	// nolint
+	s.wallet.Lock(context.Background())
```
louisinger:
```suggestion
	ctx := context.Background()
  s.wallet.Lock(ctx)
```
let's reuse the context used in GetAllSweepableVtxos ?

server/internal/core/application/service.go
```diff
@@ -172,20 +174,113 @@ func NewService(
...
+
+			txid, ins, outs, err := decodeTx(offchainTx.VirtualTx)
+			if err != nil {
+				log.WithError(err).Warn("failed to decode virtual tx")
+				return
+			}
+
+			spentVtxoKeys := make([]domain.VtxoKey, 0, len(ins))
+			for _, in := range ins {
+				spentVtxoKeys = append(spentVtxoKeys, domain.VtxoKey(in))
```
louisinger: let's return domain.VtxoKey in `decodeTx` so we can remove this for loop

server/internal/core/application/service.go
```diff
@@ -172,20 +174,113 @@ func NewService(
...
+			}
+			spentVtxos, err := svc.repoManager.Vtxos().GetVtxos(context.Background(), spentVtxoKeys)
+			if err != nil {
+				log.WithError(err).Warn("failed to get spent vtxos")
+				return
+			}
+
+			newVtxos := make([]domain.Vtxo, 0, len(outs))
+			for outIndex, out := range outs {
+				newVtxos = append(newVtxos, domain.Vtxo{
```
louisinger: same, can't we do that job in utils decodeTx function ?

server/internal/core/application/service.go
```diff
@@ -258,13 +379,12 @@ func (s *covenantlessService) SubmitRedeemTx(
...
-	vtxoMap := make(map[wire.OutPoint]domain.Vtxo)
+	indexedSpentVtxos := make(map[domain.VtxoKey]domain.Vtxo)
+	indexedCommitmentTxids := make(map[string]struct{}, 0)
 	for _, vtxo := range spentVtxos {
-		hash, err := chainhash.NewHashFromStr(vtxo.Txid)
-		if err != nil {
-			return "", "", fmt.Errorf("failed to parse vtxo txid: %s", err)
-		}
-		vtxoMap[wire.OutPoint{Hash: *hash, Index: vtxo.VOut}] = vtxo
+		// nolint
```
louisinger:
```suggestion
```

server/internal/core/domain/offchain_tx_repo.go
```diff
@@ -0,0 +1,9 @@
+package domain
+
+import "context"
+
+type OffchainTxRepository interface {
+	AddOrUpdateOffchainTx(ctx context.Context, offchainTx *OffchainTx) error
```
louisinger, line 6: it seems used only in test, is it really needed ?

server/internal/core/domain/tx_request.go
```diff
@@ -91,3 +91,29 @@ type Receiver struct {
 func (r Receiver) IsOnchain() bool {
 	return len(r.OnchainAddress) > 0
 }
+
+type TxRequests []TxRequest
+
+func (t TxRequests) CountSpentVtxos() int {
```
louisinger, line 97: worth commenting why we skip Note and Swept vtxo maybe ?

server/internal/core/ports/tx_decoder.go
```diff
@@ -0,0 +1,14 @@
+package ports
+
+import "github.com/ark-network/ark/server/internal/core/domain"
+
+type TxIn domain.VtxoKey
+
+type TxOut struct {
+	Amount   uint64
+	PkScript []byte
+}
```
louisinger, line 10:
```suggestion
type TxOut wire.TxOut
```

server/internal/core/application/service.go
altafan: doesn't make much sense as we know the scripts are correct, they are extracted from leaf txs of the vtxo tree. We are sure they are correct as they've been validated even client-side

server/internal/core/application/service.go
altafan: done

server/internal/core/application/service.go
altafan: done. Made also `decodeTx` return the new vtxo. The function now takes a `domain.OffchainTx` as arg instead os `string` (only the virtual tx)

server/internal/core/application/service.go
altafan: done

server/internal/core/application/service.go
altafan: done

server/internal/core/domain/offchain_tx_repo.go
altafan: solved by invoking it in repo manager's `updateProjectionsAfterOffchainTxEvents`

server/internal/core/domain/tx_request.go
altafan: done

server/internal/core/ports/tx_decoder.go
altafan: ignored as we loop over these outs to create `domain.Vtxo`, we never translate them to `wire.TxOut`

louisinger (approved):

PR #566, opened 2025-05-20
Add 0-value P2A output to offchain transactions

### LINKED ISSUE #215: Add anchor outputs in TxBuilder
Anchor outputs are necessary to run the clArk service on mainchain.

### LINKED ISSUE #438: Drop custom support for 0-fee offchain tx
Once Bitcoin v29 is released, we can drop the [custom support](https://github.com/ark-network/ark/pull/437) for 0-fee redeem tx and add a 0-value anchor output instead

louisinger:
_This PR adds P2A to vtxo tree transactions and redeem txs. It breaks the "react to fraud" system, waiting for checkpoint (https://github.com/ark-network/ark/issues/491) to be implemented._

### Remove min-relay-fee 

Migrating to transaction version 3 allows to remove the min-relay-fee everywhere in the vtxo tree and ark transaction. It means no more fee estimation while building the tree or an offchain payment.

### Pay to anchor bump

Because every VTXO and batch output is locked by relative CSV locktime, we need a 0-value P2A output in every transactions. 

### Unilateral exit UX

Bitcoin 0.29 allows broadcasting a package with the 1C1P limitation. It implies change in the unilateral exit because `ark redeem --force` has to be called X times, where "X" is the number of transactions to put on chain. 

Bumping the anchor also requires onchain funds, that's why a simple P2TR account has been added to the CLI.

it closes #215
It closes #438

@altafan @Kukks please review


common/tree/builder.go
```diff
@@ -22,15 +22,14 @@ const (
...
 	sweepTapTreeRoot []byte,
 ) ([]byte, int64, error) {
-	root, err := createTxTree(receivers, feeSatsPerNode, sweepTapTreeRoot, vtxoTreeRadix)
+	root, err := createTxTree(receivers, sweepTapTreeRoot, vtxoTreeRadix)
 	if err != nil {
 		return nil, 0, err
 	}
 
-	amount := root.getAmount() + int64(feeSatsPerNode)
+	amount := root.getAmount() + ANCHOR_VALUE
```
altafan, line 32: does it really make sense to keep this?

louisinger: it allows the anchor to be configurable in case we want to revert to anchor with value ?
it's just a constant, so I would keep it but as you prefer

PR #587, opened 2025-05-28
add BatchStarted event and ConfirmRegistration RPC, remove Ping RPC

### LINKED ISSUE #585: Introduce  new round flow
Instead of a constant ping, confirm by sending an event and have intents confirm via API that they will join the batch session. This provides us a non breaking api change path forward for specific features.

louisinger:
This PR removes the Ping RPC, instead the users receive a `BatchStarted` event and must respond with `ConfirmRegistration` RPC.

When the server starts the round, it selects a set of intents and sends the corresponding `BatchStarted` event. If all users confirm their intent, the server process with the round's finalization. If some intents are not confirmed, the server tries to select new ones and resend `BatchStarted` event. It tries 3 times this confirmation process before making the round to fail.

it closes #585 

@Kukks @altafan please review


api-spec/protobuf/ark/v1/service.proto
```diff
@@ -145,6 +146,11 @@ message RegisterInputsForNextRoundResponse {
   string request_id = 1;
 }
 
+message ConfirmRegistrationRequest {
+  string intent_id_hash = 1;
```
altafan: the client should reveal the intent to prove its identity?
```suggestion
  string intent_id = 1;
```

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -145,6 +146,11 @@ message RegisterInputsForNextRoundResponse {
...
+}
+message ConfirmRegistrationResponse {}
```
altafan: It will be empty for now, but let's return the credentials here
```suggestion
message ConfirmRegistrationResponse {
   string blinded_creds = 1
}
```

server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -218,6 +218,26 @@ func (h *handler) RegisterInputsForNextRound(
...
+) (*arkv1.ConfirmRegistrationResponse, error) {
+	intentIdHash := req.GetIntentIdHash()
+	if len(intentIdHash) <= 0 {
+		return nil, status.Error(codes.InvalidArgument, "missing intent id hash")
+	}
+
+	intentIdHashArray, err := parseIntentIdHash(intentIdHash)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
```
altafan: The first check can be moved into the parser function

```suggestion
	intentIds, err := parseIntentIds(req.GetIntentIds())
	if err != nil {
		return nil, status.Error(codes.InvalidArgument, err.Error())
	}
```

server/internal/core/application/service.go
```diff
@@ -2293,6 +2321,116 @@ func (s *covenantlessService) scheduleSweepVtxosForRound(round *domain.Round) {
...
+	}
+
+	if !bytes.Equal(output.PkScript, expectedScriptPubkey) {
+		return nil, fmt.Errorf(
+			"invalid boarding utxo taproot key: got %x expected %x",
+			output.PkScript, expectedScriptPubkey,
+		)
+	}
+
+	if err := boardingScript.Validate(s.pubkey, s.unilateralExitDelay); err != nil {
```
altafan: shouldn't be boarding exit delay?

server/internal/core/application/service.go
```diff
@@ -2293,6 +2321,116 @@ func (s *covenantlessService) scheduleSweepVtxosForRound(round *domain.Round) {
 	}
 }
 
+func (s *covenantlessService) newBoardingInput(tx wire.MsgTx, input ports.Input) (*ports.BoardingInput, error) {
```
altafan: this function doesn't really need to be defined for the service. It can just expect 2 more args `serverPubkey` and `boardingExitDelay` and it can be moved to utils.go

server/internal/core/application/service.go
```diff
@@ -2532,6 +2670,46 @@ func newMusigSigningSession(cosigners map[string]struct{}) *musigSigningSession
...
+
+	s.intentsHashes[hash] = true
+
+	for _, confirmed := range s.intentsHashes {
+		if !confirmed {
+			return nil
+		}
+	}
+
+	s.confirmedC <- struct{}{}
```
altafan: this can be moved into a goroutine and be executed in the background?

server/internal/core/application/service.go
```diff
@@ -1850,9 +1774,93 @@ func (s *covenantlessService) startFinalization(roundEndTime time.Time) {
...
+		case <-confirmationTicker.C:
+			numberOfConfirmationAttemps--
+			if numberOfConfirmationAttemps == 0 {
+				// no more attempts, abort the round
+				confirmationTicker.Stop()
+				err := fmt.Errorf("confirmation timed out")
+				round.Fail(err)
+				log.WithError(err).Warn("confirmation timed out")
+				return
+			}
```
altafan: here we want to make the round fail only if nobody confirmed (change the failing reason to something more meaningful like `0 confirmations received`).
In case some user confirmed, we should instead proceed if they are above the min required threshold (let's use a var configurable via env that can default to `1` for now).

pkg/client-sdk/client.go
```diff
@@ -1935,19 +1935,14 @@ func (a *covenantlessArkClient) joinRoundWithRetry(
...
 	signerSessions []tree.SignerSession,
 	replayEventsCh chan<- client.RoundEvent,
 ) (string, error) {
-	round, err := a.client.GetRound(ctx, "")
-	if err != nil {
-		return "", err
-	}
-
-	eventsCh, close, err := a.client.GetEventStream(ctx, requestID)
+	eventsCh, close, err := a.client.GetEventStream(ctx, intentID)
```
altafan: the intent id was needed in the past to spare a ping request. We can get rid of it form the GetEventRequest message.

pkg/client-sdk/client.go
```diff
@@ -1990,40 +1976,55 @@ func (a *covenantlessArkClient) handleRoundStream(
...
 			if notify.Err != nil {
 				return "", notify.Err
 			}
+
 			if replayEventsCh != nil {
 				go func() {
 					replayEventsCh <- notify.Event
 				}()
 			}
+			fmt.Printf("notify %+v\n", notify.Event)
```
altafan: remove

altafan (approved):

PR #597, opened 2025-06-04
Implement  Websocket Functionality For ClientSDK

aruokhai:
## Objective
Leverage Mempool.space’s WebSocket API to monitor block‐inclusion events for boarding addresses for ClientSDK. Implement real‐time event handling and ensure a reliable fallback mechanism for error conditions or when running on Regtest.


pkg/client-sdk/client.go
```diff
@@ -1005,6 +1007,68 @@ func (a *covenantlessArkClient) refreshVtxoDb(spendableVtxos, spentVtxos []clien
...
+	_, boardingAddresses, _, err := a.wallet.GetAddresses(ctx)
+	if err == nil {
+		for _, boardingAddress := range boardingAddresses {
+			if err := a.explorer.TrackAddress(boardingAddress.Address); err != nil {
+				log.WithError(err).Errorf("failed to track boarding address %s", boardingAddress.Address)
+			}
+		}
+	}
+
+	err = a.explorer.ListenAddresses(func(utxos []explorer.BlockUtxo) error {
```
altafan: we need logic to either add or confirm txs here.. and also we should make use of this https://mempool.space/docs/api/websocket#track-full-rbf to track rbf txs (happens a lot on signet)

aruokhai: Unfortunately, the endpoint wouldn't be effective due to it being a general endpoint, which therefore activates based on any RBF transaction. Ultimately, we can still rely of mempool address event, because Rbf also activates a new mempool event.

pkg/client-sdk/client.go
```diff
@@ -205,6 +206,9 @@ func LoadArkClient(sdkStore types.Store) (ArkClient, error) {
 		},
 	}
 
+	// subscribe to boarding address events
+	explorerSvc.SubscribeToAddressEvent(walletSvc.SubscribeAddressEvent(context.TODO()))
+
```
altafan: this should be moved within the `if cfgData.WithTransactionFeed` block?
for the cli, that is a use case of the sdk **without tx feed**, it's useless to subscribe for addresses event as it won't ever use the subscription.

PR #599, opened 2025-06-04
Subscription API

### LINKED ISSUE #582: Add Subscribe Stream API
Similar to the existing SubscribeToAddress API, which will be removed anyway, we need a more scalable and optimized option.

It:
* need to allow listening to multiple addresses
* provides the virtual transactions in the even along with effected vtxos
* have a way to update the subscription without killing it

louisinger:
This PR adds subscription APIs to indexed

1. `SubscribeForScripts`: create a subscription or add new scripts to watch. Returns subscription ID.
2. `GetSubscribtion`:  returns the server side stream associated with a subscription ID.
3. `UnsubscribeForScripts`: remove watched addresses from a subscription, delete the subscription if no more addresses are watched.

it closes #582 

@altafan @Kukks please review


Kukks (approved):
proto ACK

altafan (approved):

PR #603, opened 2025-06-10
Service live store

sekulicd:
This moves away data structs from app service and introduce LiveStore port interface to do cache crud operation. 
LiveStore is implemented as inmemory and redis.

@altafan please review.


docker-compose.regtest.yml
```diff
@@ -1,4 +1,22 @@
 services:
```
altafan: please revert the changes, but let's not forget to make use of inmemory live store, badger event repo and sqlite repo

server/envs/dev.env
```diff
@@ -4,6 +4,9 @@ ARK_VTXO_TREE_EXPIRY=604672
 ARK_UNILATERAL_EXIT_DELAY=1024
 ARK_BOARDING_EXIT_DELAY=604672
 ARK_DATADIR=./data/regtest
+ARK_ESPLORA_URL=http://localhost:3000
 ARK_WALLET_ADDR=localhost:6060
+ARK_DB_URL=postgresql://root:secret@127.0.0.1:5432/ark-db?sslmode=disable
+ARK_LIVE_STORE_TYPE=inmemory
```
altafan: must be redis

docker-compose.regtest.yml
sekulicd: Done

server/envs/dev.env
sekulicd: Done

PR #618, opened 2025-06-16
Verify exit closures & Verify CSV in RegisterIntent

altafan:
This adds verification of the exit closures when spending a VTXO and also adds loigc to RegisterIntent to validate the CSV delay of the exit path used  to prove ownership of the funds.

Please @louisinger @Kukks review.

### DIFF
--- a/server/internal/core/application/service.go
+++ b/server/internal/core/application/service.go
@@ -37,6 +37,7 @@ type covenantlessService struct {
 	roundInterval       time.Duration
 	unilateralExitDelay common.RelativeLocktime
 	boardingExitDelay   common.RelativeLocktime
+	allowCSVBlockType   bool
 
 	wallet      ports.WalletService
 	repoManager ports.RepoManager
@@ -92,6 +93,7 @@ func NewService(
 	utxoMinAmount int64,
 	vtxoMaxAmount int64,
 	vtxoMinAmount int64,
+	allowCSVBlockType bool,
 ) (Service, error) {
 	pubkey, err := walletSvc.GetPubkey(context.Background())
 	if err != nil {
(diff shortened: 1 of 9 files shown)


server/internal/core/application/service.go
```diff
@@ -835,29 +850,29 @@ func (s *covenantlessService) RegisterIntent(ctx context.Context, bip322signatur
...
 				// if the exit path is available, forbid registering the boarding utxo
 				if time.Unix(blocktime, 0).Add(time.Duration(exitDelay.Seconds()) * time.Second).Before(time.Now()) {
 					return "", fmt.Errorf("tx %s expired", vtxoKey.Txid)
 				}
 
+				// If the intent is registered using a exit path that contains CSV delay, we want to verify it
+				// by shifitng the current "now" in the future of the duration of the smallest exit delay.
+				// This way, any exit order guaranteed by the exit path is maintained at intent registration
+				if !disabled {
+					delta := time.Now().Add(time.Duration(exitDelay.Seconds())*time.Second).Unix() - blocktime
```
Kukks: can we take a snapshot of time.Now in the beginning so that no weird conditions happen midway through

server/internal/core/application/service.go
```diff
@@ -916,17 +931,22 @@ func (s *covenantlessService) RegisterIntent(ctx context.Context, bip322signatur
 			PkScript: pkScript,
 		}
 
+		// If the intent is registered using a exit path that contains CSV delay, we want to verify it
+		// by shifitng the current "now" in the future of the duration of the smallest exit delay.
+		// This way, any exit order guaranteed by the exit path is maintained at intent registration
+		if !disabled {
```
Kukks: duplicate code as above?

Kukks (approved):
small nit.

server/internal/core/application/service.go
altafan: yeah it's (almost) duplicated because the first is inside of an if that `continue`s at its end. The second instead is outside the if so they're mutually exclusive... didn't want to refactor the code because of this. I'll eventually take care of this in the final renaming

server/internal/core/application/service.go
altafan: done

louisinger (approved):

louisinger (approved):

Issue #589, opened 2025-05-29
Client should use `BatchTreeEvent.topic` to select only the branches to sign instead of the whole tree


louisinger:
Instead of processing the whole tx trees, client should only care about the branches they are signing.

louisinger:
need https://github.com/arkade-os/arkd/pull/617

PR #627, opened 2025-06-18
Remove matrix, replace with `TxGraph` 

### LINKED ISSUE #607: Convert Tree structure to graph instead of matrix
This anticipates the new partial tree  broadcast system where a matrix wont allow tree construction.  The Levelindex becomes bound to its own node branch 

Exmaple:

 node 8e000e9479a0841d98e2981e6bf2e34b58f8c1793d419dbcd06da0c13f1851ae has 2 children: 

c018e8da997cd0f4df6dfc7650efdd3be42b435ba9f0de9d8f835d759cda9f54
98b450a298c9307f89f844aeca8aaffccccd7841a74d67c6a059521c1564a554

The levelIndex would be 0 and 1 regardless of an earlier node on a similar depth

louisinger:
_This PR replaces the matrix data structure by maps (for musig2 nonces and sigs) and TxGraph (for TxTree)_

### TxGraph

TxTree struct has been removed and replaced by a `TxGraph` data structure. Graph allows flexible manipulation of batch tree in order to prepare https://github.com/arkade-os/arkd/issues/589.

A `TxGraph` is serializable as a list of `TxGraphChunk`, letting to remove the need for index and levelIndex in APIs. The psbt itself is used to recompute the position of the transaction.

`TxGraph` is composed of `psbt.Packet` instead of serialized base64 strings, allowing to remove a lot of serialization roundtrip in the code.

### Musig2 nonces and signatures

This PR also migrates musig2 `tree.TreeNonces` and `tree.TreeSignatures` from matrixes to maps. Now, the nonces and sigs are sent mapped to their transaction's id.

**This PR is not ready to be merged, it misses the Postgres migrations files.  waiting for first reviews on sqlite schema to update Postgres.**

it closes #607 

@altafan @Kukks @sekulicd please review


server/internal/core/domain/round_event.go
```diff
@@ -21,8 +21,8 @@ type RoundStarted struct {
 
 type RoundFinalizationStarted struct {
 	RoundEvent
-	VtxoTree           tree.TxTree
-	Connectors         tree.TxTree
+	VtxoTree           []tree.TxGraphChunk
+	Connectors         []tree.TxGraphChunk
```
altafan, line 25: why not `tree.TxGraph`?

common/tree/tx_graph.go
```diff
@@ -0,0 +1,330 @@
...
+
+// TxGraphChunk is a chunk of TxGraph
+// it is used to serialize and deserialize the graph because TxGraph is recursive
+// a list of TxGraphChunk can be used to reconstruct the TxGraph
+type TxGraphChunk struct {
+	// Tx is the base64 encoded root PSBT
+	Tx string
+	// Children maps root output index to child txid
+	Children map[uint32]string
+}
```
altafan: isn't it better to have also the Txid here as field so we can get rid of `TxGraphChunk.TxID()` and spare psbt deserialization?

louisinger: yes in fact that's planned 👍

server/internal/core/domain/round_event.go
louisinger: to keep the domain as close as possible to DB schema. moreover TxGraph is a recursive struct and should be build only when really needed (most of the operations on Round domain do not require TxGraph)

common/tree/tx_graph.go
louisinger: done

common/tree/builder.go
```diff
@@ -90,88 +89,75 @@ func CraftConnectorsOutput(
...
-
-			nodeTxHash, err := chainhash.NewHashFromStr(treeNode.Txid)
-			if err != nil {
-				return nil, err
-			}
-
-			treeLevel = append(treeLevel, treeNode)
-
-			children := node.getChildren()
+func nodeToGraph(n node, initialInput *wire.OutPoint, expiry *common.RelativeLocktime) (*TxGraph, error) {
```
altafan: let's make this a method of node

common/tree/builder.go
```diff
@@ -90,88 +89,75 @@ func CraftConnectorsOutput(
...
 }
 
 type leaf struct {
-	amount    int64
-	pkScript  []byte
-	cosigners []*secp256k1.PublicKey
+	amount        int64
+	pkScript      []byte
+	inputPkScript []byte
+	cosigners     []*secp256k1.PublicKey
```
altafan:
```suggestion
	output *wire.TxOut
	inputScript []byte
	cosigners     []*secp256k1.PublicKey
```

common/tree/musig2.go
```diff
@@ -30,58 +28,87 @@ type Musig2Nonce struct {
...
-func (n *Musig2Nonce) Encode(w io.Writer) error {
-	_, err := w.Write(n.PubNonce[:])
-	return err
+type TreeNonces map[string]*Musig2Nonce                  // txid -> public nonces only
+type TreePartialSigs map[string]*musig2.PartialSignature // txid -> partial signature
+
+var _ json.Marshaler = TreeNonces{}
+var _ json.Unmarshaler = (*TreeNonces)(nil)
+var _ json.Marshaler = TreePartialSigs{}
+var _ json.Unmarshaler = (*TreePartialSigs)(nil)
```
altafan: drop

common/tree/musig2.go
```diff
@@ -277,46 +283,53 @@ func (t *treeSignerSession) SetAggregatedNonces(nonces TreeNonces) {
...
 		return nil, errors.New("nonces not set")
 	}
 
-	sigs := make(TreePartialSigs, 0, len(t.txs))
-	for i := range t.txs {
-		sigs = append(sigs, make([]*musig2.PartialSignature, len(t.txs[i])))
-	}
-
-	signerPubKey := schnorr.SerializePubKey(t.secretKey.PubKey())
+	musignParamsMap := make(map[string]musignParams)
```
altafan: typo

```suggestion
	signingParamsMap := make(map[string]musignParams)
```

common/tree/tx_graph.go
```diff
@@ -0,0 +1,324 @@
...
+
+	// Create a map to store all chunks by their txid for easy lookup
+	chunksByTxid := make(map[string]decodedTxGraphChunk)
+
+	for _, chunk := range chunks {
+		packet, err := psbt.NewFromRawBytes(strings.NewReader(chunk.Tx), true)
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode PSBT: %w", err)
+		}
+		txid := packet.UnsignedTx.TxID()
```
altafan, line 54: shall we make sure this txid matches the one of the chunk?

common/tree/tx_graph.go
```diff
@@ -0,0 +1,324 @@
...
+		childChunks, err := child.Serialize()
+		if err != nil {
+			return nil, err
+		}
+		chunks = append(chunks, childChunks...)
+	}
+
+	// serialize the current node's root
+	var serializedTx bytes.Buffer
+	if err := g.Root.Serialize(&serializedTx); err != nil {
```
altafan: why don't we encode directly in b64 with the method of the packet?

pkg/client-sdk/client/grpc/types.go
```diff
@@ -99,19 +98,13 @@ func (e event) toRoundEvent() (client.RoundEvent, error) {
...
 			Topic:      ee.GetTopic(),
 			BatchIndex: ee.GetBatchIndex(),
-			Node: tree.Node{
-				Txid:       treeTx.GetTxid(),
-				Tx:         treeTx.GetTx(),
-				ParentTxid: treeTx.GetParentTxid(),
-				Level:      treeTx.GetLevel(),
-				LevelIndex: treeTx.GetLevelIndex(),
-				Leaf:       treeTx.GetLeaf(),
+			TxGraphChunk: tree.TxGraphChunk{
```
altafan, line 88: missing Txid?

pkg/client-sdk/client/rest/client.go
```diff
@@ -260,18 +255,15 @@ func (a *restClient) SubmitTreeNonces(
 	ctx context.Context, roundID, cosignerPubkey string,
 	nonces tree.TreeNonces,
 ) error {
-	var nonceBuffer bytes.Buffer
-
-	if err := nonces.Encode(&nonceBuffer); err != nil {
+	sigsJSON, err := nonces.MarshalJSON()
```
altafan: the name of the var is wrong, and also, in other parts of the codebase we use json.Marshal, can we keep using only that instead of mixing it with this kind of invocation?

server/internal/core/ports/live_store.go
```diff
@@ -33,7 +33,7 @@ type TxRequestsStore interface {
 }
 
 type ForfeitTxsStore interface {
-	Init(connectors tree.TxTree, requests []domain.TxRequest) error
+	Init(connectors *tree.TxGraph, requests []domain.TxRequest) error
```
altafan: maybe better to store the []TxGraphChunk version? easier to be memorized

server/internal/infrastructure/db/service_test.go
```diff
@@ -34,81 +35,109 @@ const (
 	virtualTxid = txida
 )
 
-var (
-	vtxoTree = [][]tree.Node{
-		{
+func randomTx() string {
```
altafan: please move this at the very end of the file

server/internal/infrastructure/db/service_test.go
```diff
@@ -34,81 +35,109 @@ const (
...
+	b64, _ := ptx.B64Encode()
+	return b64
+}
+
+var (
+	vtxoTree = []tree.TxGraphChunk{
+		{
+			Txid:     randomString(32),
+			Tx:       randomTx(),
+			Children: map[uint32]string{},
```
altafan, line 44: if this is the leaf tx, we can just skip defining Children

server/internal/infrastructure/db/service_test.go
```diff
@@ -130,13 +159,13 @@ var (
 
 func TestMain(m *testing.M) {
 	m.Run()
-	_ = os.Remove("test.db")
+	// _ = os.Remove("test.db")
```
altafan: just a reminder to uncoment these lines when pg is ok

server/internal/infrastructure/db/sqlite/migration/20250618123533_tx_graph.up.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE tx DROP COLUMN tree_level;
+ALTER TABLE tx DROP COLUMN parent_txid;
+ALTER TABLE tx DROP COLUMN is_leaf;
+
+ALTER TABLE tx ADD COLUMN children TEXT;
```
altafan, line 5: @sekulicd this children is a map[uint32]string that (at the moment) can contain max 2 entries. Here we're storing its json serialization in sqlite. Do you think it's ok? Do you suggest something else?

common/tree/tx_graph.go
louisinger: the chunk's Txid is not used to create the graph, so not sure if it's really needed

common/tree/tx_graph.go
```diff
@@ -0,0 +1,335 @@
...
+	for _, txid := range txids {
+		txidSet[txid] = true
+	}
+
+	return g.buildSubGraph(txidSet)
+}
+
+// crawl recursively crawls the graph to find the subgraph starting from root to the given txids
+// it returns the list of txs from the root to the txid and the tx graph where the txid is found
+func (g *TxGraph) crawl(txid string, branch []*psbt.Packet) ([]*psbt.Packet, *TxGraph, error) {
```
altafan: rename `find` and drop `branch` arg as it's unused

common/tree/tx_graph.go
```diff
@@ -0,0 +1,335 @@
...
+	leaves := make([]*psbt.Packet, 0)
+
+	for _, child := range g.Children {
+		leaves = append(leaves, child.Leaves()...)
+	}
+
+	return leaves
+}
+
+// Find returns the tx graph where root is equal to the given txid
```
altafan:
```suggestion
// Find returns the tx in the graph that matches the provided txid
```

common/tree/tx_graph.go
```diff
@@ -0,0 +1,321 @@
...
+		return fmt.Errorf("unexpected version: %d, expected 3", g.Root.UnsignedTx.Version)
+	}
+
+	nbOfOutputs := uint32(len(g.Root.UnsignedTx.TxOut))
+	nbOfInputs := uint32(len(g.Root.UnsignedTx.TxIn))
+
+	if nbOfInputs != 1 {
+		return fmt.Errorf("unexpected number of inputs: %d, expected 1", nbOfInputs)
+	}
+
```
altafan, line 153: let's add a check to ensure Children has the same len of `nbOfOutputs`? The check below is not enough to ensure this condition?

server/internal/infrastructure/db/service_test.go
```diff
@@ -367,7 +369,7 @@ func testEventRepository(t *testing.T, svc ports.RepoManager) {
 		}
 		ctx := context.Background()
 
-		for _, f := range fixtures {
+		for _, f := range fixtures[:1] {
```
altafan: why?

PR #629, opened 2025-06-19
LiveStore/inmemory: add extra checks before closing channels

louisinger:
@altafan please review

### DIFF
--- a/server/internal/infrastructure/live-store/inmemory/tree_signing_session.go
+++ b/server/internal/infrastructure/live-store/inmemory/tree_signing_session.go
@@ -52,8 +52,14 @@ func (s *treeSigningSessionsStore) Get(roundId string) (*ports.MusigSigningSessi
 func (s *treeSigningSessionsStore) Delete(roundId string) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	close(s.nonceCollectedCh[roundId])
-	close(s.sigsCollectedCh[roundId])
+
+	if ch, exists := s.nonceCollectedCh[roundId]; exists {
+		close(ch)
+	}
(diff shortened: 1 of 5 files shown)


server/internal/infrastructure/live-store/inmemory/tree_signing_session.go
```diff
@@ -76,7 +82,12 @@ func (s *treeSigningSessionsStore) AddNonces(
 	s.sessions[roundId].Nonces[pubkey] = nonces
 
 	if len(s.sessions[roundId].Nonces) == s.sessions[roundId].NbCosigners-1 {
-		s.nonceCollectedCh[roundId] <- struct{}{}
+		if ch, exists := s.nonceCollectedCh[roundId]; exists {
```
altafan: let's return error if the channel doesn't exist because it means something bad happened! we don't want to ignore such situation if we have this kind of check. The existing code assumed something bad can't just happen instead

server/internal/infrastructure/live-store/inmemory/tree_signing_session.go
```diff
@@ -99,16 +110,31 @@ func (s *treeSigningSessionsStore) AddSignatures(
...
+			}
+		}
 	}
 
 	return nil
 }
 
 func (s *treeSigningSessionsStore) NoncesCollected(roundId string) <-chan struct{} {
-	return s.nonceCollectedCh[roundId]
+	s.lock.RLock()
```
altafan: please revert the changes to NoncesCollected and SignaturesCollected. As per impl, the functions are meant to be called only once. Even with the lock, if this function is called twice, the 2 invocation will compete to receive a msg and we simply don't want this, neither expect it in the future.

PR #632, opened 2025-06-23
Move to vtxo output script & Change address encoding

### LINKED ISSUE #623: Change ark address data
* switch the server to the second half
* switch the user to a taproot payment script

### LINKED ISSUE #624: Use payment script instead of ark address
inside the API, only use payment scripts instead of ark addresses.

altafan:
This makes the server return and accept vtxo output scripts instead of pubkeys.

This also changes the address encoding format from <hrp serverKey vtxoKey> to <hrp vtxoScript serverKey>.

Closes #623
Closes #624 

Please @louisinger @Kukks review

### DIFF
--- a/common/encoding.go
+++ b/common/encoding.go
@@ -11,8 +11,8 @@ import (
 // Address represents an Ark address with HRP, server public key, and VTXO Taproot public key
 type Address struct {
 	HRP        string
+	VtxoScript []byte
 	Server     *secp256k1.PublicKey
-	VtxoTapKey *secp256k1.PublicKey
 }
 
 // Encode converts the address to its bech32m string representation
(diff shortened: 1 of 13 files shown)


louisinger:
I would keep tapkey in the address encoding to make it "smaller". Also to make it closer to what's done in bitcoin (taproot address = bech32m prefix + tapkey)

common/encoding.go
```diff
@@ -34,35 +35,37 @@ func (a *Address) Encode() (string, error) {
...
-		return nil, fmt.Errorf("invalid prefix")
+		return nil, fmt.Errorf("unknown prefix")
 	}
 	grp, err := bech32.ConvertBits(buf, 5, 8, false)
 	if err != nil {
 		return nil, err
 	}
 
-	serverKey, err := schnorr.ParsePubKey(grp[:32])
+	version := uint32(grp[0])
```
louisinger: Shouldn't we fail if version != 0 ?

common/encoding.go
```diff
@@ -34,35 +35,37 @@ func (a *Address) Encode() (string, error) {
...
+	serverKey, err := schnorr.ParsePubKey(grp[1:33])
```
louisinger, line 66: I'm realizing that maybe it's not safe to access `grp` without checking its length first ? it may panic if someone pass a smaller string ?

louisinger (approved):

PR #633, opened 2025-06-24
Batch "GetEventStream" pub/sub API

### LINKED ISSUE #570: Do not hardcode forfeit tx input expected indexes
We should scan a forfeit tx for the vtxo /connector instead of expecting them in a specific place.

### LINKED ISSUE #586: Tree Session with VTXO pubsub
Instead of sending the entire tree of a batch session to everyone, create a pubsub system where every relevant virtual transaction of a new vtxo is published with the vtxo as the topic.

### LINKED ISSUE #589: Client should use `BatchTreeEvent.topic` to select only the branches to sign instead of the whole tree

louisinger:
_This PR reworks the `GetEventStream` RPC as well as the way signers receive graph-related events._

### GetEventStream Topic

GetEventStream accepts two types of topic:
* VTXO graph topics = the list of **musig2 public keys** involved in the input's signature.
* Connector graph topics = the list of **spent VTXO outpoint** related to the graph's leaves, a.k.a. connectors outputs.

Thus, if Alice is spending vtxo `txid:vout` and sign the tree with public key `pubkey_a`. she has to subscribe to `["txid:vout", "pubkey_a"]` topics.

It allows Alice to only get the portion of the vtxo tree in which she is involved (all nodes she has to musig2 sign) and only get the portion of the connector tree related to its spent VTXOs (the one she needs to craft the forfeit tx). 

Topics are only used to publish `TreeTxEvent` and `TreeSignatureEvent`. All other events are sent to every users.

### Connectors Index

Because the pub/sub API allows receiving only the connector you need, `BatchFinalizationEvent` doesn't need any more to return the `connectorsIndex` map.

However, we still need to keep the connectorIndex in `LiveStore` in order to compute the connectors tree topics while publishing events. For optimization purpose, this connector index map has been "inverted" (from `vtxo -> connector` to `connector -> vtxo`. 

it closes #586 
it closes #589
it closes #570 

@altafan @sekulicd @Kukks please review


common/tree/forfeit.go
```diff
@@ -8,16 +8,16 @@ import (
 
 func BuildForfeitTx(
 	vtxoInput, connectorInput *wire.OutPoint,
-	vtxoAmount, connectorAmount uint64,
-	vtxoScript, connectorScript, serverScript []byte,
+	vtxoPrevout, connectorPrevout *wire.TxOut,
+	serverScript []byte,
 	txLocktime uint32,
+	vtxoFirst bool,
```
altafan: why this?

pkg/client-sdk/client.go
```diff
@@ -2115,7 +2123,7 @@ func (a *covenantlessArkClient) handleBatchEvents(
 				}
 				vtxoGraph, err = tree.NewTxGraph(vtxoGraphChunks)
 				if err != nil {
-					return "", err
+					return "", fmt.Errorf("failed to create vtxo graph: %s", err)
```
altafan: what about? 
```suggestion
					return "", fmt.Errorf("failed to create branch of vtxo tree: %s", err)
```

pkg/client-sdk/client.go
```diff
@@ -2157,7 +2165,7 @@ func (a *covenantlessArkClient) handleBatchEvents(
 				if len(connectorsGraphChunks) > 0 {
 					connectorsGraph, err = tree.NewTxGraph(connectorsGraphChunks)
 					if err != nil {
-						return "", err
+						return "", fmt.Errorf("failed to create connectors graph: %s", err)
```
altafan: and?
```suggestion
						return "", fmt.Errorf("failed to create branch of connector tree: %s", err)
```

pkg/client-sdk/client.go
```diff
@@ -2481,14 +2488,9 @@ func (a *covenantlessArkClient) validateVtxoTree(
...
-			return fmt.Errorf("empty connectors index")
-		}
-
-		for _, vtxo := range vtxosInput {
-			if _, ok := event.ConnectorsIndex[vtxo.String()]; !ok {
-				return fmt.Errorf("missing connector index for vtxo %s", vtxo.String())
-			}
+		connectorsLeaves := connectorsGraph.Leaves()
+		if len(connectorsLeaves) != len(vtxosInput) {
+			return fmt.Errorf("number of connectors leaves (%d) does not match number of vtxo inputs (%d)", len(connectorsLeaves), len(vtxosInput))
```
altafan:
```suggestion
			return fmt.Errorf("not enough connectors received: expected %d, got %d", len(vtxosInput), len(connectorsLeaves))
```

pkg/client-sdk/client/rest/client.go
```diff
@@ -224,11 +225,20 @@ func (a *restClient) SubmitSignedForfeitTxs(
...
 	eventsCh := make(chan client.BatchEventChannel)
 	chunkCh := make(chan chunk)
+
+	// Construct the URL with proper multi-format topics parameter
 	url := fmt.Sprintf("%s/v1/batch/events", c.serverURL)
+	if len(topics) > 0 {
+		queryParams := make([]string, 0, len(topics))
+		for _, topic := range topics {
+			queryParams = append(queryParams, fmt.Sprintf("topics=%s", topic))
+		}
```
altafan, line 239: maybe i'm wrong but i think we can pass many topics like `?topics=t1,t2,t3...`

pkg/client-sdk/client.go
louisinger: "not enough" can be misleading cause it must fail also if you received more connectors than expected (it means something went wrong server side, in theory should never happen)

pkg/client-sdk/client/rest/client.go
louisinger: I tried, it seems it's not working

louisinger: maybe need proper swagger configuration in proto?

common/tree/forfeit.go
louisinger: to solve this : https://github.com/arkade-os/arkd/issues/570

pkg/client-sdk/client/rest/client.go
Kukks: usually array in query string is `x=1&x=2&x=3...`

common/tree/forfeit.go
louisinger: but I'll remove

pkg/client-sdk/client/rest/client.go
louisinger: So maybe we should name the field "topic" (not topics)

common/tree/forfeit.go
altafan: what if we pass the inputs already as list of outpoints so we delegate choosing their order directly to the func caller?

pkg/client-sdk/client.go
altafan: ok change to `unexpected num of connectors received: expected %d, got %d`

server/envs/light.env
```diff
@@ -0,0 +1,13 @@
+ARK_LOG_LEVEL=5
+ARK_NO_MACAROONS=true
+ARK_VTXO_TREE_EXPIRY=604672
+ARK_UNILATERAL_EXIT_DELAY=1024
+ARK_BOARDING_EXIT_DELAY=604672
```
altafan: set these to like 15 mins, 8 mins and 15 mins respectively. Please do the same in the other env since we're here

server/envs/light.env
```diff
@@ -0,0 +1,13 @@
...
+ARK_DATADIR=./data/regtest
+ARK_ESPLORA_URL=http://localhost:3000
+ARK_WALLET_ADDR=localhost:6060
+ARK_LIVE_STORE_TYPE=inmemory
+ARK_DB_TYPE=sqlite
+ARK_EVENT_DB_TYPE=badger
+ARK_REDIS_URL=redis://localhost:6379/0
```
altafan, line 12: drop

server/internal/core/domain/vtxo.go
```diff
@@ -14,7 +14,18 @@ type VtxoKey struct {
 	Txid string
 	VOut uint32
 }
-type Outpoint VtxoKey
+
+// implements ports.TxOutpoint interface
+func (k VtxoKey) GetTxid() string {
```
altafan: rename this Outpoint and drop the alias

server/internal/infrastructure/tx-builder/covenantless/builder.go
```diff
@@ -329,28 +328,59 @@ func (b *txBuilder) VerifyForfeitTxs(
...
+		for i, input := range tx.UnsignedTx.TxIn {
+			for _, connector := range connectorsLeaves {
+				if connector.Txid == input.PreviousOutPoint.Hash.String() {
+					connectorTx, err := psbt.NewFromRawBytes(strings.NewReader(connector.Tx), true)
+					if err != nil {
+						return nil, err
+					}
+
+					if len(connectorTx.UnsignedTx.TxOut) <= int(input.PreviousOutPoint.Index) {
+						continue
```
altafan: shall we return an error?
```suggestion
						return nil, fmt.Errorf("connector vout %d out of range [0, %d]", input.PreviousOutPoint.Index), len(connectorTx.UnsignedTx.TxOut) -1)
```

server/internal/interface/grpc/handlers/broker.go
```diff
@@ -13,6 +14,33 @@ type listener[T any] struct {
...
+	}
+	return &listener[T]{
+		id:            id,
+		topics:        topicsMap,
+		ch:            make(chan T, 1000),
+		stopTimeoutCh: make(chan struct{}),
+	}
+}
+
+func (l *listener[T]) hasTopic(topics []string) bool {
```
altafan:
```suggestion
func (l *listener[T]) includesAny(topics []string) bool {
```

server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -505,3 +546,12 @@ func (h *handler) listenToTxEvents() {
 		}
 	}
 }
+
+type eventWithTopics struct {
+	topics []string
+	event  *arkv1.GetEventStreamResponse
+}
+
+func (e *eventWithTopics) GetTopics() []string {
```
altafan: drop?

server/internal/core/application/service.go
```diff
@@ -2642,3 +2670,46 @@ func batchTreeSignatureEvents(graph *tree.TxGraph, batchIndex int32, roundId str
 
 	return events
 }
+
+// getVtxoTreeTopic returns the list of topics for vtxo graph
+// it returns the list of cosigner public keys used to compute the musig2 key
```
altafan:
```suggestion
// getVtxoTreeTopic returns the list of topics (cosigner keys) for the given vtxo subtree
```

server/internal/core/application/service.go
```diff
@@ -2642,3 +2670,46 @@ func batchTreeSignatureEvents(graph *tree.TxGraph, batchIndex int32, roundId str
...
+	topics := make([]string, 0, len(cosignerKeys))
+	for _, key := range cosignerKeys {
+		topics = append(topics, hex.EncodeToString(key.SerializeCompressed()))
+	}
+
+	return topics, nil
+}
+
+// getConnectorTreeTopic returns the list of topics for connector graph
+// it returns the list of vtxo outpoints associated with the connector
```
altafan:
```suggestion
// getConnectorTreeTopic returns the list of topics (vtxo outpoints) for the given connector subtree
```

server/internal/interface/grpc/handlers/arkservice.go
```diff
@@ -465,9 +501,14 @@ func (h *handler) listenToEvents() {
 			logrus.Debugf("forwarding event to %d listeners", len(h.eventsListenerHandler.listeners))
 			for _, l := range h.eventsListenerHandler.listeners {
 				go func(l *listener[*arkv1.GetEventStreamResponse]) {
+					sentToListeners := 0
```
altafan:
```suggestion
					sentEventsCount := 0
```

altafan (approved):

PR #634, opened 2025-06-24
Vtxo commitment txids

### LINKED ISSUE #615: List all commitment txs in a vtxo
List all root commitment txs in IndexerVtxo not a single one.

altafan:
This PR includes all the required changes to api-spec, server and SDK to replace the vtxo `commitment_txid` string field with a `commitment_txids` string list.

Notable changes:
- SDK:
   * Replaced the struct that was representing a tx history record in the indexer client with the `types.Transaction`
   * Made the sdk client fetch the tx history from the indexer instead of computing it from vtxos
   * Stored the commitment txids of a vtxo in the sqlite repo as a single string of comma separated txids
   * Updated computing the redeemable branches only for vtxo leaves (there was already no support for those preconfirmed, not added here as out of scope)
- Server:
   * Replaced the vtxo domain `CommitmentTxid string` field with `RootCommitmentTxid string` and `CommimentTxids []string`
   * The changes are reflected in the sql dbs by adding a N-to-N relational table with vtxo outpoint and commitment txid fields that form the unique primary key, and vtxo outpoint as foreign key referring to the vtxo table.
   There could be room for improvement and set also the commitment txid as fkey that refers to the tx table but this would require a lot of changes in the unit tests and it's not really relevant, as we only read the commitment txids without making any query over them. A handy vtxo_vw VIEW is introduced to simplify fetching a vtxo with commitment txids.
   In this view, txids are returned as a unique string of comma separated hashes that are properly parsed in the relative repo impl. Tests make sure everything's ok (even tho the original order is not preserved when reading but not important)

Closes #615.

Please @louisinger @Kukks review.


pkg/client-sdk/redemption/redeem.go
```diff
@@ -90,7 +91,15 @@ func (r *CovenantlessRedeemBranch) RedeemPath() ([]string, error) {
 func (r *CovenantlessRedeemBranch) ExpiresAt() (*time.Time, error) {
 	lastKnownBlocktime := int64(0)
 
-	confirmed, blocktime, _ := r.explorer.GetTxBlockTime(r.vtxo.CommitmentTxid)
+	var confirmed bool
+	blocktime := int64(math.MaxInt64)
+	for _, txid := range r.vtxo.CommitmentTxids {
+		rConfirmed, rBlocktime, _ := r.explorer.GetTxBlockTime(txid)
+		if confirmed && rBlocktime < blocktime {
```
louisinger, line 98: if one of the commitment tx is not confirmed, we should update the rBlocktime  to `expirationFromNow := time.Now().Add(time.Minute).Add(r.vtxoTreeExpiry)`

pkg/client-sdk/store/sql/migration/20250624211039_init.up.sql
```diff
@@ -3,7 +3,7 @@ CREATE TABLE IF NOT EXISTS vtxo (
 	vout INTEGER NOT NULL,
 	script TEXT NOT NULL,
 	amount INTEGER NOT NULL,
-	commitment_txid TEXT NOT NULL,
+	commitment_txids TEXT NOT NULL,
```
louisinger, line 6: postgres has an `ARRAY` type, maybe worth to use it ?

altafan: this is client's SQLite store tho

tiero: It's sounds an opportunity to have a table then instead of having one string that is hard to query, encode/decode in memory etc...Or virtual table whatever is called

tiero: https://sqlite.org/vtab.html#table_valued_functions

tiero: Not sure we need to do query tho. If we always return all the list all the time (no granular filtering) and never update that field often, may be ok cc/ @Kukks

Issue #609, opened 2025-06-13
Introduce more tx types for indexer txs


Kukks:
Currently we have commitment transaction and virtual, but I would like the virtual to be more defined: 

* Tree Transaction
* Ark transaction
* Checkpoint
* Forfeit

The reason being that clients need to watch onchain in case an ark tx goes on chain, but not an ark tx in order to activate double spend protection by unilater exits. 

We willneed this for the Akrade signer to differentiate as well.

louisinger:
closed by #635

PR #635, opened 2025-06-25
Fix and update GetVtxoChain API

louisinger:
This PR adds checkpoints transactions and tree transactions to the list returned by GetVtxoChain RPC.
It also rework the SDK unilateral exit to use that endpoint instead of GetFullVtxoTree, letting to support unilateral exit of preconfirmed vtxos.

it closes https://github.com/arkade-os/arkd/issues/609

@altafan @Kukks please review


api-spec/openapi/swagger/ark/v1/indexer.swagger.json
```diff
@@ -957,8 +942,9 @@
       "type": "string",
       "enum": [
         "INDEXER_CHAINED_TX_TYPE_UNSPECIFIED",
-        "INDEXER_CHAINED_TX_TYPE_VIRTUAL",
-        "INDEXER_CHAINED_TX_TYPE_COMMITMENT"
+        "INDEXER_CHAINED_TX_TYPE_ARK",
```
Kukks, line 946: No forfeit?

Kukks: How do you know the root of the chain (commitment tx)

louisinger: to fetch forfeit we have `GetForfeitTxs`, no need to return them also here ? 

In my mind the main use case for `GetVtxoChain` is when you receive a preconfirmed VTXO and you want to store and/or verify the chain leading to your coin (so the VTXO is unspent = no forfeit), wdyt?

louisinger: > How do you know the root of the chain (commitment tx)

`GetVtxoChainResponse` still returns the root_commitment_txid (should maybe be a list to fit the new Vtxo type?). 

Moreover `GetVtxos` RPC also allows to get the commitment txids

altafan: yeah forfeits are not returned by this api, but already agreed offline on:
- type `ark` for every non-tree tx
- type `tree` for every tree tx
- type `commitment`  for any commitment tx

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -285,13 +283,8 @@ message IndexerVtxo {
 
 message IndexerChain {
   string txid = 1;
-  repeated IndexerChainedTx spends = 2;
```
altafan, line 288: this must be restored

louisinger: done

altafan (approved):

PR #642, opened 2025-06-27
Ark Up metric

sekulicd:
This adds ark_service_up metric that can/will be used by Alerting system to register that Ark is down.

@altafan please review.

### DIFF
--- a/docker-compose.regtest.yml
+++ b/docker-compose.regtest.yml
@@ -40,7 +40,6 @@ services:
       - ARK_LIVE_STORE_TYPE=inmemory
       - ARK_EVENT_DB_TYPE=badger
       - ARK_DB_TYPE=sqlite
-      - ARK_OTEL_COLLECTOR_ENDPOINT=http://otel-collector:4318
     volumes:
       - type: tmpfs
         target: /app/data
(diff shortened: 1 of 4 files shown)


server/internal/interface/grpc/otel.go
```diff
@@ -70,7 +70,7 @@ func initOtelSDK(ctx context.Context, otelCollectorUrl string) (func(context.Con
 
 	reader := sdkmetric.NewPeriodicReader(
 		metricExp,
-		sdkmetric.WithInterval(5*time.Second),
+		sdkmetric.WithInterval(10*time.Second),
```
tiero: Should be configurable via ENVV

docker-compose.regtest.yml
```diff
@@ -40,6 +40,7 @@ services:
       - ARK_LIVE_STORE_TYPE=inmemory
       - ARK_EVENT_DB_TYPE=badger
       - ARK_DB_TYPE=sqlite
+      - ARK_OTEL_COLLECTOR_ENDPOINT=http://otel-collector:4318
```
altafan: revert please

server/internal/interface/grpc/otel.go
sekulicd: Done

docker-compose.regtest.yml
sekulicd: Done

PR #649, opened 2025-07-02
Fix stack exceed panic

altafan:
This fixes a panic that happened after a very long time of the server's inactivity.

The panic was due to the fact that we invoked the function handling the next phase of the round execution in a `defer`ed function, so basically the functions were never returning, and eventually this would have made logrus crash because of too many Debugf invokations in the same context.

To fix the bug, it was enough to spawn go routines for the invocation of such functions. Then, I also added fields to the `service` struct to handle the interruption of the spawned go routines nicely when the server is switched off, preserving the graceful stop behavior already implemented in arkd.

Please @Kukks @sekulicd @louisinger review.

## Summary by CodeRabbit

* **New Features**
  * Improved service shutdown process to ensure all ongoing operations complete gracefully.
  * Enhanced reliability and stability during service lifecycle transitions.

* **Bug Fixes**
  * Fixed potential issues with incomplete shutdowns by synchronizing all background operations before stopping the service.

### DIFF
--- a/server/internal/core/application/service.go
+++ b/server/internal/core/application/service.go
@@ -65,8 +65,12 @@ type covenantlessService struct {
 	eventsCh                 chan []domain.Event
 	transactionEventsCh      chan TransactionEvent
 	forfeitsBoardingSigsChan chan struct{}
-	// TODO remove this in v7
-	indexerTxEventsCh chan TransactionEvent
+	indexerTxEventsCh        chan TransactionEvent
+
+	// stop and go routine handlers
+	stop func()
+	ctx  context.Context
+	wg   *sync.WaitGroup
 }
 
 func NewService(
@@ -124,6 +128,8 @@ func NewService(
 		utxoMinAmount = int64(dustAmount)
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+
 	svc := &covenantlessService{
 		network:                   network,
 		pubkey:                    pubkey,
@@ -151,6 +157,9 @@ func NewService(
 		vtxoMinSettlementAmount:   vtxoMinSettlementAmount,
 		vtxoMinOffchainTxAmount:   vtxoMinOffchainTxAmount,
 		indexerTxEventsCh:         make(chan TransactionEvent),
+		stop:                      cancel,
+		ctx:                       ctx,
+		wg:                        &sync.WaitGroup{},
 	}
 
 	repoManager.Events().RegisterEventsHandler(
@@ -301,6 +310,7 @@ func (s *covenantlessService) Start() error {
 	}
 
 	log.Debug("starting app service...")
+	s.wg.Add(1)
 	go s.start()
 	return nil
 }
@@ -308,6 +318,8 @@ func (s *covenantlessService) Start() error {
 func (s *covenantlessService) Stop() {
 	ctx := context.Background()
 
+	s.stop()
+	s.wg.Wait()
 	s.sweeper.stop()
 	// nolint
 	vtxos, _ := s.repoManager.Vtxos().GetAllSweepableVtxos(ctx)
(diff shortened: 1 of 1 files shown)


altafan:
fbe45a6af043e2af2c2861075b1fcc596ffadc5a shows you how to test this. If you try to apply the change on next-version, you'll see that arkd crashes almost immediately (you may need to redirect the logs to a file if you want to check the whole stack trace), while here instead arkd keeps creating and aborting rounds indefinitely.

altafan:
@coderabbitai review intensively

altafan:
@coderabbitai review

server/internal/core/application/service.go
```diff
@@ -1484,16 +1504,23 @@ func (s *covenantlessService) startRound() {
...
 	log.Debugf("started registration stage for new round: %s", round.Id)
+
+	roundTiming := newRoundTiming(s.roundInterval)
+	<-time.After(roundTiming.registrationDuration())
+	s.wg.Add(1)
+	go s.startConfirmation(roundTiming)
 }
 
 func (s *covenantlessService) startConfirmation(roundTiming roundTiming) {
+	defer s.wg.Done()
```
sekulicd, line 1516: Instead of handling wg groups and ctx.Done in each func you can do bellow

```
// runStageAsync starts a long-living “stage” goroutine.
// It handles:
//
//   • wg.Add / wg.Done
//   • panic recovery with stack trace
//   • bails out immediately if ctx is already cancelled
//
func runStageAsync(wg *sync.WaitGroup, ctx context.Context, fn func(ctx context.Context)) {
    // If the parent context was already cancelled, skip the launch
    if ctx.Err() != nil {
        return
    }

    wg.Add(1)
    go func() {
        defer wg.Done()
        defer func() {
            if r := recover(); r != nil {
                log.Errorf("panic: %v\n%s", r, debug.Stack())
            }
        }()

        // Exit straight away if the context is now cancelled
        select {
        case <-ctx.Done():
            return
        default:
        }

        fn(ctx)
    }()
}
```

You can run it with

```
runStageAsync(s.wg, s.ctx, func(ctx context.Context) {
    s.startConfirmation(ctx, rt)
})
```

sekulicd (commented):
This is important fix and definitely something we didnt catch considering panic recovery doesnt "cover" stack overflow.
I think in future we should improve round process for clarity, maintainability but also for scalability and option to decouple signer etc.

What i feel is that we should not spawn go routine for each stage as round is sequential so we should follow one go routine pattern, smtn like bellow:

```
// engine spins forever (until top-level ctx is canceled) and runs rounds
// engine spins forever (until top-level ctx is canceled) and runs rounds
func (s *covenantlessService) engine(ctx context.Context) {
	for {
		roundCtx, cancel := context.WithCancel(ctx)
		err := s.runRound(roundCtx) // ONE goroutine – step through stages
		cancel()                    // shorten the life of stage helpers

		if err != nil {
			log.Warnf("round aborted: %v", err)
		}

		// graceful shutdown
		if ctx.Err() != nil {
			return
		}
	}
}

// runRound executes Registration → Confirmation → Finalisation;
// it returns nil when the round completes successfully,
// or a wrapped error if the round failed or was canceled.
func (s *covenantlessService) runRound(ctx context.Context) error {
	rt := newRoundTiming(s.roundInterval) // helper that knows the durations

	// ---------- 1. REGISTRATION ----------
	round, err := s.stageRegistration(ctx, rt)
	if err != nil {
		return fmt.Errorf("registration: %w", err) // engine logs & restarts
	}

	// ---------- 2. CONFIRMATION ----------
	if err := s.stageConfirmation(ctx, round, rt); err != nil {
		return fmt.Errorf("confirmation: %w", err)
	}

	// ---------- 3. FINALISATION ----------
	if err := s.stageFinalisation(ctx, round, rt); err != nil {
		return fmt.Errorf("finalisation: %w", err)
	}

	// ---------- 4. POST-ROUND maintenance ----------
	// sweep scheduling, DB persistence, etc.
	if err := s.afterRound(ctx, round); err != nil {
		log.WithError(err).Warn("post-round housekeeping failed")
		// do not abort the engine loop – move on to next round
	}

	return nil
}

```

PR #661, opened 2025-07-18
Fix mutinynet network name

altafan:
This fixes the network returned on mutinynet from.

Please @louisinger review.


louisinger (approved):
tACK

PR #702, opened 2025-09-03
Tidy up wallet & Introduce Signer

altafan:
This pr switches the _arkd-wallet_ packages so that:
* `arkd-wallet` becomes `arkd-wallet-btcwallet` deprecated, going to be dropped later
* `arkd-wallet-nbxplorer` becomes `arkd-wallet`

This way, we don't need to change anything in the release pipeline or arkd, and it also makes sense as we are going to drop support for the btcwallet wallet implementation.

~~WIP:~~

This adds a new feature to arkd for loading either a url or a private key via env vars or via admin apis.

This changes arkd-wallet (former arkd-wallet-nbxplorer) so that it stops using a default account as signer; instead, the key is provided by the operator either via env var (`ARKD_WALLET_SIGNER_KEY`) or via api (`LoadSignerKey`). If not provided, arkd-wallet just doesn't act as signer.

This changes arkd so that the signer is split from the wallet with a new `Signer` interface at application level in ports. The related implementation is just a simple grpc client for the newly introduced signer/v1 protos.

By default, arkd uses arkd-wallet as signer, but it can be customized via newly introduced env var (`ARKD_SIGNER_ADDR`) or admin api (`LoadSigner`). The api allows, alternatively,  to load a private key into arkd-wallet in case the service is not configured with a signer key (`ARKD_WALLET_SIGNER_KEY` unset).

The signer can be loaded via the newly added cli command `arkd signer load [--addr <host:port> | --prvkey <hex>]`

Please @sekulicd @Kukks review.

## Summary by CodeRabbit

- New Features
  - Introduces external Signer service and endpoints; wallet can act as signer or connect to a remote signer.
  - New endpoints: load signer key; signer status, pubkey, and signing operations.
  - New Make targets: run-wallet, run-wallet-nosigner, run-signer.
  - Added docker/regtest and env presets for wallet and signer.

- Breaking Changes
  - Replaces forfeit-address endpoint with forfeit-pubkey; removes wallet pubkey endpoint.

- CLI
  - Adds signer load and genkey commands.

- Configuration
  - NBXplorer-based wallet setup; supports ARKD_SIGNER_ADDR and signer key.

- Documentation
  - Updated setup, run commands, and signer configuration guidance.


Makefile
```diff
@@ -154,14 +154,21 @@ docker-stop:
 	@echo "Stopping dockerized arkd and arkd wallet in test mode on regtest..."
 	@docker compose -f docker-compose.regtest.yml down -v
 
-## run-wallet-bitcoind: run arkd wallet in dev mode with bitcoind on regtest
+## run-wallet: run arkd wallet based on nbxplorer in dev mode on regtest
+run-wallet:
+	@echo "Running arkd wallet in dev mode with Bitcoin Core on regtest..."
+	@docker compose -f docker-compose.regtest.yml up -d pgnbxplorer nbxplorer
+	$(call setup_env, envs/arkd-wallet.regtest.env)
+	@go run ./cmd/arkd-wallet
```
sekulicd: in dev mode maybe we just want to stop/start arkd-wallet and not both

```suggestion
## run-nbcplorer: runs nbxplorer
run-nbcplorer:
    @echo "Running arkd wallet in dev mode with Bitcoin Core on regtest..."
	@docker compose -f docker-compose.regtest.yml up -d pgnbxplorer nbxplorer

## run-wallet: run arkd wallet based on nbxplorer in dev mode on regtest
run-wallet:
	$(call setup_env, envs/arkd-wallet.regtest.env)
	@go run ./cmd/arkd-wallet
```

internal/core/application/service.go
```diff
@@ -1008,15 +1009,7 @@ func (s *service) RegisterIntent(
...
-			isOnchain := false
-			for _, index := range message.OnchainOutputIndexes {
-				if index == outputIndex {
-					isOnchain = true
-					break
-				}
-			}
-
-			if isOnchain {
+			if slices.Contains(message.OnchainOutputIndexes, outputIndex) {
```
sekulicd:
```suggestion
            hasOnchainOutput := slices.Contains(message.OnchainOutputIndexes, outputIndex)
			if hasOnchainOutput {
```

pkg/arkd-wallet-btcwallet/config/config.go
```diff
@@ -0,0 +1,186 @@
...
+
+var (
+	Port         = "PORT"
+	Datadir      = "DATADIR"
+	LogLevel     = "LOG_LEVEL"
+	Network      = "NETWORK"
+	EsploraURL   = "ESPLORA_URL"
+	NeutrinoPeer = "NEUTRINO_PEER"
+	// #nosec G101
+	BitcoindRpcUser = "BITCOIND_RPC_USER"
```
sekulicd, line 24: maybe from now on we should load secrets by providing interface to config, so we have diff providers: env, file, aws etc.

pkg/arkd-wallet-btcwallet/interface/grpc/handlers/healthservice.go
```diff
@@ -12,6 +12,16 @@ func NewHealthHandler() grpchealth.HealthServer {
 	return &healthHandler{}
 }
 
+func (h *healthHandler) List(context.Context, *grpchealth.HealthListRequest) (*grpchealth.HealthListResponse, error) {
+	return &grpchealth.HealthListResponse{
+		Statuses: map[string]*grpchealth.HealthCheckResponse{
+			"arkd-wallet": {
```
sekulicd, line 18: put arkd-wallet in const

pkg/arkd-wallet-btcwallet/interface/grpc/handlers/wallet_handler.go
```diff
@@ -300,15 +312,8 @@ func (h *WalletServiceHandler) GetCurrentBlockTime(
 func (h *WalletServiceHandler) IsTransactionConfirmed(
 	ctx context.Context, req *arkwalletv1.IsTransactionConfirmedRequest,
 ) (*arkwalletv1.IsTransactionConfirmedResponse, error) {
-	confirmed, blocknumber, blocktime, err := h.scanner.IsTransactionConfirmed(ctx, req.GetTxid())
+	confirmed, blocknumber, blocktime, err := h.walletSvc.IsTransactionConfirmed(ctx, req.GetTxid())
 	if err != nil {
-		if errors.Is(err, nbxplorer.ErrTransactionNotFound) {
```
sekulicd, line 305: what is the reasoning for removing nbxplorer.ErrTransactionNotFound check ?

Makefile
altafan: arkd-wallet at startup needs an up and running nbxplorer to connect to.. i still need to update the readme but the idea is that you have to take care of deleting the explorer when you want to tear down the whole dev env.

pkg/arkd-wallet-btcwallet/interface/grpc/handlers/wallet_handler.go
altafan: i think this is just the github diff that is messed up. I just renamed the whole folder, didn't touch anything on the btcwallet impl

Makefile
```diff
@@ -154,21 +154,23 @@ docker-stop:
...
-	@go run ./cmd/arkd-wallet-btcwallet
\ No newline at end of file
+## run-walle-without-signert: run arkd wallet based on nbxplorer in dev mode on regtest without a pre-loaded signer private key
+run-wallet-without-signer:
+	@echo "Running arkd wallet in dev mode with NBXplorer on regtest..."
+	@docker compose -f docker-compose.regtest.yml up -d pgnbxplorer nbxplorer
+	$(call setup_env, envs/arkd-wallet-nosigner.regtest.env)
+	@go run ./cmd/arkd-wallet
+
+## run-signer: run arkd wallet as signer without a wallet
```
sekulicd, line 171: without a wallet ?

Makefile
```diff
@@ -154,21 +154,23 @@ docker-stop:
...
+run-signer:
```
sekulicd, line 172: run-wallet

scripts/ec-gen-key
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
sekulicd: add makefile cmd genkey or smtn

cmd/arkd/commands.go
```diff
@@ -68,6 +75,12 @@ var (
 		Action: walletWithdrawAction,
 		Flags:  []cli.Flag{withdrawAmountFlag, withdrawAddressFlag},
 	}
+	signerLoadCmd = &cli.Command{
+		Name:   "load",
+		Usage:  "Load the ark signer address or private key",
```
sekulicd, line 87: better replace address with url as CRabbit suggested

cmd/arkd/commands.go
```diff
@@ -269,6 +282,32 @@ func walletWithdrawAction(ctx *cli.Context) error {
 	return nil
 }
 
+func signerLoadAction(ctx *cli.Context) error {
+	baseURL := ctx.String(urlFlagName)
+	signerKey := ctx.String(signerKeyFlagName)
+	signerUrl := ctx.String(signerUrlFlagName)
+	if signerKey != "" && signerUrl != "" {
+		return fmt.Errorf("either private key or url must be provided")
+	}
```
sekulicd:
```suggestion
if signerKey != "" {
		return fmt.Errorf(“signer key must be provided")
}

if signerUrl != "" {
		return fmt.Errorf(“signer url must be provided")
}
```

api-spec/protobuf/ark/v1/signer_manager.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+package ark.v1;
+
+import "google/api/annotations.proto";
+
+service SignerManagerService {
+  rpc LoadSigner(LoadSignerRequest) returns (LoadSignerResponse) {
```
sekulicd, line 8: I would split this into two RPCs:
- ConnectWithSigner
- LoadSignerKey

cmd/arkd/commands.go
sekulicd: everywhere

altafan:
> Why we wouldn't use go code to also generare a free signer key? so that we don't have more pieces around, and we retain the arkd command as the only interface to provision everything

added a `make genkey` command to generate a key. The idea is that we want to be able to start arkd-wallet without key and later load one via api, so arkd-wallet must not create a random key by default if none is provided via env var

tiero:
> so arkd-wallet must not create a random key by default 

thats fine, what I am saying is sinstead of having the creation done by a bash script, can be done by a go command under arkd ie. `arkd genkey` so we don need to sideload anythign else

altafan:
> > so arkd-wallet must not create a random key by default
> 
> thats fine, what I am saying is sinstead of having the creation done by a bash script, can be done by a go command under arkd ie. `arkd genkey` so we don need to sideload anythign else

done

PR #707, opened 2025-09-12
Migrate to grpc-api-gateway

altafan:
With this pr we change the dependency used to define rest api along rpc in protos and use `grpc-api-gateway` instead of `grpc-gateway`.

With this, we replace mapping streaming rpc to JSON stream with SSE and allow native mobile clients to consume arkd APIs.


tiero:
> time="2025-09-13T04:59:46Z" level=warning msg="websocket dial failed, falling back to polling" error="websocket: bad handshake" network=regtest url="ws://localhost:3000/v1/ws"

What we use websocket for?

PR #715, opened 2025-09-24
FinalizeTx fix: verify ark signer signature in checkpoints

### LINKED ISSUE #700: Submitting checkpoint PSBT without server sig via `/ark.v1.ArkService/FinalizeTx` breaks unilateral exit
When working on the upgrade to `arkd:0.7.0` for `arkade-os/rust-sdk`, I ran into this situation when sending to an Ark address:

1. Client calls `/ark.v1.ArkService/SubmitTx`.
2. Server responds with signed Ark PSBT and checkpoint PSBTs.
3. Client _replaces_ the server's signature on a checkpoint PSBT with its own signature and calls `/ark.v1.ArkService/FinalizeTx`.
4. Server accepts the request.
5. Client attempts to perform unilateral exit, calling `/ark.v1.IndexerService/GetVirtualTxs` in the process.
6. Unilateral exit fails on broadcasting a checkpoint transaction missing the server signature: `mandatory-script-verify-flag-failed (Invalid Schnorr signature)`.

Step 3 is a client implementation error (which I've fixed since we first ran into this), but I think the Ark server should not allow the client to submit an incomplete checkpoint PSBT when calling `/ark.v1.ArkService/FinalizeTx`, which then leads to this bug if the client relies on `/ark.v1.IndexerService/GetVirtualTxs` to perform unilateral exit (I realise it shouldn't, but still).

---

Tested against 60595e4b46bcb951a9e3633a4a0c3584712e887e.

Check out https://github.com/arkade-os/rust-sdk/commit/682394ae69f3e5a58640d2390baf879678c6719c to reproduce the bug with the Rust SDK.

louisinger:
by default `TxBuilder.VerifyTapscriptPartialSigs` is excluding the verification of the ark signer signature because most of the time, the signer doesn't sign transactions he is verifying. However, that's not the case for checkpoint tx submitted in `FinalizeTx` RPC leading to a potential attack where the user is removing the operator's signature from the checkpoint psbt. 

the new argument `checkSignerSig` in VerifyTapscriptPartialSigs fixes this.

it closes #700 

@altafan @Kukks  please review

## Summary by CodeRabbit

* **New Features**
  * Introduced an optional operator-signature flag to make signature verification stage-aware, allowing select stages to skip or require the operator's signature.

* **Bug Fixes**
  * Reduced false verification failures during off‑chain submission, finalization checkpoints, and forfeit processing.

* **Tests**
  * Updated mocks and tests to cover the new verification behavior.

* **Chores**
  * Internal API and implementation updates to propagate the new flag; no user-facing workflow changes.

### DIFF
--- a/internal/infrastructure/tx-builder/covenantless/builder.go
+++ b/internal/infrastructure/tx-builder/covenantless/builder.go
@@ -49,16 +49,22 @@ func (b *txBuilder) GetTxid(tx string) (string, error) {
 	return ptx.UnsignedTx.TxID(), nil
 }
 
-func (b *txBuilder) VerifyTapscriptPartialSigs(tx string) (bool, string, error) {
+func (b *txBuilder) VerifyTapscriptPartialSigs(
+	tx string,
+	checkSignerSig bool,
+) (bool, string, error) {
 	ptx, err := psbt.NewFromRawBytes(strings.NewReader(tx), true)
 	if err != nil {
 		return false, "", err
 	}
 
-	return b.verifyTapscriptPartialSigs(ptx)
+	return b.verifyTapscriptPartialSigs(ptx, checkSignerSig)
 }
 
-func (b *txBuilder) verifyTapscriptPartialSigs(ptx *psbt.Packet) (bool, string, error) {
+func (b *txBuilder) verifyTapscriptPartialSigs(
+	ptx *psbt.Packet,
+	checkSignerSig bool,
+) (bool, string, error) {
 	txid := ptx.UnsignedTx.TxID()
 
 	signerPubkey, err := b.signer.GetPubkey(context.Background())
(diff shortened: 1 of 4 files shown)


internal/infrastructure/tx-builder/covenantless/builder.go
```diff
@@ -49,16 +49,22 @@ func (b *txBuilder) GetTxid(tx string) (string, error) {
...
 	}
 
-	return b.verifyTapscriptPartialSigs(ptx)
+	return b.verifyTapscriptPartialSigs(ptx, checkSignerSig)
 }
 
-func (b *txBuilder) verifyTapscriptPartialSigs(ptx *psbt.Packet) (bool, string, error) {
+func (b *txBuilder) verifyTapscriptPartialSigs(
+	ptx *psbt.Packet,
+	checkSignerSig bool,
```
altafan:
```suggestion
	mustIncludeSignerSig bool,
```

PR #726, opened 2025-10-02
Rework Ark PSBT fields

louisinger:
This PR reworks the custom PSBT input fields and fix its encoding. Including a constant PSBT key type (`255`).

The abstraction is inspired by [ts-sdk one](https://github.com/arkade-os/ts-sdk/blob/master/src/utils/unknownFields.ts). 

It also properly adds the PSBT key type `255` when serializing the field Key.

@altafan @Kukks please review

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Taproot/taptree extraction, validation and error messages (now include checkpoint context) to reduce malformed transaction failures.

- **Refactor**
  - Consolidated PSBT metadata into a unified Ark PSBT field system; standardized encoding for taproot trees, condition witnesses, cosigners and expiry; updated signing, sweep and intent flows to use the new fields.

- **Tests**
  - Updated unit and e2e tests for the new PSBT field APIs.

- **Chores**
  - Bumped ark-lib and go-sdk deps; unified test runner; updated RPC permission mappings.


internal/core/application/service.go
```diff
@@ -488,7 +483,22 @@ func (s *service) SubmitOffchainTx(
...
 		}
 
-		vtxoScript, err := script.ParseVtxoScript(tapscripts)
+		taptreeFields, err := txutils.GetArkPsbtFields(
+			checkpointPsbt,
+			0,
+			txutils.VtxoTaprootTreeField,
+		)
+		if err != nil {
+			return nil, "", "", fmt.Errorf("missing taptree: %s", err)
```
```suggestion
			return nil, "", "", fmt.Errorf("failed to extract taptree from tx: %s", err)
```

internal/infrastructure/tx-builder/covenantless/builder.go
```diff
@@ -1256,11 +1242,25 @@ func (b *txBuilder) extractSweepLeaf(input psbt.PInput) (
...
+	}
+
+	if len(cosignerFields) == 0 {
+		return nil, nil, nil, fmt.Errorf("no cosigner pubkeys found")
+	}
+
+	cosignerPubKeys, err := txutils.MakeCosignerPublicKeyList(cosignerFields)
+	if err != nil {
+		return nil, nil, nil, err
+	}
```
altafan: shall we abstract this code into a single function in `txutils.GetCosignersFromArkPsbt(pts, inputIndex)`?

louisinger: IMO better to "split" so the sorting is explicitly used only when useful (sometimes we don't need the keys to be ordered). wdyt ?

altafan: if one doesn't want the func that combines txutils.GetArkPsbtFields + txutils.MakeCosignerPublicKeyList, he would still be free to use these functions... i'm not saying to replace 2 funcs with 1, but to have 1 that calls both

PR #762, opened 2025-10-13
OffchainTx: set vtxo.CreatedAt to StartingTimestamp

louisinger:
For precomfirmed vtxos, set CreatedAt timestamp to `StartingTimestamp` (time of `SubmitOffchainTx`) instead of `EndingTimestamp` (time of `FinalizeOffchainTx`).

@altafan please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated timestamps for newly created outputs from off‑chain transactions to reflect the transaction start time, improving accuracy in activity logs and details.
  * Ensures consistent ordering and display of recent activity based on when a transaction began.

* **Chores**
  * Internal consistency improvements to timestamp handling for off‑chain transaction outputs, with no changes to public interfaces.

### DIFF
--- a/internal/core/application/utils.go
+++ b/internal/core/application/utils.go
@@ -134,7 +134,7 @@ func decodeTx(offchainTx domain.OffchainTx) (string, []domain.Outpoint, []domain
 			CommitmentTxids:    offchainTx.CommitmentTxidsList(),
 			RootCommitmentTxid: offchainTx.RootCommitmentTxId,
 			Preconfirmed:       true,
-			CreatedAt:          offchainTx.EndingTimestamp,
+			CreatedAt:          offchainTx.StartingTimestamp,
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


tiero:
> some point the tx might be replaceable

After an expiration is better ie. Client have 30 seconds to finalize or they need to sign it again

Kukks:
> > some point the tx might be replaceable
> 
> > After an expiration is better ie. Client have 30 seconds to finalize or they need to sign it again
> 
> @altafan @tiero impossible because after the server has signed the transaction + the checkpoint, it **can't allow any other transaction spending the same inputs** : there is a risk to "fork" the offchain branch = allow double spend.
> 
> > maybe in this case we should block the finalization because the "spent" vtxo expired in the meantime?
> 
> IMO better to not block the finalization but instead mark them "swept". This way, the new vtxos can be recovered in next batches.

Agreed here, @louisinger and I already explored these scenarios at length at previous sprints

PR #813, opened 2025-11-17
Fix redis implementation of current round

altafan:
This adds a context parm to the apis of the current round interface in cache and also fixes the redis implementation of the `Upsert` method.

NOTE: I also made some changes to the app svc SignCommitmentTx and moved the cache upsert call inside a go routine.

Please @louisinger @sekulicd @Kukks review.

## Summary by CodeRabbit

* **Refactor**
  * Round lifecycle and event flows now consistently thread execution context, making state updates and error propagation context-aware.

* **New Features**
  * Persistent boarding-signature store with add/get/delete operations.
  * Transaction verification now returns detailed per-input signature and script data.

* **Bug Fixes**
  * Fixed finalization and signature-handling issues that could cause missed or duplicate signals.

* **Tests**
  * Tests updated to use context-aware APIs and improved e2e timing/stability.

* **Chores**
  * Local dev DB port and related env settings updated.


internal/core/application/service.go
```diff
@@ -1706,16 +1704,17 @@ func (s *service) SignCommitmentTx(ctx context.Context, signedCommitmentTx strin
...
+			return round
+		}); err != nil {
+			log.WithError(err).WithField("signed_commitment_tx", signedCommitmentTx).
+				Errorf("failed to upsert current round: %v", err)
+			return
+		}
 
-	go s.checkForfeitsAndBoardingSigsSent()
+		s.checkForfeitsAndBoardingSigsSent()
+	}()
```
louisinger: if Upsert fails, we lost the signature, but the user does not see any error from the API ? 

I have the feeling we should stop mutating the round psbt in SignCommitmentTx, but instead we can push only the partial signatures to the cache. Then, during round finalization, we put that signatures in the psbt. Thus, only 1 routine is mutating the transaction. And this operation becomes a simple "push sig to cache".

internal/core/ports/live_store.go
```diff
@@ -53,9 +53,9 @@ type OffChainTxStore interface {
 }
 
 type CurrentRoundStore interface {
-	Upsert(fn func(m *domain.Round) *domain.Round) error
-	Get() *domain.Round
-	Fail(err error) []domain.Event
+	Upsert(ctx context.Context, fn func(m *domain.Round) *domain.Round) error
```
louisinger, line 56:
```suggestion
	Upsert(ctx context.Context, round *domain.Round) error
```

We never use the functional pattern, can't we simplify the signature ?

internal/core/application/service.go
```diff
@@ -3050,10 +3079,11 @@ func (s *service) propagateRoundSigningNoncesGeneratedEvent(
 	publicNoncesMap map[string]tree.TreeNonces,
 	vtxoTree *tree.TxTree,
 ) {
-	events := treeTxNoncesEvents(vtxoTree, s.cache.CurrentRound().Get().Id, publicNoncesMap)
+	ctx := context.Background()
```
louisinger: Why not passing the context created in `startFinalization` ?

internal/core/application/service.go
```diff
@@ -2746,6 +2753,45 @@ func (s *service) finalizeRound(roundTiming roundTiming) {
...
+		round.CommitmentTx = commitmentTxWithSignedBoardingIns
+		if err := s.cache.CurrentRound().Upsert(ctx, func(_ *domain.Round) *domain.Round {
+			return round
+		}); err != nil {
+			changes = s.cache.CurrentRound().Fail(ctx, errors.INTERNAL_ERROR.New(
+				"failed to update round in cache: %s", err,
+			))
+			return
+		}
+
```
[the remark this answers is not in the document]
altafan: I agree with the first fix; I don't really understand the second, though. The unit and integration tests pass, and that proves you're wrong?

internal/infrastructure/live-store/redis/round.go
```diff
@@ -182,3 +189,159 @@ func (b *boardingInputsStore) Get() int {
...
+			}
+			return nil
+		}); err == nil {
+			return nil
+		}
+		<-time.After(100 * time.Millisecond)
+	}
+
+	return err
+}
```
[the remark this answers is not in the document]
altafan: > or range b.numOfRetries does not compile because numOfRetries is an int, not a rangeable type.

unit tests pass, you're wrong?

>  any network/EXEC error will bubble up in `err`

bubble up? how?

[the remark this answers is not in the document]
altafan: what if i use Exec().Err() instead of just .Err()?

internal/infrastructure/tx-builder/covenantless/builder.go
```diff
@@ -1035,59 +1034,71 @@ func (b *txBuilder) VerifyBoardingTapscriptSigs(
...
+// 				len(sourceTx.Inputs),
+// 			)
+// 		}
+// 		if len(destinationTx.Inputs) <= inputIndex {
+// 			return "", fmt.Errorf(
+// 				"input index out of bounds %d, len(inputs)=%d",
+// 				inputIndex,
+// 				len(destinationTx.Inputs),
+// 			)
+// 		}
```
louisinger: remove ?

altafan: done

internal/infrastructure/live-store/redis/round.go
```diff
@@ -182,3 +189,160 @@ func (b *boardingInputsStore) Get() int {
...
+	// Transactional update with retry logic
+	for range b.numOfRetries {
+		if err = b.rdb.Watch(ctx, func(tx *redis.Tx) error {
+			_, err := tx.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
+				for _, fv := range fields {
+					pipe.HSetNX(ctx, key, fv.field, fv.value)
+				}
+				return nil
+			})
+			return err
```
sekulicd, line 223:
```suggestion
			}, key)
			return err
```

Just one note on Redis transactions: here we use Watch(ctx, fn, key) + TxPipelined. Watch adds optimistic locking on the given key, and inside the callback TxPipelined wraps the commands in a MULTI/EXEC block so they execute as a single atomic batch if the watched key hasn’t changed; if it has changed, Redis aborts EXEC and go-redis returns redis.TxFailedErr, which we can retry. Unlike SQL transactions, Redis does not provide full rollback semantics: commands inside EXEC are not undone if one fails or if something like HSETNX just returns 0 for an existing field – the batch is atomic with respect to other clients, but it’s not all-or-nothing in the SQL sense.

altafan: done

Issue #728, opened 2025-10-03
Update topics without restarting stream


Kukks:
ALlow updating the topics without restarting the event stream.

PR #829, opened 2025-11-23
rpc for updating event stream topics

bitcoin-coder-bob:
Issue: #728 

Adds new rpc `UpdateStreamTopics` allowing client to update topics on their rpc stream from `GetEventStream`. Offers ability to add or remove topics at the same time, or to overwrite all topics at once by specifying a list of topics.

A new event type `StreamStartedEvent` was added that gets sent immediately on the creation of a stream from `GetEventStream` and returns the id of the listener. This id is then used by the client when calling `UpdateStreamTopics`.

Changes to the `go-sdk` repo in the directory `/go-sdk/blob/master/client` will need to be made to support calling this new rpc. I do not see an appropriate place in this arkd repo to call this new rpc without it also being supported in the `go-sdk` repo. It appears the `e2e_test.go` would be the appropriate place to add the test for this new call once supported in the `go-sdk`. I have added these changes here: https://github.com/arkade-os/go-sdk/pull/75. This PR you are reading calls the sdk inside the `e2e_test.go` to test out the new sdk calls for updating stream topics.

## Summary by CodeRabbit

* **Documentation**
  * Fixed a typo in the README.

* **New Features**
  * Event streams now emit a StreamStarted event immediately with a stream ID.
  * Added an API to update stream topics (add, remove, or overwrite) and return updated topic lists.

* **Tests**
  * Added tests for topic overwrite and concurrent modifications; e2e handlers now surface stream-start events.

* **Chores**
  * Bumped Go/gRPC dependencies; simplified client call signatures (SendOffChain, Settle).


internal/interface/grpc/permissions/permissions.go
```diff
@@ -156,6 +156,10 @@ func Whitelist() map[string][]bakery.Op {
 			Entity: EntityArk,
 			Action: "read",
 		}},
+		fmt.Sprintf("/%s/UpdateStreamTopics", arkv1.ArkService_ServiceDesc.ServiceName): {{
+			Entity: EntityArk,
+			Action: "read",
```
bitcoin-coder-bob: should this be `write` ?

internal/interface/grpc/handlers/broker.go
```diff
@@ -96,6 +100,8 @@ func (h *broker[T]) getTopics(id string) []string {
 	if !ok {
 		return nil
 	}
+	listener.lock.Lock()
+	defer listener.lock.Unlock()
```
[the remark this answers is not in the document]
altafan: @bitcoin-coder-bob  we should follow the suggestion here.

Also, maybe worth adding add/remove/overwrite funcs to the `listener` struct so we handle the lock there and here in broker we just invoke them?

internal/interface/grpc/permissions/permissions.go
altafan: yep

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -222,9 +234,27 @@ message GetEventStreamResponse {
...
 
+// Adding and removing topics can both be supplied in the same request,
+// allowing for simultaneous changes.
+// overwrite_topics will take precedence, and if set, then the add/remove
+// topics are ignored. The stream_id is required.
+message UpdateStreamTopicsRequest {
+  string stream_id = 1;
+  repeated string add_topics = 2;
+  repeated string remove_topics = 3;
+  repeated string overwrite_topics = 4;
```
Kukks: Can we use oneOf here to make it clearer by definition?

internal/interface/grpc/handlers/broker.go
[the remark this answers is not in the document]
bitcoin-coder-bob: good idea I will add these

api-spec/protobuf/ark/v1/service.proto
bitcoin-coder-bob: yes I like that, will add

internal/interface/grpc/handlers/broker.go
```diff
@@ -41,6 +45,55 @@ func (l *listener[T]) includesAny(topics []string) bool {
...
+	l.lock.Lock()
+	defer l.lock.Unlock()
+	newTopics := make(map[string]struct{}, len(topics))
+	for _, topic := range topics {
+		newTopics[formatTopic(topic)] = struct{}{}
+	}
+	l.topics = newTopics
+}
+
+func (l *listener[T]) GetTopics() []string {
```
louisinger: This change seems to not be needed, we can keep using the existing `getTopics`

internal/interface/grpc/handlers/broker.go
```diff
@@ -41,6 +45,55 @@ func (l *listener[T]) includesAny(topics []string) bool {
...
+	defer l.lock.Unlock()
+	if l.topics == nil {
+		return
+	}
+	for _, topic := range topics {
+		delete(l.topics, formatTopic(topic))
+	}
+}
+
+func (l *listener[T]) OverwriteTopics(topics []string) {
```
louisinger: no need to be exported, it is only used within the package

internal/interface/grpc/handlers/broker.go
```diff
@@ -41,6 +45,55 @@ func (l *listener[T]) includesAny(topics []string) bool {
...
+func (l *listener[T]) RemoveTopics(topics []string) {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+	if l.topics == nil {
+		return
+	}
+	for _, topic := range topics {
+		delete(l.topics, formatTopic(topic))
+	}
+}
```
louisinger: Similar to GetTopics change, it seems not needed. can't we use the existing `removeTopics` ?

internal/interface/grpc/handlers/broker.go
```diff
@@ -41,6 +45,55 @@ func (l *listener[T]) includesAny(topics []string) bool {
...
+	l.lock.Lock()
+	defer l.lock.Unlock()
+	if l.topics == nil {
+		l.topics = make(map[string]struct{}, len(topics))
+	}
+	for _, topic := range topics {
+		l.topics[formatTopic(topic)] = struct{}{}
+	}
+}
+
```
louisinger: not needed ?

internal/interface/grpc/handlers/broker.go
```diff
@@ -41,6 +45,55 @@ func (l *listener[T]) includesAny(topics []string) bool {
...
+	}
+	return out
+}
+
+func (l *listener[T]) Channel() chan T {
+	// no strong reason to lock here, but keep RLock to be safe if ch could be replaced
+	l.lock.RLock()
+	defer l.lock.RUnlock()
+	return l.ch
+}
```
louisinger: > could be replaced 

in theory, it is never replaced, so I'd remove this method and keep accessing via `l.ch`, except if I'm missing something ? In any case, we shouldn't export the method

internal/interface/grpc/handlers/broker.go
```diff
@@ -79,68 +132,65 @@ func (h *broker[T]) removeListener(id string) {
...
-	if _, ok := h.listeners[id]; !ok {
+	h.lock.RLock()
+	listener, ok := h.listeners[id]
+	h.lock.RUnlock()
+	if !ok {
 		return fmt.Errorf("subscription %s not found", id)
 	}
+	listener.OverwriteTopics([]string{})
+	return nil
+}
```
louisinger, line 177: what's the reason to change removeAllTopics ? I see the lock change, but then the lock is acquired again by OverwriteTopics. it seems to me we can revert that change

api-spec/protobuf/ark/v1/service.proto
```diff
@@ -222,9 +234,38 @@ message GetEventStreamResponse {
...
+// Adding and removing topics can both be supplied in the same request,
+// allowing for simultaneous changes.
+// overwrite_topics will take precedence, and if set, then the add/remove
+// topics are ignored. The stream_id is required.
+message UpdateStreamTopicsRequest {
+  string stream_id = 1;
+
+  oneof topics_change {
+    ModifyTopics modify = 2;
+    OverwriteTopics overwrite = 3;
```
louisinger, line 282: I like the overwrite feature, but do we have real needs for it @Kukks ? If not a common use case, maybe worth to let client handles it (by removing current topics and adding new ones) so we simplify the proto spec

internal/interface/grpc/handlers/broker.go
bitcoin-coder-bob: Per this comment I made the change: https://github.com/arkade-os/arkd/pull/829#discussion_r2554103903

This idea of breaking these listener functions out was so that we use the specific listener's lock.

Should I still revert back?

internal/interface/grpc/handlers/broker.go
bitcoin-coder-bob: Per this comment I made the change: https://github.com/arkade-os/arkd/pull/829#discussion_r2554103903

This idea of breaking these listener functions out was so that we use the specific listener's lock.

Should I still revert back?

internal/interface/grpc/handlers/broker.go
bitcoin-coder-bob: Per this comment I made the change: https://github.com/arkade-os/arkd/pull/829#discussion_r2554103903

This idea of breaking these listener functions out was so that we use the specific listener's lock.

Should I still revert back?

internal/interface/grpc/handlers/broker.go
bitcoin-coder-bob: Per this comment I made the change: https://github.com/arkade-os/arkd/pull/829#discussion_r2554103903

This idea of breaking these listener functions out was so that we use the specific listener's lock.

Should I still revert back?

in the meantime I fix, so that we do not export the method

internal/interface/grpc/handlers/broker.go
bitcoin-coder-bob: we can do the behavior of removing all the topics by calling `OverwriteTopics` with an empty array of topics. Calling `OverwriteTopics` uses the listener's locks. The old way of setting the topics directly in `removeAllTopics` did not use the listener's locks.

internal/interface/grpc/handlers/broker.go
```diff
@@ -41,6 +45,55 @@ func (l *listener[T]) includesAny(topics []string) bool {
...
+	}
+	return out
+}
+
+func (l *listener[T]) channel() chan T {
+	// no strong reason to lock here, but keep RLock to be safe if ch could be replaced
+	l.lock.RLock()
+	defer l.lock.RUnlock()
+	return l.ch
+}
```
altafan: better to revert and drop this function. The lock is used to handle concurrent access to `l.topics`. If we ever need to handle concurrent access to the channel (and it seems not to be the case here) we'll add a dedicated mutex.

internal/interface/grpc/handlers/broker_test.go
```diff
@@ -173,6 +173,38 @@ func TestBroker(t *testing.T) {
 		require.ErrorContains(t, err, "subscription non-existent not found")
 	})
 
+	t.Run("overwriteTopics", func(t *testing.T) {
```
altafan, line 176: guess it makes sense to add also a dedicated test for concurrency here since we added support for it

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -258,6 +270,62 @@ func (h *handler) GetEventStream(
...
+
+	switch topicsChange := req.GetTopicsChange().(type) {
+	case nil:
+		return nil, status.Error(codes.InvalidArgument, "no topics provided")
+	// when overwrite topics is provided, it takes precedence, we will not
+	// process add/remove topics in this case
+	case *arkv1.UpdateStreamTopicsRequest_Overwrite:
+		if topicsChange.Overwrite == nil {
+			return nil, status.Error(codes.InvalidArgument, "overwrite topics is nil")
+		}
```
altafan: a long time ago, we made the decision to use only getters, please fix like so also in the rest of the switch:

```suggestion
		if req.GetOverwrite() == nil {
			return nil, status.Error(codes.InvalidArgument, "missing topics to overwrite")
		}
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+	ctx context.Context,
+	req *arkv1.UpdateStreamTopicsRequest,
+) (*arkv1.UpdateStreamTopicsResponse, error) {
+	if req.GetStreamId() == "" {
+		return nil, status.Error(codes.InvalidArgument, "missing stream id")
+	}
+
+	switch req.GetTopicsChange().(type) {
+	case nil:
+		return nil, status.Error(codes.InvalidArgument, "no topics provided")
```
altafan:
```suggestion
		return nil, status.Error(codes.InvalidArgument, "missing topics")
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+	// when overwrite topics is provided, it takes precedence, we will not
+	// process add/remove topics in this case
+	case *arkv1.UpdateStreamTopicsRequest_Overwrite:
+		if req.GetOverwrite() == nil {
+			return nil, status.Error(codes.InvalidArgument, "overwrite topics is nil")
```
altafan:
```suggestion
			return nil, status.Error(codes.InvalidArgument, "missing topics to overwrite")
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+		if err := h.eventsListenerHandler.overwriteTopics(
+			req.GetStreamId(), req.GetOverwrite().GetTopics(),
+		); err != nil {
+			return nil, status.Errorf(codes.NotFound, "overwrite topics error: %s", err.Error())
+		}
+		return &arkv1.UpdateStreamTopicsResponse{
+			AllTopics:     h.eventsListenerHandler.getTopics(req.GetStreamId()),
+			TopicsAdded:   []string{},
+			TopicsRemoved: []string{},
+		}, nil
```
altafan, line 314:
```suggestion
		return &arkv1.UpdateStreamTopicsResponse{
			AllTopics:     h.eventsListenerHandler.getTopics(req.GetStreamId()),
		}, nil
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+	// allow adding/removing topics simultaneously
+	case *arkv1.UpdateStreamTopicsRequest_Modify:
+		if req.GetModify() == nil {
+			return nil, status.Error(codes.InvalidArgument, "modify topics is nil")
```
altafan:
```suggestion
			return nil, status.Error(codes.InvalidArgument, "missing topics to add or remove")
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+		}
+		if len(req.GetModify().GetAddTopics()) > 0 {
```
altafan: let's make sure topics to add/remove are not empty:

```suggestion
		if len(req.GetModify().GetAddTopics()) <= 0 && len(req.GetModify().GetRemoveTopics()) <= 0 {
		        return nil, status.Error(codes.InvalidArgument, "missing topics to add or remove")
		}
		
		if len(req.GetModify().GetAddTopics()) > 0 {
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+			return nil, status.Errorf(codes.NotFound, "overwrite topics error: %s", err.Error())
+		}
+		return &arkv1.UpdateStreamTopicsResponse{
+			AllTopics:     h.eventsListenerHandler.getTopics(req.GetStreamId()),
+			TopicsAdded:   []string{},
+			TopicsRemoved: []string{},
+		}, nil
+	// allow adding/removing topics simultaneously
+	case *arkv1.UpdateStreamTopicsRequest_Modify:
+		if req.GetModify() == nil {
```
altafan: given how extensively this is used, better to put it into a variable:
```suggestion
		modify := req.GetModify()
		if modify == nil {
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+				return nil, status.Errorf(codes.NotFound, "remove topics error: %s", err.Error())
+			}
+		}
+		return &arkv1.UpdateStreamTopicsResponse{
+			TopicsAdded:   req.GetModify().GetAddTopics(),
+			TopicsRemoved: req.GetModify().GetRemoveTopics(),
+			AllTopics:     h.eventsListenerHandler.getTopics(req.GetStreamId()),
+		}, nil
+	default:
+		return nil, status.Error(codes.InvalidArgument, "no topics provided")
```
altafan:
```suggestion
		return nil, status.Error(codes.InvalidArgument, "unknown topics to change")
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+	// when overwrite topics is provided, it takes precedence, we will not
+	// process add/remove topics in this case
+	case *arkv1.UpdateStreamTopicsRequest_Overwrite:
+		if req.GetOverwrite() == nil {
+			return nil, status.Error(codes.InvalidArgument, "overwrite topics is nil")
+		}
+		if err := h.eventsListenerHandler.overwriteTopics(
+			req.GetStreamId(), req.GetOverwrite().GetTopics(),
+		); err != nil {
+			return nil, status.Errorf(codes.NotFound, "overwrite topics error: %s", err.Error())
```
altafan:
```suggestion
			return nil, status.Error(codes.NotFound, err.Error())
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+	// allow adding/removing topics simultaneously
+	case *arkv1.UpdateStreamTopicsRequest_Modify:
+		if req.GetModify() == nil {
+			return nil, status.Error(codes.InvalidArgument, "modify topics is nil")
+		}
+		if len(req.GetModify().GetAddTopics()) > 0 {
+			if err := h.eventsListenerHandler.addTopics(
+				req.GetStreamId(), req.GetModify().GetAddTopics(),
+			); err != nil {
+				return nil, status.Errorf(codes.NotFound, "add topics error: %s", err.Error())
```
altafan:
```suggestion
				return nil, status.Error(codes.NotFound, err.Error())
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -274,6 +286,62 @@ func (h *handler) GetEventStream(
...
+			}
+		}
+		if len(req.GetModify().GetRemoveTopics()) > 0 {
+			if err := h.eventsListenerHandler.removeTopics(
+				req.GetStreamId(), req.GetModify().GetRemoveTopics(),
+			); err != nil {
+				return nil, status.Errorf(codes.NotFound, "remove topics error: %s", err.Error())
```
altafan:
```suggestion
				return nil, status.Error(codes.NotFound, err.Error())
```

internal/test/e2e/delegate_utils_test.go
```diff
@@ -213,6 +213,12 @@ func (h *delegateBatchEventsHandler) OnBatchFinalization(
 	)
 }
 
+func (h *delegateBatchEventsHandler) OnStreamStartedEvent(
+	event client.StreamStartedEvent,
+) {
+	fmt.Printf("delegate_utils_test OnStreamStartedEvent: %+v\n", event)
```
altafan, line 219: drop?

internal/test/e2e/e2e_test.go
```diff
@@ -3055,6 +3053,75 @@ func TestIntent(t *testing.T) {
 	})
 }
 
+func TestTopics(t *testing.T) {
```
altafan: Don't think an integration test is needed for this feature. This can become a unit test on the indexer client of the sdk. Here we're just good with the unit tests for now.

Please, revert the changes to the go.mod so we keep importing the go-sdk version we have on master and also drop this test.

I'll take care of making this a unit test in the indexer client in #865

pkg/ark-cli/go.mod
```diff
@@ -8,9 +8,9 @@ replace github.com/arkade-os/arkd/pkg/ark-lib => ../ark-lib
 
 require (
 	github.com/arkade-os/arkd/pkg/ark-lib v0.8.1-0.20251210130629-1fad876f84f9
-	github.com/arkade-os/go-sdk v0.8.2-0.20251210132110-e6d35e5a3c7c
+	github.com/arkade-os/go-sdk v0.0.0-20260107170900-941ddcfc3309
```
altafan: please revert also this

internal/interface/grpc/handlers/arkservice.go
bitcoin-coder-bob: we can do this but it then removes the ability for the caller to unsubscribe from all topics by passing 0 topics. When 0 topics are in the `AddTopics` of the `ModifyTopics` struct, it sets the listener's topics to be empty. Your code change suggestion would prohibit this from happening, I want to double check that is a behavior we actually want to prevent (removing all topics)

pkg/ark-cli/main.go
```diff
@@ -551,7 +550,7 @@ func sendCovenantLess(ctx *cli.Context, receivers []types.Receiver, withZeroFees
 		return printJSON(map[string]string{"txid": txid})
 	}
 
-	arkTxid, err := arkSdkClient.SendOffChain(ctx.Context, computeExpiration, offchainReceivers)
+	arkTxid, err := arkSdkClient.SendOffChain(ctx.Context, offchainReceivers)
```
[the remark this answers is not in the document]
bitcoin-coder-bob: not sure what this param was passed for

PR #836, opened 2025-11-25
Update live store

altafan:
This adds context and error return value to all methods of the interfaces that compose `ports.LiveStore`.

In the app service, this moves from relying fully on cache to store cached data in variables that are reused in the rest of the function bodies.

## Summary by CodeRabbit

* **Bug Fixes**
  * Operations now honor cancellation/timeouts and fail fast on downstream store errors to avoid silent continuation.

* **Improvements**
  * Broad context propagation across services and stores; APIs now return explicit errors for better observability and cancellation semantics.
  * Stronger error handling and safer concurrency in in-memory components.

* **New Features**
  * Retry/backoff and stricter validation for Redis-backed stores.
  * New boarding-inputs stores (in-memory and Redis) and context-aware signing/confirmation flows.

* **Tests**
  * Tests updated to use per-test contexts and handle error-aware APIs.

### DIFF
--- a/internal/infrastructure/live-store/inmemory/forfeits.go
+++ b/internal/infrastructure/live-store/inmemory/forfeits.go
@@ -154,10 +154,6 @@ func (m *forfeitTxsStore) AllSigned(_ context.Context) (bool, error) {
 	m.lock.RLock()
 	defer m.lock.RUnlock()
 
-	if len(m.forfeitTxs) == 0 {
-		return false, nil
-	}
-
 	for _, forfeit := range m.forfeitTxs {
 		if len(forfeit.Tx) == 0 {
 			return false, nil
--- a/internal/infrastructure/live-store/live_store_test.go
+++ b/internal/infrastructure/live-store/live_store_test.go
@@ -291,14 +291,14 @@ func runLiveStoreTests(t *testing.T, store ports.LiveStore) {
 
 		allSigned, err = store.ForfeitTxs().AllSigned(ctx)
 		require.NoError(t, err)
-		require.False(t, allSigned)
+		require.True(t, allSigned)
 
 		// sign after the session is deleted
 		require.Error(t, store.ForfeitTxs().Sign(ctx, []string{txs[0]}))
 
 		allSigned, err = store.ForfeitTxs().AllSigned(ctx)
 		require.NoError(t, err)
-		require.False(t, allSigned)
+		require.True(t, allSigned)
 
 		forfeitsLen, err = store.ForfeitTxs().Len(ctx)
 		require.NoError(t, err)
--- a/internal/infrastructure/live-store/redis/forfeits.go
+++ b/internal/infrastructure/live-store/redis/forfeits.go
@@ -242,10 +242,6 @@ func (s *forfeitTxsStore) AllSigned(ctx context.Context) (bool, error) {
 		return false, err
 	}
 
-	if len(hash) == 0 {
-		return false, nil
-	}
-
 	for _, forfeitJSON := range hash {
 		if len(forfeitJSON) == 0 {
 			return false, nil


internal/core/application/ban.go
```diff
@@ -148,10 +148,12 @@ func (s *service) banSignaturesCollectionTimeout(
...
 
 func (s *service) banForfeitCollectionTimeout(
-	ctx context.Context,
-	roundId string,
+	ctx context.Context, roundId string,
 ) {
-	unsignedVtxoKeys := s.cache.ForfeitTxs().GetUnsignedInputs()
+	unsignedVtxoKeys, err := s.cache.ForfeitTxs().GetUnsignedInputs(ctx)
+	if err != nil {
+		log.WithError(err).Warn("failed to get unsigned inputs from cache")
```
louisinger, line 155:
```suggestion
		log.WithError(err).Warn("failed to get unsigned inputs from cache")
		return
```
we don't want to launch GetVtxos call on empty list ?

internal/core/application/service.go
```diff
@@ -1059,12 +1066,23 @@ func (s *service) SubmitOffchainTx(
...
+				WithMetadata(map[string]any{"vtxo": spentVtxo.Outpoint.String()})
+		}
+		if isSpent {
 			return nil, errors.VTXO_ALREADY_SPENT.New("%s already spent", spentVtxo.Outpoint.String()).
 				WithMetadata(errors.VtxoMetadata{VtxoOutpoint: spentVtxo.Outpoint.String()})
 		}
 	}
-	s.cache.OffchainTxs().Add(*offchainTx)
+	if err := s.cache.OffchainTxs().Add(ctx, *offchainTx); err != nil {
+		return nil, errors.INTERNAL_ERROR.New("something went wrong").
```
louisinger, line 1083: let's move `errors.INTERNAL_ERROR.New("something went wrong")` to error.go to avoid duplicating "something went wrong"

internal/core/application/service.go
```diff
@@ -2455,21 +2549,21 @@ func (s *service) startConfirmation(
...
+			round.Fail(errors.INTERNAL_ERROR.New("failed to register intents: %s", err))
 			return
 		}
 		if err := s.cache.CurrentRound().Upsert(ctx, func(_ *domain.Round) *domain.Round {
 			return round
 		}); err != nil {
-			s.cache.CurrentRound().Fail(
-				ctx, errors.INTERNAL_ERROR.New("failed to upsert round: %s", err),
-			)
+			round.Fail(errors.INTERNAL_ERROR.New("failed to update round in cache: %s", err))
```
louisinger, line 2567: User will receive that error message, maybe worth to change the message ?

internal/core/application/service.go
```diff
@@ -2488,9 +2582,7 @@ func (s *service) startConfirmation(
 
 		// make the round fail if we didn't receive enoush confirmations
 		if len(confirmedIntents) == 0 {
-			s.cache.CurrentRound().Fail(
-				ctx, errors.INTERNAL_ERROR.New("not enough intent confirmations received"),
-			)
+			round.Fail(errors.INTERNAL_ERROR.New("not enough intent confirmations received"))
```
louisinger, line 2586:
```suggestion
			round.Fail(errors.INTERNAL_ERROR.New("not enough users confirmed participation, retry later"))
```

internal/core/application/service.go
```diff
@@ -2787,7 +2871,7 @@ func (s *service) startFinalization(
 			go s.banSignaturesCollectionTimeout(ctx, roundId, signingSession, registeredIntents)
 			return
 		case <-s.cache.TreeSigingSessions().SignaturesCollected(roundId):
-			signingSession, _ := s.cache.TreeSigingSessions().Get(roundId)
+			signingSession, _ := s.cache.TreeSigingSessions().Get(ctx, roundId)
```
louisinger: Let's check error to prevent panic in case Get failed ?

internal/core/application/service.go
```diff
@@ -2934,31 +3033,32 @@ func (s *service) finalizeRound(roundTiming roundTiming) {
...
 		s.roundReportSvc.OpEnded(WaitForForfeitTxsOp)
 
-		forfeitTxList, err := s.cache.ForfeitTxs().Pop()
+		forfeitTxList, err := s.cache.ForfeitTxs().Pop(ctx)
 		if err != nil {
-			changes = s.cache.CurrentRound().Fail(
-				ctx, errors.INTERNAL_ERROR.New("failed to finalize round: %s", err),
-			)
+			log.WithError(err).Error("failed to pop forfeit txs from cache")
+			changes = round.Fail(errors.INTERNAL_ERROR.New("failed to finalize round: %s", err))
```
louisinger, line 3050:
```suggestion
			changes = round.Fail(errors.INTERNAL_ERROR.New("failed to finalize round"))
```
let's not return err to users

internal/core/application/service.go
```diff
@@ -2934,31 +3033,32 @@ func (s *service) finalizeRound(roundTiming roundTiming) {
...
 			return
 		}
 
 		// some forfeits are not signed, we must ban the associated scripts
-		if !s.cache.ForfeitTxs().AllSigned() {
+		allForfeitTxsSigned, err := s.cache.ForfeitTxs().AllSigned(ctx)
+		if err != nil {
+			log.WithError(err).Error("failed to check all signed forfeit txs in cache")
+			changes = round.Fail(errors.INTERNAL_ERROR.New("failed to finalize round: %s", err))
```
louisinger, line 3058:
```suggestion
			changes = round.Fail(errors.INTERNAL_ERROR.New("failed to finalize round"))
```
same

internal/core/application/service.go
```diff
@@ -2934,31 +3033,32 @@ func (s *service) finalizeRound(roundTiming roundTiming) {
...
+			return
+		}
+		if !allForfeitTxsSigned {
 			go s.banForfeitCollectionTimeout(ctx, roundId)
 
-			changes = s.cache.CurrentRound().Fail(
-				ctx, errors.INTERNAL_ERROR.New("missing forfeit transactions"),
-			)
+			changes = round.Fail(errors.INTERNAL_ERROR.New("missing forfeit transactions"))
```
louisinger, line 3064:
```suggestion
			changes = round.Fail(errors.INTERNAL_ERROR.New("some users didn't submitted their forfeit in time"))
```

internal/core/application/service.go
```diff
@@ -3050,10 +3147,10 @@ func (s *service) finalizeRound(roundTiming roundTiming) {
...
 
 			txToSign, err = s.signer.SignTransactionTapscript(
-				ctx, s.cache.CurrentRound().Get(ctx).CommitmentTx, boardingInputsIndexes,
+				ctx, round.CommitmentTx, boardingInputsIndexes,
 			)
 			if err != nil {
-				changes = s.cache.CurrentRound().Fail(ctx, errors.INTERNAL_ERROR.New(
+				changes = round.Fail(errors.INTERNAL_ERROR.New(
 					"failed to sign boarding inputs of commitment tx: %s", err,
 				))
```
louisinger, line 3166:
```suggestion
				changes = round.Fail(errors.INTERNAL_ERROR.New(
"failed to sign boarding inputs of commitment tx",
))
```

internal/core/application/service.go
```diff
@@ -3083,49 +3180,44 @@ func (s *service) finalizeRound(roundTiming roundTiming) {
 
 	signedCommitmentTx, err := s.wallet.SignTransaction(ctx, txToSign, true)
 	if err != nil {
-		changes = s.cache.CurrentRound().
-			Fail(ctx, errors.INTERNAL_ERROR.New("failed to sign commitment tx: %s", err))
+		changes = round.Fail(errors.INTERNAL_ERROR.New("failed to sign commitment tx: %s", err))
```
louisinger, line 3194:
```suggestion
		changes = round.Fail(errors.INTERNAL_ERROR.New("failed to sign commitment tx"))
```

internal/infrastructure/live-store/redis/round.go
```diff
@@ -25,44 +22,56 @@ const (
...
+	if round == nil {
+		round = &domain.Round{}
+	}
+	updated := fn(round)
+	val, err := json.Marshal(updated)
+	if err != nil {
+		return err
+	}
+
+	for range s.numOfRetries {
```
louisinger, line 52: everywhere we do this retry logic, shouldn't we check is context.Done() and cancel the loops ?

internal/core/application/service.go
```diff
@@ -251,7 +251,7 @@ func NewService(
 	repoManager.Events().RegisterEventsHandler(
 		domain.RoundTopic, func(events []domain.Event) {
 			round := domain.NewRoundFromEvents(events)
-			go svc.propagateEvents(round)
+			go svc.propagateEvents(context.Background(), round)
```
sekulicd, line 254: since in this PR, beside other things, we are handling context shall we use one context on service level that would be used for cancellation so that everything is shutdown gracefully when service Stop(), which is also using context.Background, is invoked.

internal/core/application/service.go
```diff
@@ -513,13 +513,20 @@ func (s *service) SubmitOffchainTx(
 		}
 
 		// check if already spent by another offchain tx
-		if s.cache.OffchainTxs().Includes(vtxo.Outpoint) {
+		isSpent, err := s.cache.OffchainTxs().Includes(ctx, vtxo.Outpoint)
+		if err != nil {
```
sekulicd, line 517: general comment for log handling introduced in this PR 
 
 ```go
  if err != nil {
      log.WithError(err).Errorf("failed to check spent status of inputs against tx in cache")
      return nil, errors.INTERNAL_ERROR.New("something went wrong").
          WithMetadata(map[string]any{"vtxos": vtxo.Outpoint.String()})
  }

  Issues:
  1. Double handling: Logging is handling. Log+return handles the error twice
  2. Lost context: Original error is discarded. "something went wrong" is not debuggable.
  3. Error interceptor should check error and if internal hide it from users

 I would recommend to return error and then handle logging + sanitization at the gRPC interceptor level:
  - Log full error with context (once, at API boundary)
  - Return sanitized message to client based on error type

  This gives you: single log entry, full error chain for debugging, safe client messages.

internal/infrastructure/live-store/redis/round.go
altafan: in another pr we'll fix the usage of the context in app svc. For now we're fine as a graceful shutdown would wait for such loop to complete before quitting

internal/core/application/service.go
```diff
@@ -1210,7 +1236,10 @@ func (s *service) FinalizeOffchainTx(
 	}
 
 	changes = []domain.Event{event}
-	s.cache.OffchainTxs().Remove(txid)
+
+	if err := s.cache.OffchainTxs().Remove(ctx, txid); err != nil {
```
sekulicd: ### Cache not cleaned on error path

  Currently, `OffchainTxs().Remove()` here

  If FinalizeOffchainTx fails, the offchain tx remains in cache but is marked failed in DB. This causes:
  - Memory/Redis leak (keys accumulate)
  - Stale data if txid is reused
  - OffchainTxs().Includes() returns incorrect results

  Fix: Move Remove() to the defer block so it runs regardless of success/failure:

  defer func() {
      // Always clean cache regardless of outcome
      if removeErr := s.cache.OffchainTxs().Remove(ctx, txid); removeErr != nil {
          log.WithError(removeErr).Warn("failed to remove offchain tx from cache")
      }

      if structErr != nil {
          change := offchainTx.Fail(structErr)
          changes = append(changes, change)
      }
      // ... save to DB
  }()

internal/core/application/service.go
altafan: I did this because the failure reason we store should contain the real error, not the one that should be forwarded to the clients, or we lose visibility on our side. Let's focus on these changes in a dedicated pr

internal/infrastructure/live-store/inmemory/tree_signing_session.go
```diff
@@ -26,30 +26,32 @@ func NewTreeSigningSessionsStore() ports.TreeSigningSessionsStore {
...
+) (*ports.MusigSigningSession, error) {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
-	sess, ok := s.sessions[roundId]
-	return sess, ok
+	session := s.sessions[roundId]
+	return session, nil
 }
-func (s *treeSigningSessionsStore) Delete(roundId string) {
+func (s *treeSigningSessionsStore) Delete(_ context.Context, roundId string) error {
```
sekulicd, line 54: ### Goroutine channel pattern needs refactoring

changes to handling panics/timeout added in prev PR were not perfect imo.

  The current pattern uses `recover()` to catch panics from writing to closed channels:

  ```go
  func() {
      defer func() {
          if r := recover(); r != nil {
              log.Warnf("watchNoncesCollected:recovered from panic: %v", r)
          }
      }()
      select {
      case ch <- struct{}{}:
      case <-ctx.Done():
      }
  }()

  Problems:
  1. recover() masks a real race condition instead of fixing it
  2. Race exists because Delete() closes channels while watchers might be sending
  3. Complex nested selects add unnecessary cognitive load

  Recommended fix - Watcher Ownership Pattern:

  1. Watchers own the channels - only they send and close via defer close(ch)
  2. Delete() never closes channels - just cancels context
  3. Channels passed as arguments to goroutines (not read from struct fields)

  func (s *treeSigningSessionsStore) watchNoncesCollected(
      ctx context.Context, roundId string, ch chan struct{},
  ) {
      defer close(ch)  // watcher owns the channel

      ticker := time.NewTicker(s.pollInterval)
      defer ticker.Stop()

      for {
          select {
          case <-ctx.Done():
              return  // defer close(ch) runs
          case <-ticker.C:
              if s.checkNoncesReady(ctx, roundId) {
                  select {
                  case ch <- struct{}{}:
                  case <-ctx.Done():
                  }
                  return  // defer close(ch) runs
              }
          }
      }
  }

  func (s *treeSigningSessionsStore) Delete(ctx context.Context, roundId string) error {
      // ... redis cleanup ...

      s.lock.Lock()
      defer s.lock.Unlock()

      if s.currentRoundID == roundId && s.cancelRound != nil {
          s.cancelRound()  // watchers exit via ctx.Done(), close their channels
          s.cancelRound = nil
          s.currentRoundID = ""
          s.nonceCh = nil  // just nil, don't close
          s.sigsCh = nil
      }
      return nil
  }

  This eliminates the race entirely - no recover() needed.
  This is applicable to watchSigsCollected, watchNoncesCollected, watchSessionCompletion

internal/core/application/service.go
altafan: done

internal/core/application/ban.go
altafan: done

internal/infrastructure/live-store/inmemory/tree_signing_session.go
```diff
@@ -26,30 +26,32 @@ func NewTreeSigningSessionsStore() ports.TreeSigningSessionsStore {
...
 
-	sess := &ports.MusigSigningSession{
+	session := &ports.MusigSigningSession{
 		Cosigners:   uniqueSignersPubKeys,
 		NbCosigners: len(uniqueSignersPubKeys) + 1, // operator included
 		Nonces:      make(map[string]tree.TreeNonces),
 		Signatures:  make(map[string]tree.TreePartialSigs),
 	}
-	s.sessions[roundId] = sess
+	s.sessions[roundId] = session
```
sekulicd, line 40: ### Single-round model

Multiple rounds support in some tree_signing_session.go were added in prev PR as well
nonceChs     map[string]chan struct{}
sigsChs      map[string]chan struct{}
currently we agreed we have only one round and there are no parallel rounds for processing, also one server no cluster and we should simplify and be consistent like other redis live store impls that have only one round.
	

  Current struct uses maps for channels:
  ```go
  type treeSigningSessionsStore struct {
      nonceChs map[string]chan struct{}
      sigsChs  map[string]chan struct{}
      ctxs     map[string]context.CancelFunc
  }

  Since there's only ever one active round at a time (confirmed by startRound() always calling Delete(existingRound.Id) first), these maps add unnecessary complexity.

  Simplify to:
  type treeSigningSessionsStore struct {
      rdb *redis.Client
      mu  sync.RWMutex

      // single active round (not maps)
      currentRoundID string
      nonceCh        chan struct{}
      sigsCh         chan struct{}
      cancelRound    context.CancelFunc

      pollInterval time.Duration
      numOfRetries int
      retryDelay   time.Duration
  }

  Benefits:
  - Simpler code, fewer allocations
  - No map cleanup needed
  - Matches actual usage pattern
  - Easier to reason about lifecycle

internal/core/application/service.go
```diff
@@ -1086,10 +1104,18 @@ func (s *service) FinalizeOffchainTx(
 ) (structErr errors.Error) {
 	var changes []domain.Event
 
-	offchainTx, exists := s.cache.OffchainTxs().Get(txid)
-	if !exists {
-		return errors.TX_NOT_FOUND.New("offchain tx %s not found", txid).
-			WithMetadata(errors.TxNotFoundMetadata{Txid: txid})
+	offchainTx, err := s.cache.OffchainTxs().Get(ctx, txid)
```
sekulicd, line 1107: The fallback masks potential bugs. We expect cache to always have the data, then a miss indicates something is wrong. Silently falling back to DB means:

  1. We won't know cache sync is broken
  2. Performance degrades (DB hit instead of cache)
  3. Could indicate data integrity issues

internal/core/application/service.go
```diff
@@ -2222,21 +2268,45 @@ func (s *service) startRound() {
 	}
 
 	ctx := context.Background()
-	existingRound := s.cache.CurrentRound().Get(ctx)
+	existingRound, err := s.cache.CurrentRound().Get(ctx)
```
sekulicd, line 2270: ### Cleanup not resilient to cache failures

  In `startRound()`, if `CurrentRound().Get()` fails, the function returns early and **no cleanup happens**:

  ```go
  existingRound, err := s.cache.CurrentRound().Get(ctx)
  if err != nil {
      log.WithError(err).Error("failed to get current round from cache")
      return  // NO CLEANUP!
  }

  If Redis is temporarily unavailable after a round failure:
  - Old ForfeitTxs, ConfirmationSessions, TreeSigingSessions remain
  - New round is never created (server stuck)

  Fix: Move unconditional cleanup before the Get(), and don't return early.
  ```

internal/core/application/service.go
altafan: we'll do in another pr

internal/core/application/service.go
altafan: totally agree, we'll do in another pr or this will get too messy

internal/infrastructure/live-store/inmemory/tree_signing_session.go
altafan: this is in-memory cache, we'll add support once we are ready to support multiple batches

sekulicd: yeah but if we keep this we introduce inconsistency since in other stores we only support one round

internal/core/application/service.go
altafan: if this happens, it means we can't access the cache and can't do anything, not even cleaning it up, so we just return and stop any operation here, in an another pr we improve the handling of this scenario

internal/core/application/service.go
altafan: done

internal/infrastructure/live-store/inmemory/tree_signing_session.go
altafan: done for confirmation session, more tricky for tree signing session because we use a map of channels. It's ok anyway because the lock guarantees atomicity, we'll optimize in a later pr, so we can get rid of the recover

internal/core/application/service.go
altafan: It's actually ok. We need to store the offchain tx in the cache because clients may call FinalizeTx before the db is updated, otherwise we could rely fully on db. If we don't find the tx in the cache, nor in the db we return an error. If data is not found in cache maybe we deleted it manually, this is intentional, i don't see drawbacks.

PR #838, opened 2025-11-27
Wait for confirmation before scheduling sweep task

### LINKED ISSUE #837: Sweep scheduler should wait for commitment tx confirmation
Batch output are swept via CSV lock time, and the scheduler targets the "expired" block as soon as the commitment transaction is broadcasted. Most of the time, the tx is confirmed in the next block. But if it takes time to confirm, the sweep tx will fail with "BIP68-not-final", causing a loop of retry until the CSV is valid:

https://github.com/ark-network/ark/blob/10dde8e23425cdac379362089f89438c6890cb38/internal/core/application/sweeper.go#L578-L585

while scheduling a sweep task, the sweeper should first wait for confirmation and then compute the right time depending on the block mediantimestamp.

louisinger:
Reuse the utils function `waitForConfirmation` used by checkpoint and `fraud.go` to wait for root input to be confirmed before scheduling the batch output sweep.

Also get the vtxo tree expiry from the ark psbt fields, moving expiration time computation outside of application service.

it closes #837 

@altafan please review

## Summary by CodeRabbit

* **Refactor**
  * Unified on-chain confirmation reporting to a single timestamp object and tightened cancellation/error handling.
  * Simplified startup cleanup to rely on identifiers for safer resource deletion.
  * Removed legacy scheduler helper from the public API.

* **New Features / Improvements**
  * Sweep scheduling now waits for root confirmations, derives expirations from transaction metadata, and schedules subtree sweeps concurrently with improved logging and fallback timing.

* **Tests / Chores**
  * Updated mocks and wallet client returns to use the unified timestamp structure.

### DIFF
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -3441,12 +3441,11 @@ func (s *service) scheduleSweepBatchOutput(round *domain.Round) {
 		return
 	}
 
-	rootInput := round.VtxoTree.RootTxid()
-	blockTimestamp, err := waitForConfirmation(context.Background(), rootInput, s.wallet)
+	blockTimestamp, err := waitForConfirmation(context.Background(), round.CommitmentTxid, s.wallet)
 	if err != nil {
 		log.WithError(err).Warnf(
-			"failed to wait for confirmation of batch input tx %s, schedule task time may be inaccurate",
-			rootInput,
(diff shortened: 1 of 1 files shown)


altafan:
@louisinger solve conflicts

PR #854, opened 2025-12-04
Watch batches instead of vtxos pkscript

### LINKED ISSUE #851: Watch batch output scripts instead of vtxo scripts
Every time a batch is finalized, we start watching for all vtxo leaves, and we also do the same whenever an offchain tx is finalized.
The purpose is to ensure we detect unrolls, and to protect against fraud, but subscribing to the entire vtxo set is just a bit overkill.

We can instead go for an "incremental watch"  approach here:
* whenever a batch is finalized, we start watching for its output script until it shows up onchain, and we stop the watch as soon as it happens
* if we swept, nothing left to do; otherwise, we start watching their children

By repeating these 2 steps we end up watching only the vtxos that are actually being unrolled instead of the entire trees and virtual mempool.

louisinger:
This PR reworks the watching logic. Instead of subscribing to any vtxo pkscript, arkd watches only batch output script. If unrolled, the children will be watched.

It reduces drastically the number of scripts watched by nbxplorer.

it closes #851 

@altafan @Kukks please review

## Summary by CodeRabbit

* **New Features**
  * Added an API/RPC to unwatch all tracked scripts in a single operation.

* **Improvements**
  * Switched script-watching to derive and resume watches from root transactions.
  * Added bulk unwatch support in scanner/backends to clear all tracked addresses.
  * Added child-transaction lookup used to better traverse transaction trees.

* **Tests**
  * Expanded end-to-end tests covering fraud handling across node restarts.


internal/core/application/watcher.go
```diff
@@ -0,0 +1,189 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to parse tx: %s", err)
+		}
+		for _, out := range ptx.UnsignedTx.TxOut {
+			if bytes.Equal(out.PkScript, txutils.ANCHOR_PKSCRIPT) {
+				continue
+			}
+			scripts = append(scripts, hex.EncodeToString(out.PkScript))
+		}
+	}
```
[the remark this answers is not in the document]
louisinger: no, batch output can't have subdust scripts

internal/core/application/service.go
```diff
@@ -3492,52 +3410,74 @@ func (s *service) getSpentVtxos(intents map[string]domain.Intent) []domain.Vtxo
...
 
-	commitmentTxIds, err := s.repoManager.Rounds().GetSweepableRounds(ctx)
-	if err != nil {
-		return err
-	}
+		// if there are sweepable outputs, we need to watch the children batch scripts
+		if len(sweepableOutputs) > 0 {
+			for _, outputs := range sweepableOutputs {
+				for _, output := range outputs {
+					childrenTxs, err := s.repoManager.Rounds().
```
altafan, line 3573: why do we have to source them from the repo if we already have the whole vtxo tree here (L3430)?

internal/core/application/service.go
```diff
@@ -3560,53 +3500,6 @@ func (s *service) restoreWatchingVtxos() error {
 	return nil
```
altafan, line 3621: fix the log at L3359
```
log.Debugf("restored watching %d scripts", len(scripts))
```

internal/core/application/watcher.go
```diff
@@ -0,0 +1,189 @@
...
+				return
+			}
+
+			ptx, err := psbt.NewFromRawBytes(strings.NewReader(txs[0]), true)
+			if err != nil {
+				log.WithError(err).Warn("failed to parse tx, skipping...")
+				return
+			}
+
+			// remove sweeper task for the associated checkpoint outputs
```
altafan: I would change the comment to something more meaningful, probably better to move it between L136 and L137
```suggestion
			// An ark tx hit the chain, the related checkpoint txs were previously scheduled for the sweep and we can now remove those tasks
```

internal/core/application/watcher.go
```diff
@@ -0,0 +1,189 @@
...
+			for _, in := range ptx.UnsignedTx.TxIn {
+				taskId := in.PreviousOutPoint.Hash.String()
+				s.sweeper.removeTask(taskId)
+				log.Debugf("sweeper: unscheduled task for tx %s", taskId)
```
altafan: let's move this log inside `removeTask`

pkg/arkd-wallet/core/infrastructure/nbxplorer/service.go
```diff
@@ -631,6 +631,21 @@ func (n *nbxplorer) UnwatchAddresses(ctx context.Context, addresses ...string) e
 	return nil
 }
 
+// UnwatchAllAddresses removes all addresses from group via DELETE /v1/groups/{groupID}/children/delete endpoint.
+func (n *nbxplorer) UnwatchAllAddresses(ctx context.Context) error {
```
altafan, line 635: @Kukks can you double-check the endpoints used here to watch/unwatch?

internal/test/e2e/e2e_test.go
```diff
@@ -2362,6 +2362,322 @@ func TestReactToFraud(t *testing.T) {
 			require.NoError(t, err)
 			require.NotContains(t, aliceVtxos, vtxoToFraud)
 		})
+
+		t.Run("cltv vtxo script with arkd restart", func(t *testing.T) {
```
altafan: replace with a copycat of `default vtxo script` and name it:
```suggestion
		t.Run("after arkd restart", func(t *testing.T) {
```

internal/core/application/service.go
louisinger: GetChildrenTxs also handle the case the leaf is onchain (so children is not in the tree).
btw this logic is a "workaround" and is not efficient. It should be replaced by the proper "onchain" state of batches persisted in DB

PR #864, opened 2025-12-11
move the start sub-command to cobra.

### LINKED ISSUE #782: Improve main action flag visibility
Due to the current commands layout it is impossible to print the configuration options within the CLI, instead the operator needs to refer to the current [documentation](https://github.com/arkade-os/arkd?tab=readme-ov-file#configuration-options) or the code itself. More over this requires coordinating code changes with docs update, which can lead to inconsistencies[1]
This is because the `mainAction` flags are inherited by all sub-commands.

In order to improve this, the current `mainAction` could be moved to the `serve` (or alike) sub-command so that it can define its own flags, that way running `arkd serve --help` should print all the flag usage.
To not break anyone's flow, we could rely on `app.DefaultAction = serve` to keep the current behavior.

1: `SETTLEMENT_MIN_EXPIRY_GAP` flag is defined in code but missing in the docs.

qustavo:
This is the first (of multiple) commit that starts the migration from urfave/cli to spf13/cobra so that we have a unified commmand and flag handler.

Now flag usage is documented.

Closes #782

## Summary by CodeRabbit

* **New Features**
  * Added CLI flag-based configuration and wiring for server startup.
  * Environment variables are now bound with an ARKD_ prefix and support hyphen-to-underscore mapping for keys (improved env var/flag consistency).

* **Chores**
  * Reworked command wiring and startup handling to route execution through a wrapper-based CLI flow.


altafan:
@qustavo sorry for the late reply  🙏. Can we go directly for the migration from urfave/cli to cobra? This way we don't need to have "wrappers"

qustavo:
> @qustavo sorry for the late reply 🙏. Can we go directly for the migration from urfave/cli to cobra? This way we don't need to have "wrappers"

We could but that would turn the PR into a greater change which makes it harder to properly review it. To better understand your motivation, what's the downside of a wrapper (which will eventually be removed)

qustavo:
It might be more code changes, split in smaller PRs, IMO these are easier to reason about hence to review, also it should minimize branch conflicts.

Issue #866, opened 2025-12-15
Admin apis for fees


altafan:
At the moment, restarting arkd is required to update the fee policies as they can be changed only via env vars.

Let's get rid of the env vars in favor of admin apis so that fee polices can be updated at runtime and prevent any service restart.

This task requires changes at pretty much all layers:
- domain: add `IntentFee` domain and `FeeRepository` interface with CRUD apis:
```
type IntentFee struct {
   OnchainInputFee string
   OnchainInputFee string

}
```
- ports: add `Fees() domain.FeeRepository` api to repo manager
- infrastructure: add implementations of `domain.FeeRepository` (badger, sqlite and postgres)
- application: add CRUD apis to admin service for fee management
- interface: add new rpc `UpdateIntentFees` to admin.proto

altafan:
@bitcoin-coder-bob in `internal/infrastructure/db` you can find `badger`, `sqlite` and postgres` folders that include the implementations of the repository interfaces defined in `internal/core/domain`.

The idea is to add a new file to those folders containing the impl of the newly interface added to domain.

Your definition of the table looks fine for pg and sqlite. Badger is a kv value store so there's no need for a table (just start from a copy paste of scheduled session repo if you're not confident with it)

bitcoin-coder-bob:
@Kukks can we close this?

PR #860, opened 2025-12-05
Introduce fee system using CEL formulas

louisinger:
_This PR is introducing the fee system specified by #639_

`ark-lib/arkfee` package is exporting all the structs and types parsing and evaluating CEL programs.

Add new env variables letting to specify the fee model for intent input `ARKD_INTENT_INPUT_FEE_PROGRAM` , and intent output : `ARKD_INTENT_OUTPUT_FEE_PROGRAM`. 

the CEL envs are speicifed in [arkfee/README.md](https://github.com/arkade-os/arkd/tree/arkfee/pkg/ark-lib/arkfee
)
Example of valid CEL programs: 
```
inputType == 'note' || inputType == 'recoverable' ? 0.0 : amount*0.01 

outputType == 'onchain' ? 200.0 : 0.0

expiry - now() < double(duration('5m').getSeconds()) ? 0.0 : amount / 2.0

outputType == 'onchain' ? amount * 0.2 : 0.0
```

@Kukks @altafan please review

it closes #857 
it closes #856 

## Summary by CodeRabbit

* **New Features**
  * Added EstimateFee API/RPC to request dynamic fee estimates.
  * Introduced configurable fee estimator with CEL-based programs for offchain/onchain inputs and outputs.

* **Breaking Changes**
  * Replaced single OnchainOutputFee with four program-based fee config fields.
  * Removed includeOffchain/includeOnchain flags from SDK Balance and CollaborativeExit calls.

* **Documentation**
  * Added arkfee README with usage examples.

* **Tests**
  * Added estimator unit tests and an end-to-end fee test.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #868, opened 2025-12-16
Admin fee api

bitcoin-coder-bob:
Issue #866 and a follow up to https://github.com/arkade-os/arkd/pull/860

This PR adds support for dynamically updating intent fees via an admin api endpoint, thus not requiring an arkd restart to change the itent fee env vars. These env vars are now removed entirely, and management of the intent fees can be performed via new admin api endpoints.

The new admin endpoints:

`GET /v1/admin/intentFees` will pull the latest intent fee programs

`POST /v1/admin/intentFees` will update the intent fee programs. Partial updates are allowed, thus retaining the intent fee programs for any unset fields.

`POST /v1/admin/clearIntentFees` will set all fee programs to `0.0`.

A new repo manager called `FeeRepository` was created and added to the repo manager.

The `FeeManager` port had `GetIntentFees` renamed to `GetFeesFromIntent` to remove confusion with the `GetIntentFees` used by the `FeeRespository`.

The data stores (sqlite, postgres, badger) now store the intent fee programs. A default row with all intent fees programs set to `0.0` is added on startup. This allows us to also track changes to intent fees over time. A possible TODO would be to reference this table in other tables that use intents, so we can keep track of what intent fees were used at that time.

## Summary by CodeRabbit

* **New Features**
  * Admin API and CLI to get, update, and clear intent fees at runtime; fees persisted to the database.
* **Refactor**
  * Replaced environment-based fee configuration with runtime, repository-backed fee management and validation.
* **Tests**
  * End-to-end tests updated to use the admin fee API for updates.
* **Documentation**
  * Library examples updated to reflect new fee handling.

### DIFF
--- a/cmd/arkd/commands.go
+++ b/cmd/arkd/commands.go
@@ -1238,7 +1238,10 @@ func updateIntentFees(ctx *cli.Context) error {
 	url := fmt.Sprintf("%s/v1/admin/intentFees", baseURL)
 	body := fmt.Sprintf(
 		`{"fees":{"onchain_input_fee": "%s", "offchain_input_fee": "%s", "onchain_output_fee": "%s", "offchain_output_fee": "%s"}}`,
-		onchainInputFee, offchainInputFee, onchainOutputFee, offchainOutputFee,
+		onchainInputFee,
+		offchainInputFee,
+		onchainOutputFee,
+		offchainOutputFee,
 	)
 
 	if _, err := post[struct{}](url, body, "", macaroon, tlsConfig); err != nil {


internal/config/config.go
```diff
@@ -247,14 +243,18 @@ var (
...
+	defaultOtelPushInterval               = 10 // seconds
+	defaultHeartbeatInterval              = 60 // seconds
+	defaultRoundReportServiceEnabled      = false
+	defaultSettlementMinExpiryGap         = 0 // disabled by default
+	defaultVtxoNoCsvValidationCutoffDate  = 0 // disabled by default
+	defaultEnablePprof                    = false
+	defaultIntentOffchainInputFeeProgram  = "0.0"
+	defaultIntentOnchainInputFeeProgram   = "0.0"
+	defaultIntentOffchainOutputFeeProgram = "0.0"
+	defaultIntentOnchainOutputFeeProgram  = "0.0"
```
altafan: let's drop also these

internal/config/config.go
```diff
@@ -405,10 +405,10 @@ func LoadConfig() (*Config, error) {
...
 		VtxoNoCsvValidationCutoffDate: viper.GetInt64(VtxoNoCsvValidationCutoffDate),
 		EnablePprof:                   viper.GetBool(EnablePprof),
-		IntentOffchainInputProgram:    viper.GetString(IntentOffchainInputFeeProgram),
-		IntentOnchainInputProgram:     viper.GetString(IntentOnchainInputFeeProgram),
-		IntentOffchainOutputProgram:   viper.GetString(IntentOffchainOutputFeeProgram),
-		IntentOnchainOutputProgram:    viper.GetString(IntentOnchainOutputFeeProgram),
+		IntentOffchainInputProgram:    defaultIntentOffchainInputFeeProgram,
+		IntentOnchainInputProgram:     defaultIntentOnchainInputFeeProgram,
+		IntentOffchainOutputProgram:   defaultIntentOffchainOutputFeeProgram,
+		IntentOnchainOutputProgram:    defaultIntentOnchainOutputFeeProgram,
```
altafan: and these... the idea is to fully drop env var configuration for fees

bitcoin-coder-bob: I can drop them but we do want some sort of initial value set for the fee programs. Deleting these will make the default set to an empty string which is not a valid program. If we do not set a default then arkd is responsible for ensuring to make an update to these values on startup by calling to update the intent fees

internal/core/application/admin.go
```diff
@@ -374,6 +377,44 @@ func (s *adminService) DeleteIntents(ctx context.Context, intentIds ...string) e
...
+		if fees.OnchainInputFee == "" {
+			fees.OnchainInputFee = existingFees.OnchainInputFee
+		}
+		if fees.OffchainOutputFee == "" {
+			fees.OffchainOutputFee = existingFees.OffchainOutputFee
+		}
+		if fees.OnchainOutputFee == "" {
+			fees.OnchainOutputFee = existingFees.OnchainOutputFee
+		}
+	}
```
altafan: we could move this logic into the repo impls.. in sql we can handle this directly with the update query and spare the one to get intent fees

internal/core/ports/fee.go
```diff
@@ -8,9 +8,12 @@ import (
...
 type FeeManager interface {
-	GetIntentFees(
+	GetFeesFromIntent(
 		ctx context.Context,
 		boardingInputs []wire.TxOut, vtxoInputs []domain.Vtxo,
 		onchainOutputs, offchainOutputs []wire.TxOut,
 	) (int64, error)
+	GetIntentFees(ctx context.Context) (*domain.IntentFees, error)
+	UpsertIntentFees(ctx context.Context, fees domain.IntentFees) error
+	ClearIntentFees(ctx context.Context) error
```
altafan: the feeManager should take care of calculating the intent fees based on the given onchain/offchain inputs and outputs and the formulas stored in db.

I'd revert the changes to GetIntentFees so we keep it to compute the intent fee amount, and we add:
```suggestion
	UpdateIntentFormulas(ctx context.Context, fees domain.IntentFees) error
	ClearIntentFormulas(ctx context.Context) error
```

that should be invoked by the admin service in the application layer

internal/infrastructure/db/sqlite/intent_fees_repo.go
```diff
@@ -0,0 +1,81 @@
...
+func (r *intentFeesRepo) Close() {
+	// nolint:all
+	r.db.Close()
+}
+
+func (r *intentFeesRepo) GetIntentFees(ctx context.Context) (*domain.IntentFees, error) {
+	intentFees, err := r.querier.SelectLatestIntentFees(ctx)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return &domain.IntentFees{}, nil
```
altafan, line 43:
```suggestion
			return nil, nil
```

internal/config/config.go
altafan: in ref to https://github.com/arkade-os/arkd/pull/868#discussion_r2627454140, the idea is to init here in config a fee manager sourcing the formulas from the db, that is shared between app and admin services.

While admin would take care of invoking Update/ClearIntentFormulas (write ops), app would invoke only GetIntentFees (read op). We may need to add a lock to prevent concurrent access between read and write ops

pkg/ark-lib/arkfee/estimator.go
```diff
@@ -33,37 +33,36 @@ func (p *program) Eval(args map[string]any) (FeeAmount, error) {
 }
 
 type Estimator struct {
-	Config               Config
-	intentOffchainInput  *program
-	intentOnchainInput   *program
-	intentOffchainOutput *program
-	intentOnchainOutput  *program
```
louisinger, line 40: if not necessary let's keep them unexported

bitcoin-coder-bob:
@coderabbitai review

internal/config/config.go
```diff
@@ -849,12 +844,7 @@ func (c *Config) appService() error {
 		c.ScheduledSessionMinRoundParticipantsCount, c.ScheduledSessionMaxRoundParticipantsCount,
 		c.SettlementMinExpiryGap,
 		time.Unix(c.VtxoNoCsvValidationCutoffDate, 0),
-		application.IntentFeeInfo{
-			OffchainInput:  c.IntentOffchainInputProgram,
-			OnchainInput:   c.IntentOnchainInputProgram,
-			OffchainOutput: c.IntentOffchainOutputProgram,
-			OnchainOutput:  c.IntentOnchainOutputProgram,
-		},
+		intentFees,
```
altafan: not needed? can be dropped?

altafan: (and we spare the query to the db this way)

internal/core/ports/fee.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 type FeeManager interface {
-	GetIntentFees(
+	GetFeesFromIntent(
```
altafan: let's revert? or eventually, if you don't like the overlap with the repo api, let's rename it `ComputeIntentFees()`?

bitcoin-coder-bob: this PR introduces `GetIntentFees` which calls the underlying data store. I'd prefer to rename here to `ComputeIntentFees` so that we do not have any confusion

internal/infrastructure/db/badger/intent_fees_repo.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if err := r.store.Upsert("intent_fees", &intentFees); err != nil {
+		return fmt.Errorf("failed to clear intent fees: %w", err)
+	}
+
+	return nil
+}
+
+type IntentFees struct {
```
altafan: no need to export this struct, can be renamed to 
```suggestion
type intentFeesDTO struct {
```

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
```diff
@@ -0,0 +1,16 @@
...
+    offchain_input_fee_program TEXT NOT NULL DEFAULT '0.0',
+    onchain_input_fee_program TEXT NOT NULL DEFAULT '0.0',
+    offchain_output_fee_program TEXT NOT NULL DEFAULT '0.0',
+    onchain_output_fee_program TEXT NOT NULL DEFAULT '0.0'
+);
+
+-- add a row with default values if not exists
+INSERT INTO intent_fees (id, created_at)
+SELECT '00000000-0000-0000-0000-000000000001', EXTRACT(EPOCH FROM NOW())::BIGINT
+WHERE NOT EXISTS (SELECT 1 FROM intent_fees WHERE id = '00000000-0000-0000-0000-000000000001');
```
altafan: can you explain this?

internal/infrastructure/db/postgres/sqlc/query.sql
```diff
@@ -354,3 +354,27 @@ ORDER BY created_at ASC;
 SELECT * FROM conviction 
 WHERE crime_round_id = @round_id
 ORDER BY created_at ASC;
+
+
+-- name: SelectIntentFees :one
+SELECT * FROM intent_fees WHERE id = @id;
```
altafan: not used? can be dropped?

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
bitcoin-coder-bob: the idea is that we never want this db to be empty, we want there to be 1 row in there to start with zero value (`"0.0"`) fee programs. This code snippet will add that row (it uses the `DEFAULT` values to set the zero values for the fee programs) for us on db creation, if the row does not already exist

internal/infrastructure/db/service_test.go
```diff
@@ -1285,6 +1286,101 @@ func testConvictionRepository(t *testing.T, svc ports.RepoManager) {
...
+		err = repo.UpdateIntentFees(ctx, newFees)
+		require.NoError(t, err)
+
+		updatedFees, err = repo.GetIntentFees(ctx)
+		require.NoError(t, err)
+		require.NotNil(t, updatedFees)
+		require.Equal(t, newFees.OnchainInputFee, updatedFees.OnchainInputFee)
+		require.Equal(t, newFees.OffchainInputFee, updatedFees.OffchainInputFee)
+		require.Equal(t, newFees.OnchainOutputFee, updatedFees.OnchainOutputFee)
+		require.Equal(t, newFees.OffchainOutputFee, updatedFees.OffchainOutputFee)
```
altafan, line 1473: this is basically repeating the prev code, that we already tested. What if, instead, here we check sometihng similar to the case below?

with an "empty" db we should be able to update the fees partially too

internal/infrastructure/feemanager/arkfee.go
```diff
@@ -49,14 +61,34 @@ func (a arkFeeManager) GetIntentFees(
...
+// update intent fees programs, will only update intent fee programs that are non-empty
+func (a *arkFeeManager) UpdateIntentFees(ctx context.Context, fees domain.IntentFees) error {
+	return a.repo.UpdateIntentFees(ctx, fees)
+}
+
+// resets intent fees to zero-fee programs
+func (a *arkFeeManager) ClearIntentFees(ctx context.Context) error {
+	return a.repo.ClearIntentFees(ctx)
+}
+
```
altafan: i guess this is a left over, can be dropped

internal/infrastructure/db/service_test.go
bitcoin-coder-bob: lines 1339 to 1348 test out doing a partial updates, which is slightly different than the early test where we update all the fields at once. I can also add doing a partial update on an empty db (where empty db means all values in db are "0.0")

internal/interface/grpc/handlers/adminservice.go
```diff
@@ -449,6 +449,56 @@ func (a *adminHandler) RevokeAuth(
...
+	feesProto := req.GetFees()
+	if feesProto == nil {
+		return nil, status.Error(codes.InvalidArgument, "missing intent fees")
+	}
+
+	fees := domain.IntentFees{
+		OffchainInputFee:  feesProto.OffchainInputFee,
+		OnchainInputFee:   feesProto.OnchainInputFee,
+		OffchainOutputFee: feesProto.OffchainOutputFee,
+		OnchainOutputFee:  feesProto.OnchainOutputFee,
```
altafan: use getters

```suggestion
		OffchainInputFee:  feesProto.GetOffchainInputFee(),
		OnchainInputFee:   feesProto.GetOnchainInputFee(),
		OffchainOutputFee: feesProto.GetOffchainOutputFee(),
		OnchainOutputFee:  feesProto.GetOnchainOutputFee(),
```

internal/interface/grpc/permissions/permissions.go
```diff
@@ -144,6 +144,18 @@ func Whitelist() map[string][]bakery.Op {
...
+			Action: "read",
+		}},
+		fmt.Sprintf("/%s/UpdateIntentFees", arkv1.ArkService_ServiceDesc.ServiceName): {{
+			Entity: EntityArk,
+			Action: "write",
+		}},
+		fmt.Sprintf("/%s/ClearIntentFees", arkv1.ArkService_ServiceDesc.ServiceName): {{
+			Entity: EntityArk,
+			Action: "write",
+		}},
```
altafan: drop from whitelist, we want these endpoints to be eventually protected by auth (macaroon)

internal/infrastructure/db/service_test.go
altafan: yeah, my idea here is to not re-do exactly the things done 'till L1314, rather we can take the occasion to test smtn different, ie a partial update of the fees with empty db (where empty db means all values in db are "0.0"), and we keep also what's from L1351 to test partial update also on non-empty db

internal/infrastructure/feemanager/arkfee.go
```diff
@@ -49,14 +61,34 @@ func (a arkFeeManager) GetIntentFees(
...
 
-	fee, err := a.Eval(offchainInputs, onchainInputs, arkfeeOffchainOutputs, arkfeeOnchainOutputs)
+	fee, err := estimator.Eval(
+		offchainInputs,
+		onchainInputs,
+		arkfeeOffchainOutputs,
+		arkfeeOnchainOutputs,
+	)
 	if err != nil {
 		return 0, err
```
altafan:
```suggestion
		return -1, err
```

api-spec/protobuf/ark/v1/admin.proto
```diff
@@ -55,6 +55,23 @@ service AdminService {
...
+      post: "/v1/admin/intentFees"
+      body: "*"
+    };
+  }
+  rpc ClearIntentFees(ClearIntentFeesRequest) returns (ClearIntentFeesResponse){
+    option (meshapi.gateway.http) = {
+      post: "/v1/admin/intentFees/clear"
+      body: "*"
+    };
+  }
```
louisinger, line 74: Do we really need a clear RPC ? can be an `UpdateIntentFees` with empty strings? Just asking, maybe we need it cc @altafan @Kukks

api-spec/protobuf/ark/v1/admin.proto
```diff
@@ -55,6 +55,23 @@ service AdminService {
       body: "*"
     };
   }
+  rpc GetIntentFees(GetIntentFeesRequest) returns (GetIntentFeesResponse){
```
louisinger, line 58: not sure we need this Admin RPC because we can get the same data from GetInfo ?

internal/core/application/admin.go
```diff
@@ -36,6 +36,9 @@ type AdminService interface {
 	ClearScheduledSessionConfig(ctx context.Context) error
 	ListIntents(ctx context.Context, intentIds ...string) ([]IntentInfo, error)
 	DeleteIntents(ctx context.Context, intentIds ...string) error
+	GetIntentFees(ctx context.Context) (*domain.IntentFees, error)
+	UpdateIntentFees(ctx context.Context, fees domain.IntentFees) error
+	ClearIntentFees(ctx context.Context) error
```
louisinger, line 38: like in proto, do we need clear if update can set to default values ?

internal/core/application/service.go
```diff
@@ -2023,6 +2020,19 @@ func (s *service) GetInfo(ctx context.Context) (*ServiceInfo, errors.Error) {
 		}
 	}
 
+	// TODO: caching?
```
louisinger: in fact, we need to cache the whole info response and update it only when it is mutated ? let's open an issue @altafan?

internal/core/domain/fee.go
```diff
@@ -0,0 +1,17 @@
+package domain
+
+import "context"
+
+type IntentFees struct {
```
louisinger, line 5: nit: it makes sense to have it in domain but looks duplicate of `IntentFeeInfo` defined in application/types.go. let's reuse domain.IntentFees there ? wdyt ?

internal/infrastructure/db/postgres/intent_fees_repo.go
```diff
@@ -0,0 +1,98 @@
...
+	return &domain.IntentFees{
+		OnchainInputFee:   row.OnchainInputFeeProgram,
+		OffchainInputFee:  row.OffchainInputFeeProgram,
+		OnchainOutputFee:  row.OnchainOutputFeeProgram,
+		OffchainOutputFee: row.OffchainOutputFeeProgram,
+	}, nil
+}
+
+func (r *intentFeesRepo) UpdateIntentFees(ctx context.Context, fees domain.IntentFees) error {
+	// determine if any of the fees passed are empty, if so we need to grab existing fees to avoid overwriting
```
louisinger: nit: I think you can implement this logic in the QUERY (I don't know how ^^)

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE IF NOT EXISTS intent_fees (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    created_at BIGINT NOT NULL,
```
louisinger:
```suggestion
    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
this way you don't have to handle created_at at all.

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE IF NOT EXISTS intent_fees (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```
louisinger:
```suggestion
    id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
```
I personnaly prefer incremented integer instead of uuid because then it's easier to read as human

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
bitcoin-coder-bob: i generally do use this way but I was following how the rest of our date fields are stored in the db and we used BIGINT so I assumed that was the preferred method use. Should I change to use `TIMESTAMP` ?

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
bitcoin-coder-bob: we don't need to access the id's of this table, I just needed a primary key. In fact none of our queries against this table manually set the id. If we do choose to make it a BIGINT we have to be cognizant that the id value we set is not already used to prevent conflict. Maybe there is a way to have the SQL auto increment for us?

internal/infrastructure/db/postgres/intent_fees_repo.go
bitcoin-coder-bob: I think something like:

```
INSERT INTO intent_fees (
    created_at, offchain_input_fee_program, onchain_input_fee_program,
    offchain_output_fee_program, onchain_output_fee_program
) VALUES (
    @created_at,
    COALESCE(NULLIF(@offchain_input_fee_program, ''), (SELECT offchain_input_fee_program FROM intent_fees ORDER BY created_at DESC LIMIT 1)),
    COALESCE(NULLIF(@onchain_input_fee_program, ''), (SELECT onchain_input_fee_program FROM intent_fees ORDER BY created_at DESC LIMIT 1)),
    COALESCE(NULLIF(@offchain_output_fee_program, ''), (SELECT offchain_output_fee_program FROM intent_fees ORDER BY created_at DESC LIMIT 1)),
    COALESCE(NULLIF(@onchain_output_fee_program, ''), (SELECT onchain_output_fee_program FROM intent_fees ORDER BY created_at DESC LIMIT 1))
);
```

Is this preferable? I think it could actually result in more db reads, since it does a read for each empty fee program passed in, while the currently implemented code will do 1 call to get them all if any of the fee programs were empty

api-spec/protobuf/ark/v1/admin.proto
bitcoin-coder-bob: it was specified in: https://github.com/arkade-os/arkd/issues/866

I think its a helpful convenience function but you are correct we could remove it and just pass 0.0 to `UpdateIntentFees`.

api-spec/protobuf/ark/v1/admin.proto
bitcoin-coder-bob: it was a specified ask in: https://github.com/arkade-os/arkd/issues/866

internal/infrastructure/db/service_test.go
bitcoin-coder-bob: gotcha so the test flow would be:

empty db -> partial update -> full update -> ClearIntentFees

is this correct?

internal/core/application/admin.go
bitcoin-coder-bob: will defer making this change until we sort out a decision on: https://github.com/arkade-os/arkd/pull/868#discussion_r2632399054

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
louisinger: IMO yes, we probably used bigint in vtxo to stick with unix timestamp but here we're not really reading this data.

Regardless the type, I was more suggesting the default value to avoid having to pass the created at while inserting (can be bigint or timestamp as you prefer)

bitcoin-coder-bob: ok defaults for `created_at` added here: [a02b3cc](https://github.com/arkade-os/arkd/pull/868/commits/a02b3cc6d4410360e73cc8a412f8b98745cc6b7c)

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
louisinger: > Maybe there is a way to have the SQL auto increment for us?

yep `GENERATED ALWAYS AS IDENTITY PRIMARY KEY` handles it 
I didn't test myself, but it comes from postres doc: https://www.postgresql.org/docs/current/ddl-identity-columns.html#DDL-IDENTITY-COLUMNS

internal/core/application/admin.go
```diff
@@ -374,6 +377,24 @@ func (s *adminService) DeleteIntents(ctx context.Context, intentIds ...string) e
...
+	ctx context.Context,
+) (*domain.IntentFees, error) {
+	return s.repoManager.Fees().GetIntentFees(ctx)
+}
+
+func (s *adminService) UpdateIntentFees(
+	ctx context.Context,
+	fees domain.IntentFees,
+) error {
+	return s.repoManager.Fees().UpdateIntentFees(ctx, fees)
```
louisinger, line 396: Before updating, we should validate the new CEL programs thanks to `arkfee.parse` function (we need to export the function and add a validate method to feemanager)

louisinger (changes requested):

internal/core/application/admin.go
```diff
@@ -387,6 +389,25 @@ func (s *adminService) UpdateIntentFees(
...
+
+	_, err = arkfee.Parse(fees.OnchainOutputFee, celenv.IntentOutputEnv)
+	if err != nil {
+		return fmt.Errorf("invalid onchain output fee: %w", err)
+	}
+
+	_, err = arkfee.Parse(fees.OffchainOutputFee, celenv.IntentOutputEnv)
+	if err != nil {
+		return fmt.Errorf("invalid offchain output fee: %w", err)
+	}
```
louisinger: We should move this logic to `ports.FeeManager` (the idea is `application` does not depend on `arkfee` pkg but on `ports` interface)

bitcoin-coder-bob: moved them from here into the 3 `intent_fees_repo.go` files: [4b9c0bb](https://github.com/arkade-os/arkd/pull/868/commits/4b9c0bbbd81ef1d2780b808d58be6d384d0fe63c)

api-spec/protobuf/ark/v1/admin.proto
altafan: let's keep these apis. The abstraction (endpoints) we provide here should affect the apis of the underlying layers.

Let's keep the clear in all levels, and use actually update to 0.0 only in the repo impls.

As this is just impl detail and optimization, we want to keep the distinction for human readability and easier maintenance of the code.

internal/core/application/admin.go
altafan: answered here https://github.com/arkade-os/arkd/pull/868/files#r2640098418

internal/core/application/service.go
altafan: dunno, i think we're just fine like this. What would be the advantage here of duplicating the state?

During the batch session we need the cache to speed up the process, but here, even if this adds a small overhead, is it really a concern?

internal/infrastructure/db/postgres/intent_fees_repo.go
altafan: since we basically not allow empty values to be persisted (we default to 0.0 "at worst"), we could leverage this and do not require any read at all here if in the update query, we set only the non-empty fields.

I'd add a safe check in the repo api to make sure not all fields are empty to prevent having an update query that does nothing (and allow noticing this weird behavior that should never happen)

Guess the same can be done in sqlite

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
altafan: let's go for autoincremental interger this way we can also know how many updates have been done by just checking the id of the latest record if we ever need

internal/infrastructure/db/service_test.go
altafan: yeah, this is the flow starting from L1339, what's written before is ok and doesn't need to be touched

altafan: but i can see you added this below. We're fine

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
altafan: this way we can drop L10-13

internal/infrastructure/db/badger/intent_fees_repo.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, fmt.Errorf("no intent fees found")
+	}
+
+	latestIdx := 0
+	for i := 1; i < len(all); i++ {
+		if all[i].CreatedAt > all[latestIdx].CreatedAt {
+			latestIdx = i
+		}
+	}
+	intentFees := all[latestIdx]
```
altafan: we can use `sort.SliceStable`, or `slices.SortStableFunc` to order the records by creation time and just get the first elem, but probably we can do this directly with a badgerhold query so we get the exact result instead of implementing it in pure go

internal/infrastructure/db/badger/intent_fees_repo.go
```diff
@@ -0,0 +1,182 @@
...
+		OnchainInputFee:   intentFees.OnchainInputFeeProgram,
+		OffchainInputFee:  intentFees.OffchainInputFeeProgram,
+		OnchainOutputFee:  intentFees.OnchainOutputFeeProgram,
+		OffchainOutputFee: intentFees.OffchainOutputFeeProgram,
+	}, nil
+}
+
+func (r *intentFeesRepo) UpdateIntentFees(ctx context.Context, fees domain.IntentFees) error {
+	var newEntry intentFeesDTO
+	if fees.OnchainInputFee == "" || fees.OffchainInputFee == "" || fees.OnchainOutputFee == "" ||
```
altafan, line 83: like i said, let's add a safe check here to prevent an empty domain is passed:

```
emptyFees := domain.IntentFees{}
if fees == emptyFees {
    return fmt.Errorf("missing fees to update")
}
```

altafan: let's do this in all repo impls and also add something in tests to make sure the call fails in case the fees are empty

internal/infrastructure/db/postgres/sqlc/query.sql
```diff
@@ -354,3 +354,23 @@ ORDER BY created_at ASC;
...
+SELECT * FROM intent_fees ORDER BY created_at DESC LIMIT 1;
+
+-- name: AddIntentFees :exec
+INSERT INTO intent_fees (
+    offchain_input_fee_program, onchain_input_fee_program,
+    offchain_output_fee_program, onchain_output_fee_program
+) VALUES (
+    @offchain_input_fee_program, @onchain_input_fee_program,
+    @offchain_output_fee_program, @onchain_output_fee_program
+);
```
altafan: https://github.com/arkade-os/arkd/pull/868/files#r2640130996

pkg/ark-lib/arkfee/estimator.go
```diff
@@ -42,28 +41,28 @@ type Estimator struct {
 
 // New parses the intent input and output programs if not empty and returns a new Estimator
 func New(config Config) (estimator *Estimator, err error) {
-	estimator = &Estimator{Config: config}
+	estimator = &Estimator{}
```
altafan, line 44: The changes to this file are not needed. Let's revert them

bitcoin-coder-bob: the `Parse` function I modified to be exported so it could be used in the `intent_fes_repo.go` files so we can validate the programs before insertion into the db. Should I keep this exported? https://github.com/arkade-os/arkd/pull/868#discussion_r2634054007

bitcoin-coder-bob: do we need the config to be a field in the `Estimator` if we only use it in the `New` function?

internal/infrastructure/db/postgres/intent_fees_repo.go
bitcoin-coder-bob: if we pass an empty field (but not the case where all 4 are empty) we want to ensure to use the latest set value for that field which requires the lookup that I have nested inside the query, otherwise it would default to `0.0` which may differ from the actual latest value stored. In order to write a new row where we want to update some subset of the values, a lookup is required for the unset values in order to use them in the new row.

I will add the safe check you mentioned about ensuring all fields are not empty

internal/infrastructure/db/badger/intent_fees_repo.go
bitcoin-coder-bob: added here: [2c08db7](https://github.com/arkade-os/arkd/pull/868/commits/2c08db7d899a6b4dee3a40c8c43035973e75a134)

internal/infrastructure/db/badger/intent_fees_repo.go
bitcoin-coder-bob: added here: [2c08db7](https://github.com/arkade-os/arkd/pull/868/commits/2c08db7d899a6b4dee3a40c8c43035973e75a134)

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
bitcoin-coder-bob: added here: [2c08db7](https://github.com/arkade-os/arkd/pull/868/commits/2c08db7d899a6b4dee3a40c8c43035973e75a134)

internal/infrastructure/db/postgres/intent_fees_repo.go
bitcoin-coder-bob: added the query here: [2c08db7](https://github.com/arkade-os/arkd/pull/868/commits/2c08db7d899a6b4dee3a40c8c43035973e75a134)

api-spec/protobuf/ark/v1/admin.proto
bitcoin-coder-bob: added here: [2c08db7](https://github.com/arkade-os/arkd/pull/868/commits/2c08db7d899a6b4dee3a40c8c43035973e75a134)

internal/core/application/admin.go
altafan: i missed this from my prev review. The repos are not the right place where to move this code. They should be agnostic from the fee manager impl, otherwise we're adding "celenv fee repositories" in this pr, ie any db that supports persisting exclusively cel formulas. If one day we want to use something else, it would be required totouch the repo impls and this is not what we really want.

The right place where to move the code is the cel impl of the fee manager (in infra/), but it means "reverting" to a prev version where we exposed the `domain.FeeRepository` apis into `ports.FeeManager`, and have the latter shared between admin and app services. The admin service would interact with the fee mananger to get/update the fees instead of calling directly the repo apis.

This way, we could perform these checks into `feeManager.UpdateIntentFees`, right before calling the underlying repo's api with the same name.

An intermediate solution that wouldn't require a "revert" is to add a `Validate(program any) error` api to ports.FeeManager` and make the admin service call it before calling `repo.UpdateIntentFees` to ensure we persist valid cel formulas.

internal/infrastructure/db/sqlite/sqlc/query.sql
```diff
@@ -383,14 +383,8 @@ INSERT INTO intent_fees (
...
--- name: ClearIntentFees :exec
-INSERT INTO intent_fees (
-    offchain_input_fee_program, onchain_input_fee_program,
-    offchain_output_fee_program, onchain_output_fee_program
-) VALUES (
-    '0.0', '0.0', '0.0', '0.0'
+    COALESCE(NULLIF(@offchain_input_fee_program, ''), (SELECT offchain_input_fee_program FROM intent_fees ORDER BY created_at DESC LIMIT 1)),
+    COALESCE(NULLIF(@onchain_input_fee_program, ''), (SELECT onchain_input_fee_program FROM intent_fees ORDER BY created_at DESC LIMIT 1)),
+    COALESCE(NULLIF(@offchain_output_fee_program, ''), (SELECT offchain_output_fee_program FROM intent_fees ORDER BY created_at DESC LIMIT 1)),
+    COALESCE(NULLIF(@onchain_output_fee_program, ''), (SELECT onchain_output_fee_program FROM intent_fees ORDER BY created_at DESC LIMIT 1))
```
altafan: this query can be optmized with a single select as follows:
```
INSERT INTO intent_fees (
    offchain_input_fee_program, onchain_input_fee_program,
    offchain_output_fee_program, onchain_output_fee_program
)
SELECT
    COALESCE(NULLIF(:offchain_input_fee_program, ''), last.offchain_input_fee_program, ''),
    COALESCE(NULLIF(:onchain_input_fee_program, ''), last.onchain_input_fee_program, ''),
    COALESCE(NULLIF(:offchain_output_fee_program, ''), last.offchain_output_fee_program, ''),
    COALESCE(NULLIF(:onchain_output_fee_program, ''), last.onchain_output_fee_program, '')
FROM (
    SELECT * FROM intent_fees ORDER BY id DESC LIMIT 1
) AS last;
```

altafan:
@louisinger @bitcoin-coder-bob following the reasoning of https://github.com/arkade-os/arkd/pull/868#discussion_r2643138245, I think I realized it's also wrong to store a default "0.0" fee program instead of an empty string in repo impls because that value may make sense only in case we use cel. If, in the future, we need to move to something else, the fee repos impls may need to be modified and this is not something we really want.

Because of this, my proposal to tidy up the abstractions with an eye for future maintenance is:
1. fee repo impls at startup do nothing and keep their db empty.
2. GetIntentFees returns an empty struct if the db is empty instead of a "not found" error
3. We keep UpdateIntentFees almost as is: safe-check to reject empty structs, update a field only if the given value is not an empty string (i know this is an insert in the end, the logic is just easier to explain talking about update, and the code is pretty much validated) without using arkfee.Parse so we can revert all changes made to ark-lib (check 6)
4. ClearIntentFees sets all fields to empty strings and it is basically the only one allowed to add empty values to the db
5. In feeManager, if the repo returns an empty struct, or, in general, if any of the fields are empty strings, we default to "0.0".
6. In feeManager, we also add a `Validate(program any) error` that is implemented by arkfee.parse(), and we call this in admin service before calling repo.UpdateIntentFees

This way, repos are completely agnostic to whatever fee system is used and shouldn't require any change in the future.

And, most importantly, this way something that is implementational details like the impl of ports.FeeManager adapts to the core logic (domanii, app) instead of affecting it.

internal/infrastructure/db/sqlite/sqlc/query.sql
bitcoin-coder-bob: added here: [87b29c7](https://github.com/arkade-os/arkd/pull/868/commits/87b29c77d8cbf1b89951175d7d3614f962f15c27)

internal/core/application/service.go
```diff
@@ -2023,6 +2020,25 @@ func (s *service) GetInfo(ctx context.Context) (*ServiceInfo, errors.Error) {
...
+	}
+	if resolvedFees.OffchainInputFee == "" {
+		resolvedFees.OffchainInputFee = "0.0"
+	}
+	if resolvedFees.OnchainOutputFee == "" {
+		resolvedFees.OnchainOutputFee = "0.0"
+	}
+	if resolvedFees.OffchainOutputFee == "" {
+		resolvedFees.OffchainOutputFee = "0.0"
+	}
```
altafan: not needed, drop. We're ok with returning empty strings

internal/core/application/admin.go
```diff
@@ -374,6 +379,73 @@ func (s *adminService) DeleteIntents(ctx context.Context, intentIds ...string) e
...
+		resolvedFees.OffchainInputFee = "0.0"
+	}
+	if resolvedFees.OnchainOutputFee == "" {
+		resolvedFees.OnchainOutputFee = "0.0"
+	}
+	if resolvedFees.OffchainOutputFee == "" {
+		resolvedFees.OffchainOutputFee = "0.0"
+	}
+
+	return resolvedFees, nil
```
altafan:
```suggestion
	return s.repoManager.Fees().GetIntentFees(ctx)
```

internal/core/application/admin.go
```diff
@@ -374,6 +379,73 @@ func (s *adminService) DeleteIntents(ctx context.Context, intentIds ...string) e
...
+		if err != nil {
+			return fmt.Errorf("invalid onchain output fee program: %w", err)
+		}
+	}
+	if fees.OffchainOutputFee != "" {
+		err = feeManager.Validate(fees.OffchainOutputFee, celenv.IntentOutputEnv)
+		if err != nil {
+			return fmt.Errorf("invalid offchain output fee program: %w", err)
+		}
+	}
```
altafan: Let's add ports.FeeManager to the fields of admin service and passed it in the constructor like done for the app svc, then here we can simplify like:

```suggestion
	// validate the programs for set fields
	if err := s.feeManager.Validate(fees); err != nil {
		return err
	}
```

internal/core/ports/fee.go
```diff
@@ -5,12 +5,14 @@ import (
...
 )
 
 type FeeManager interface {
-	GetIntentFees(
+	ComputeIntentFees(
 		ctx context.Context,
 		boardingInputs []wire.TxOut, vtxoInputs []domain.Vtxo,
 		onchainOutputs, offchainOutputs []wire.TxOut,
 	) (int64, error)
+	Validate(feeProgram string, celEnv *cel.Env) error
```
altafan:
```suggestion
	Validate(fees domain.IntentFees) error
```

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
```diff
@@ -0,0 +1,11 @@
...
+    id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    created_at BIGINT NOT NULL DEFAULT (EXTRACT(EPOCH FROM NOW()) * 1000)::BIGINT,
+    offchain_input_fee_program TEXT NOT NULL DEFAULT '',
+    onchain_input_fee_program TEXT NOT NULL DEFAULT '',
+    offchain_output_fee_program TEXT NOT NULL DEFAULT '',
+    onchain_output_fee_program TEXT NOT NULL DEFAULT ''
+);
+
+-- add index on created_at for faster retrieval of latest fees
+CREATE INDEX IF NOT EXISTS idx_intent_fees_created_at ON intent_fees (created_at);
```
altafan: the index is not needed, we can drop it and fully rely on id now

internal/infrastructure/db/postgres/sqlc/query.sql
```diff
@@ -369,3 +369,38 @@ ORDER BY created_at ASC;
 SELECT * FROM conviction 
 WHERE crime_round_id = @round_id
 ORDER BY created_at ASC;
+
+-- name: SelectLatestIntentFees :one
+SELECT * FROM intent_fees ORDER BY created_at DESC LIMIT 1;
```
altafan:
```suggestion
SELECT * FROM intent_fees ORDER BY id DESC LIMIT 1;
```

internal/infrastructure/db/service_test.go
```diff
@@ -1410,6 +1411,122 @@ func testConvictionRepository(t *testing.T, svc ports.RepoManager) {
...
+		require.NoError(t, err)
+
+		updatedFees, err = repo.GetIntentFees(ctx)
+		require.NoError(t, err)
+		require.NotNil(t, updatedFees)
+		require.Equal(t, newFees.OnchainInputFee, updatedFees.OnchainInputFee)
+		require.Equal(t, newFees.OffchainInputFee, updatedFees.OffchainInputFee)
+		require.Equal(t, "", updatedFees.OnchainOutputFee)
+		require.Equal(t, "", updatedFees.OffchainOutputFee)
+
```
altafan: drop blank line

bitcoin-coder-bob: fixed in: [2096027](https://github.com/arkade-os/arkd/pull/868/commits/2096027c049bc50f1b69366d50d518d3a847829e)

internal/infrastructure/db/postgres/sqlc/query.sql
bitcoin-coder-bob: fixed in: [2096027](https://github.com/arkade-os/arkd/pull/868/commits/2096027c049bc50f1b69366d50d518d3a847829e)

internal/infrastructure/db/postgres/migration/20251215000000_add_intent_fees.up.sql
bitcoin-coder-bob: fixed in: [2096027](https://github.com/arkade-os/arkd/pull/868/commits/2096027c049bc50f1b69366d50d518d3a847829e)

internal/core/ports/fee.go
bitcoin-coder-bob: fixed in: [2096027](https://github.com/arkade-os/arkd/pull/868/commits/2096027c049bc50f1b69366d50d518d3a847829e)

internal/core/application/admin.go
bitcoin-coder-bob: fixed in: [2096027](https://github.com/arkade-os/arkd/pull/868/commits/2096027c049bc50f1b69366d50d518d3a847829e)

internal/core/application/admin.go
bitcoin-coder-bob: fixed in: [2096027](https://github.com/arkade-os/arkd/pull/868/commits/2096027c049bc50f1b69366d50d518d3a847829e)

internal/core/application/service.go
bitcoin-coder-bob: fixed in: [2096027](https://github.com/arkade-os/arkd/pull/868/commits/2096027c049bc50f1b69366d50d518d3a847829e)

internal/infrastructure/db/postgres/intent_fees_repo.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (r *intentFeesRepo) ClearIntentFees(ctx context.Context) error {
+	if err := r.querier.ClearIntentFees(ctx); err != nil {
+		return fmt.Errorf("failed to clear intent fees: %w", err)
+	}
+
+	return nil
+}
+func (r *intentFeesRepo) Close() {
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: fixed in: [6300313](https://github.com/arkade-os/arkd/pull/868/commits/630031307712af45460dc053b3f00b336ad6d1e9)

louisinger (approved):

PR #869, opened 2025-12-17
feat: add GetPendingTx by txid

Chigala:
This PR extends the `GetPendingTx` RPC to allow fetching pending transactions by `txid` in addition to the existing Intent mechanism. It also introduces a dedicated structured error for when a VTXO is locked because of a pending transaction. I followed the same patterns used in [this PR](https://github.com/arkade-os/arkd/pull/832). I also added new test suites to test this [here](https://github.com/Chigala/arkd/blob/d918bfb4e02b75fbc0a1fb58ad0bc19b39b92683/internal/test/e2e/e2e_test.go#L3992).

A good follow-up task to this would be adding a `GetPendingTxByTxid` method to the go-sdk. This would remove the [`getPendingTxByTxid`](https://github.com/Chigala/arkd/blob/d918bfb4e02b75fbc0a1fb58ad0bc19b39b92683/internal/test/e2e/utils_test.go#L795) function I have at the moment in the `utils_test.go` file.

Closes #832

## Summary by CodeRabbit

* **New Features**
  * Query pending transactions by transaction ID in addition to Intent-based lookups.

* **Improvements**
  * Spend checks now detect pending transactions and return a dedicated error including outpoint and pending txid metadata to clarify conflicts and races.
  * Single-tx fetch endpoint/flow added for txid-based retrieval.

* **Tests**
  * Added e2e and helper tests for txid lookup, nonexistent/invalid txid, and double-spend scenarios.


internal/core/application/service.go
```diff
@@ -529,17 +529,21 @@ func (s *service) SubmitOffchainTx(
...
 			return nil, errors.INTERNAL_ERROR.New("something went wrong").
 				WithMetadata(map[string]any{"vtxos": vtxo.Outpoint.String()})
 		}
 
-		if isSpent {
-			return nil, errors.VTXO_ALREADY_SPENT.New("%s already spent", vtxo.Outpoint.String()).
-				WithMetadata(errors.VtxoMetadata{VtxoOutpoint: vtxo.Outpoint.String()})
+		if pendingTxid != "" {
+			return nil, errors.VTXO_PENDING_TX.New(
+				"%s is locked by pending tx %s", vtxo.Outpoint.String(), pendingTxid,
```
altafan: Let's be more generic, this way the error can be read by humans as it is, if not handled client-side

```suggestion
				"input(s) spent by another pending tx"
```

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -326,19 +326,37 @@ func (h *handler) GetPendingTx(
 		return nil, status.Error(codes.InvalidArgument, "missing identifier")
 	}
 
-	intent := req.GetIntent()
-	if intent == nil {
-		return nil, status.Error(codes.InvalidArgument, "missing intent")
-	}
+	var pendingTxs []application.AcceptedOffchainTx
+	var err error
```
altafan: no need for this var, can be dropped

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -326,19 +326,37 @@ func (h *handler) GetPendingTx(
...
-	}
+	switch identifier := req.GetIdentifier().(type) {
+	case *arkv1.GetPendingTxRequest_Txid:
+		if identifier.Txid == "" {
+			return nil, status.Error(codes.InvalidArgument, "missing txid")
+		}
+		pendingTx, svcErr := h.svc.GetPendingOffchainTxByTxid(ctx, identifier.Txid)
+		if svcErr != nil {
+			return nil, svcErr
+		}
```
altafan:
```suggestion
		pendingTx, err := h.svc.GetPendingOffchainTxByTxid(ctx, identifier.Txid)
		if svcErr != nil {
			return nil, err
		}
```

internal/test/e2e/e2e_test.go
```diff
@@ -3982,3 +3983,282 @@ func TestFee(t *testing.T) {
...
 	require.Empty(t, bobBalance.OnchainBalance.LockedAmount)
 }
+
+// TestGetPendingTxByTxid tests the GetPendingTx RPC when called with a txid identifier.
+// This test verifies that:
+// 1. A submitted (but not finalized) transaction can be retrieved by its txid
+// 2. The pending transaction contains the expected data (ark txid, final ark tx, signed checkpoints)
+// 3. Non-existent txids return an empty response
+func TestGetPendingTxByTxid(t *testing.T) {
+	t.Run("retrieve pending tx by txid", func(t *testing.T) {
```
altafan: Let's change the name of this test so we can later add another for getting a pending tx by intent. Also, let's move 3) into a dedicated sub-test `t.Run("with non existing txid", func(t *testing.T) {...})` (it's just enough to call the util function with a ranodm txid, the rest can be fully skipped)
```suggestion
// 3. Non-existent txids return an empty response
func TestGetPendingTx(t *testing.T) {
	t.Run("with txid", func(t *testing.T) {
```

internal/test/e2e/e2e_test.go
```diff
@@ -3982,3 +3983,282 @@ func TestFee(t *testing.T) {
...
+		)
+		require.NoError(t, err)
+
+		// Submit the transaction but DO NOT finalize it
+		txid, _, signedCheckpoints, err := arkSvc.SubmitTx(ctx, signedArkTx, encodedCheckpoints)
+		require.NoError(t, err)
+		require.NotEmpty(t, txid)
+		require.NotEmpty(t, signedCheckpoints)
+
+		t.Logf("Submitted pending tx with txid: %s", txid)
```
altafan: can you remove all the logs please?

internal/test/e2e/e2e_test.go
```diff
@@ -3982,3 +3983,282 @@ func TestFee(t *testing.T) {
...
+		// Use HTTP REST API to call GetPendingTx
+		httpClient := &http.Client{Timeout: 15 * time.Second}
+
+		// Test 1: Retrieve the pending tx by txid
+		// Note: The pending tx might be in the cache or repository depending on server config
+		pendingTxResp, err := getPendingTxByTxid(httpClient, txid)
+		if err != nil {
+			t.Logf("GetPendingTx returned error: %v (this might indicate the server doesn't have the tx in cache)", err)
+		}
+		require.NoError(t, err, "GetPendingTx should return the submitted pending transaction")
```
altafan: can you please remove all message overwrites?
```suggestion
		require.NoError(t, err)
```

internal/test/e2e/e2e_test.go
```diff
@@ -3982,3 +3983,282 @@ func TestFee(t *testing.T) {
...
+
+		// Test 2: Non-existent txid should return empty response
+		nonExistentResp, err := getPendingTxByTxid(httpClient, "0000000000000000000000000000000000000000000000000000000000000000")
+		require.NoError(t, err)
+		require.NotNil(t, nonExistentResp)
+		require.Empty(t, nonExistentResp.PendingTxs)
+
+		// Test 3: Empty txid should return error
+		_, err = getPendingTxByTxid(httpClient, "")
+		require.Error(t, err)
```
altafan: this can be isolated into a subtest `t.Run("invalid", func(t *testing.T) {...})`

internal/test/e2e/e2e_test.go
```diff
@@ -3982,3 +3983,282 @@ func TestFee(t *testing.T) {
...
+		time.Sleep(time.Second)
+
+		// After finalization, the tx should no longer be pending
+		afterFinalizeResp, err := getPendingTxByTxid(httpClient, txid)
+		require.NoError(t, err)
+		require.NotNil(t, afterFinalizeResp)
+		require.Empty(t, afterFinalizeResp.PendingTxs)
+	})
+
+	t.Run("vtxo pending tx error when spending locked vtxo", func(t *testing.T) {
```
altafan: in ref to https://github.com/arkade-os/arkd/pull/869/files#r2627284295 let's have many invalid cases.. this is a recap of the test structure:

```
func TestGetPendingTx(t *testing.T) {
     t.Run("valid", func (t *testing.T) {
          t.Run("with txid", func(t *testing.T) {...})
     }
     t.Run("invalid", func (t *testing.T) {
          t.Run("missing txid", func(t *testing.T) {...})
          t.Run("double spend vtxo from pending tx", func(t *testing.T) {...})
     }
}
```

internal/test/e2e/e2e_test.go
```diff
@@ -3982,3 +3983,282 @@ func TestFee(t *testing.T) {
...
+		// The server should detect this VTXO is locked by a pending transaction
+		// and return either VTXO_PENDING_TX or a "duplicated" error (if same txid)
+		_, _, _, err = arkSvc.SubmitTx(ctx, signedArkTx, encodedCheckpoints)
+		require.Error(t, err)
+		// Either error is acceptable - both indicate the VTXO cannot be spent
+		errContainsPending := strings.Contains(err.Error(), "VTXO_PENDING_TX") ||
+			strings.Contains(err.Error(), "pending")
+		errContainsDuplicated := strings.Contains(err.Error(), "duplicated")
+		require.True(t, errContainsPending || errContainsDuplicated,
+			"expected error to contain 'pending' or 'duplicated', got: %s", err.Error())
```
altafan: we should make this consistent so that here we check against only one err type. Do you know why the reason for this inconsistent behavior?

internal/test/e2e/utils_test.go
```diff
@@ -781,3 +781,44 @@ func refill(httpClient *http.Client) error {
...
+	PendingTxs []pendingTx `json:"pendingTxs"`
+}
+
+type pendingTx struct {
+	ArkTxid             string   `json:"arkTxid"`
+	FinalArkTx          string   `json:"finalArkTx"`
+	SignedCheckpointTxs []string `json:"signedCheckpointTxs"`
+}
+
+func getPendingTxByTxid(httpClient *http.Client, txid string) (*pendingTxResponse, error) {
```
altafan, line 795: we're gonna open pr on go-sdk for u so that we can drop this. Will ping u as soon as it's done

altafan (changes requested):
@Chigala Thank you for the contrib! 💪

Can you please add unit tests to `internal/infrastructure/live-store/live_store_test.go` for the newly introduced `OffChainTxStore.GetTxidByOutpoint(...)` api?

Also, please run locally `make lint` to lint the code so that ci moves on and runs unit tests

internal/test/e2e/e2e_test.go
Chigala: So here, I was trying to test for the `VTXO_PENDING_TX` error by sending the same transaction twice.Thinking about this, this is actually a bad test because it wasn't actually testing the scenario described by its name, the server would always return the "duplicated error" first and never reach the error I was actually testing for. I will fix this.

PR #874, opened 2025-12-30
Fix auth service at restart

altafan:
Atm, if the service is not gracefully shutdown, the macaroon service won't be unlocked at the very next restart. This because arkd would receive immediately a "ready" signal from arkd-wallet and would start the app svc wihtout passing from unlock, required to unlock the macaroon service to then check authorized requests.

The following pr addresses the described behavior and also makes arkd-wallet more robust:
* arkd-wallet:
    - in app layer, moved logic to send the ready update in a go routine and added a small latency to make sure on arkd the logic "on unlock" always runs before the one "on is ready"
    - fix launcher by reacting to SIGHUP
    - even if not strictly needed, added changes to interface layer to support multiple listeners on GetReadyUpdate streaming rpc
* arkd:
    - keep listening to ready update stream instead of returning once the very first event is received
    - make sure to lock the wallet if a ready update is received but the wallet was never unlocked with the password
    - fix launcher by reacting to SIGHUP

Please @louisinger review.

## Summary by CodeRabbit

* **New Features**
  * Added SIGHUP support for graceful shutdown.
  * Readiness updates now stream continuously to multiple subscribers.
  * Introduced a broker to manage multi-listener subscriptions for readiness events.
  * Macaroon service exposes lock status checks.

* **Bug Fixes**
  * Readiness notification on unlock is delayed and non-blocking to avoid blocking.
  * Improved wallet lifecycle and centralized shutdown/context cancellation handling.

* **Tests**
  * Added comprehensive tests for the broker and listener behavior.

### DIFF
--- a/internal/interface/grpc/service.go
+++ b/internal/interface/grpc/service.go
@@ -505,6 +505,8 @@ func (s *service) autoUnlock() error {
 		return nil
 	}
 
+	// If the wallet is already unlocked, force the lock to make the very next call to Unlock
+	// to take effect and run the onUnlock callback
 	if status.IsUnlocked() {
 		// nolint
 		wallet.Lock(ctx)


pkg/arkd-wallet/interface/grpc/service.go
```diff
@@ -32,6 +32,7 @@ type service struct {
 	closeFn           func()
 	otelShutdown      func(context.Context) error
 	pyroscopeShutdown func() error
+	stopFn            func()
```
louisinger, line 35:
```suggestion
	cancelFn            func()
```

louisinger (approved):

Issue #863, opened 2025-12-08
Introduce additional filters for GetVtxos


Kukks:
`from` and `until` would allow us to select data based on a "last sync timestamp".

HOWEVER, this filter must be on a  lastUpdated not simply created flag. If a vtxo is spent/swept/expires/ etc, it should have its updated timestamp...updated

altafan:
Yes, query the view.. it contains also info related to the commitment txids

PR #879, opened 2026-01-06
GetVtxos filter for time range

bitcoin-coder-bob:
Issue: https://github.com/arkade-os/arkd/issues/863

Added ability to add a `after` and `before` time on a new `updated_at` field to the vtxos table. `SelectPendingSpentVtxosWithPubkeys` and `SelectVtxosWithPubkeys` now use these new params. A value of `0` for `after` indicates a search time range from the beginning of time. A value of `0` for `before` indicates a search range until the end of time. If both are set to `0` it as an unbounded time range, searching thru all time.

## Summary by CodeRabbit

* **New Features**
  * GetVtxos supports time-range filters (after, before) to query VTXOs by last-update time.
  * VTXOs now include an updated_at timestamp and updates populate that field for time-aware queries.

* **Documentation**
  * API specs updated to expose the new after/before query parameters.

* **Bug Fixes**
  * Removed stray debug logging from intent fees handling.

* **Tests**
  * Added time-range unit tests covering valid, invalid, boundary, and unbounded scenarios.

### DIFF
--- a/internal/interface/grpc/handlers/indexer.go
+++ b/internal/interface/grpc/handlers/indexer.go
@@ -253,7 +253,7 @@ func (e *indexerService) GetVtxos(
 
 		after, before, err := parseTimeRange(request.GetAfter(), request.GetBefore())
 		if err != nil {
-			return nil, err
+			return nil, status.Error(codes.InvalidArgument, err.Error())
 		}
 
 		resp, err = e.indexerSvc.GetVtxos(
@@ -268,7 +268,7 @@ func (e *indexerService) GetVtxos(
 		resp, err = e.indexerSvc.GetVtxosByOutpoint(ctx, outpoints, page)
 	}
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "%s", err.Error())
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 
 	vtxos := make([]*arkv1.IndexerVtxo, 0, len(resp.Vtxos))


internal/infrastructure/db/postgres/migration/20260106000000_add_vtxo_updated_at.down.sql
```diff
@@ -0,0 +1,3 @@
+DROP VIEW IF EXISTS intent_with_inputs_vw;
+DROP VIEW IF EXISTS vtxo_vw;
```
altafan, line 2: we need to recreate the "old" views here or we are just dropping them instead of reverting the changes

internal/infrastructure/db/postgres/sqlc/query.sql
```diff
@@ -64,12 +64,17 @@ VALUES (
     swept = EXCLUDED.swept,
     preconfirmed = EXCLUDED.preconfirmed,
     expires_at = EXCLUDED.expires_at,
-    created_at = EXCLUDED.created_at;
+    created_at = EXCLUDED.created_at,
+    updated_at = EXCLUDED.updated_at;
```
altafan: what if we harcode "time.Now()" in sql terms here? This way we wouldn't have any change in the repo impl. If it makes sense, i'd go for this also in the other queries

internal/infrastructure/db/postgres/sqlc/query.sql
```diff
@@ -64,12 +64,17 @@ VALUES (
...
 
 -- name: InsertVtxoCommitmentTxid :exec
 INSERT INTO vtxo_commitment_txid (vtxo_txid, vtxo_vout, commitment_txid)
 VALUES (@vtxo_txid, @vtxo_vout, @commitment_txid);
 
+-- name: SelectVtxosUpdatedInTimeRange :many
+SELECT sqlc.embed(vtxo_vw) FROM vtxo_vw 
+WHERE updated_at >= @after AND updated_at <= @before;
+
```
altafan: i know this file became a bit messy, one day we'll likely split it into several ones. For now, can you please move this query at L345?

altafan: also, watch out! if _before_ is not defined, we should skip the check  `AND updated_at <= @before;` or it will make the query return nothing

internal/infrastructure/db/badger/vtxo_repo.go
```diff
@@ -656,3 +677,10 @@ func (r *vtxoRepository) GetAllChildrenVtxos(
 
 	return outpoints, nil
 }
+
+func (r *vtxoRepository) GetVtxosUpdatedInTimeRange(
+	ctx context.Context, after, before int64,
+) ([]domain.Vtxo, error) {
+	query := badgerhold.Where("UpdatedAt").Ge(after).And("UpdatedAt").Le(before)
```
altafan: if _before_ is not defined this method will always return an empty list

internal/infrastructure/db/service_test.go
```diff
@@ -1041,6 +1041,17 @@ func testVtxoRepository(t *testing.T, svc ports.RepoManager) {
...
+			vtxosInTimeRange, err := svc.Vtxos().
+				GetVtxosUpdatedInTimeRange(ctx, now.UnixMilli(), time.Now().UnixMilli())
+			require.NoError(t, err)
+			require.GreaterOrEqual(t, len(vtxosInTimeRange), 3)
+
+			// Test with a time range that yields no results
+			vtxosInTimeRange, err = svc.Vtxos().
+				GetVtxosUpdatedInTimeRange(ctx, now.UnixMilli(), now.UnixMilli()+10)
+			require.NoError(t, err)
+			require.Equal(t, len(vtxosInTimeRange), 0)
```
altafan: let's test also the cases: (1) "after not defined (0 or negative value)", and (2) "before not defined"

internal/infrastructure/db/badger/vtxo_repo.go
bitcoin-coder-bob: what behavior would you like it to exhibit? before will never be nil but could be the zero value, in which case should we throw an error?

internal/infrastructure/db/service_test.go
bitcoin-coder-bob: do we want to throw an error if neither are defined or are zero values?

internal/infrastructure/db/postgres/migration/20260106000000_add_vtxo_updated_at.down.sql
bitcoin-coder-bob: looking at another migration down file (`internal/infrastructure/db/postgres/migration/20250702115345_init.down.sql`) it does this same `DROP VIEW IF EXISTS intent_with_inputs_vw` command. The corresponding up file for my migration (`internal/infrastructure/db/postgres/migration/20250702115345_init.up.sql`) does the view creation. 

Given that can you help me understand the ask, I am a bit unclear what to do given I do the view creation

internal/infrastructure/db/badger/vtxo_repo.go
bitcoin-coder-bob:
```
if after <= 0 || before <= 0 || before < after {
		return nil, fmt.Errorf("invalid time range")
	}
```

bitcoin-coder-bob: if we allow after to be 0 then the query could be used for "get all vtxos before XYZ time" which maybe is a desirable behavior

internal/infrastructure/db/postgres/sqlc/query.sql
bitcoin-coder-bob: resolved in [cde1fc9](https://github.com/arkade-os/arkd/pull/879/commits/cde1fc9260bb243916968b4d84ee8f857e637dad)

internal/infrastructure/db/postgres/sqlc/query.sql
bitcoin-coder-bob: addressed in: [0fa216a](https://github.com/arkade-os/arkd/pull/879/commits/0fa216aadc09ba86dc40d2675384ac4c781e1b19)

internal/infrastructure/db/service_test.go
bitcoin-coder-bob: I implemented:

```
if after <= 0 {
		return nil, fmt.Errorf("after must be greater than 0")
	} else if before > 0 && before <= after {
		return nil, fmt.Errorf("before must be greater than after")
	}
```

so `before` can be 0 indicating no upper bound

internal/infrastructure/db/badger/vtxo_repo.go
bitcoin-coder-bob: https://github.com/arkade-os/arkd/pull/879#discussion_r2669264294

bitcoin-coder-bob:
@altafan added the before/after params into the `indexer.proto` field `GetVtxosRequest` and use it. I have it fallback to a value of `1` when `after` is set to 0. `after` must be above 0, and setting to `1` gives us a beginning of time value. All this added in: [a3d7388](https://github.com/arkade-os/arkd/pull/879/commits/a3d7388944513f1bab826ba3a857d2a360a70529)

internal/infrastructure/db/badger/vtxo_repo.go
```diff
@@ -656,3 +677,19 @@ func (r *vtxoRepository) GetAllChildrenVtxos(
 
 	return outpoints, nil
 }
+
+func (r *vtxoRepository) GetVtxosUpdatedInTimeRange(
+	ctx context.Context, after, before int64,
+) ([]domain.Vtxo, error) {
+	if after < 0 || before < 0 {
```
altafan: we want to return an error only in case both _after_ and _before_  are "undefined"

```suggestion
	if after <= 0 && before <= 0 {
```

bitcoin-coder-bob: I was using 0 for `after` to indicate "from the start of time" and 0 for `before` to indicate "until the end of time, unbounded". If we instead error when a 0 is passed then we lost this capability and we have to pass `1` for `after` to indicate "beginning of time" and infinity for `before` to indicate "until the end of time"

internal/infrastructure/db/postgres/migration/20260106000000_add_vtxo_updated_at.down.sql
altafan: ah! That's just because `init` is the very first migration that creates all tables and views (up) or deletes everything (down).

In this case, instead, we want to recreate the views after updating the tables (up) or rollback to previous version (down), that means recreating the views again after reverting the changes to the tables

bitcoin-coder-bob: addressed in commit: [d6fd3a1](https://github.com/arkade-os/arkd/pull/879/commits/d6fd3a16202860ad731f0e7f94337514b4a8d0fb)

internal/infrastructure/db/badger/vtxo_repo.go
bitcoin-coder-bob: addressed over a call, 0 determined to be allowed to indicate unbounded time. So 0 for `after` indicates "since the start of time" and 0 for `before` indicates "until the end of time"

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -188,6 +188,11 @@ message GetVtxosRequest {
   IndexerPageRequest page = 6;
   // Include only spent vtxos that are not finalized.
   bool pending_only = 7;
+  // vtxos from after unix time in milliseconds, must be above 0
```
altafan:
```suggestion
  // Include only vtxos with last update after the given unix time in millisecond
```

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -188,6 +188,11 @@ message GetVtxosRequest {
...
+  int64 after = 8;
+  // vtxos from before unix time in milliseconds, using 0 leaves no upper time bound, must be
+  // greater value than the after
```
altafan:
```suggestion
  // vtxos Include only vtxos with last update before the given unix time in milliseconds
  // greater value than the after when specified
```

internal/core/application/indexer.go
```diff
@@ -30,7 +30,7 @@ type IndexerService interface {
 	GetConnectors(ctx context.Context, txid string, page *Page) (*TreeTxResp, error)
 	GetVtxos(
 		ctx context.Context,
-		pubkeys []string, spendableOnly, spendOnly, recoverableOnly, pendingOnly bool, page *Page,
+		pubkeys []string, spendableOnly, spendOnly, recoverableOnly, pendingOnly bool, page *Page, after, before int64,
```
altafan: let's keep `page` as last function arg:
```suggestion
		pubkeys []string, spendableOnly, spendOnly, recoverableOnly, pendingOnly bool, after, before int64, page *Page,
```

internal/infrastructure/db/badger/vtxo_repo.go
```diff
@@ -221,11 +230,19 @@ func (r *vtxoRepository) UpdateVtxosExpiration(
...
 
 func (r *vtxoRepository) GetAllVtxosWithPubKeys(
-	ctx context.Context, pubkeys []string,
+	ctx context.Context, pubkeys []string, after, before int64,
 ) ([]domain.Vtxo, error) {
+	if after < 0 || before < 0 {
+		return nil, fmt.Errorf("after and before must be greater than or equal to 0")
+	} else if before > 0 && after > 0 && before <= after {
+		return nil, fmt.Errorf("before must be greater than after")
+	}
```
altafan: let's move to a function `validateTimeRange` so we avoid duplicating code?

internal/infrastructure/db/badger/vtxo_repo.go
```diff
@@ -452,19 +481,19 @@ func (r *vtxoRepository) addVtxos(
...
 	} else {
-		err = r.store.Get(outpoint.String(), &vtxo)
+		err = r.store.Get(outpoint.String(), &dto)
 	}
 	if err != nil && err == badgerhold.ErrNotFound {
 		return nil, nil
 	}
 
-	return &vtxo, nil
+	return &dto.Vtxo, nil
```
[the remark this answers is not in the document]
altafan: @bitcoin-coder-bob the rabbit is right, even tho it refers to already existing code. Let's take the opportunity to fix the check L492:
```
if err != nil {
    if err == badgerhold.ErrNotFound {
		return nil, nil
	}
    return nil, err
}
```

internal/infrastructure/db/postgres/migration/20260106000000_add_vtxo_updated_at.up.sql
```diff
@@ -0,0 +1,25 @@
+ALTER TABLE vtxo
+ADD COLUMN IF NOT EXISTS updated_at BIGINT NOT NULL DEFAULT (EXTRACT(EPOCH FROM NOW()) * 1000)::BIGINT;
+
+-- set existing null updated_at values to current timestamp
+UPDATE vtxo
+SET updated_at = (EXTRACT(EPOCH FROM NOW()) * 1000)::BIGINT
+WHERE updated_at IS NULL;
```
altafan, line 7: just to be sure, this is gonna set the updated_at of all existing records to _now_?

internal/infrastructure/db/postgres/sqlc/query.sql
```diff
@@ -64,7 +64,8 @@ VALUES (
     swept = EXCLUDED.swept,
     preconfirmed = EXCLUDED.preconfirmed,
     expires_at = EXCLUDED.expires_at,
-    created_at = EXCLUDED.created_at;
+    created_at = EXCLUDED.created_at,
+    updated_at = (EXTRACT(EPOCH FROM NOW()) * 1000)::BIGINT;
```
altafan, line 68: why not `EXCLUDED.updated_at`?

internal/infrastructure/db/postgres/sqlc/query.sql
```diff
@@ -112,17 +113,17 @@ UPDATE vtxo SET intent_id = @intent_id WHERE txid = @txid AND vout = @vout;
 UPDATE vtxo SET expires_at = @expires_at WHERE txid = @txid AND vout = @vout;
 
 -- name: UpdateVtxoUnrolled :exec
-UPDATE vtxo SET unrolled = true WHERE txid = @txid AND vout = @vout;
+UPDATE vtxo SET unrolled = true, updated_at = @updated_at WHERE txid = @txid AND vout = @vout;
```
altafan: why don't we use `(EXTRACT(EPOCH FROM NOW()) * 1000)::BIGINT;` so we have 0 changes on the go-side when we use the correponding func?

internal/infrastructure/db/service_test.go
```diff
@@ -1006,17 +1006,87 @@ func testVtxoRepository(t *testing.T, svc ports.RepoManager) {
...
+			require.Len(t, pendingSpentVtxosByPubkey, 1)
+
+			// Test with time range that excludes the vtxo
+			pendingSpentVtxosByPubkey, err = svc.Vtxos().GetPendingSpentVtxosWithPubKeys(
+				ctx, []string{"bbbb"}, currTime.Add(-2*time.Hour).UnixMilli(), currTime.Add(-1*time.Hour).UnixMilli(),
+			)
+			require.NoError(t, err)
+			require.Empty(t, pendingSpentVtxosByPubkey)
+
+			// Test with invalid time range where after is greater than before
```
altafan, line 1050: would be nice to have all these cases in a sub-test called `invalid`, but it needs a little bit of refactor IMO. Can you add a todo like?
```suggestion
			// TODO: move to "invalid" sub-test
			// Test with invalid time range where after is greater than before
```

internal/interface/grpc/handlers/indexer.go
```diff
@@ -248,9 +248,18 @@ func (e *indexerService) GetVtxos(
...
-			ctx, pubkeys, spendableOnly, spentOnly, recoverableOnly, pendingOnly, page,
+			ctx,
+			pubkeys,
+			spendableOnly,
+			spentOnly,
+			recoverableOnly,
+			pendingOnly,
+			page,
+			request.GetAfter(),
+			request.GetBefore(),
```
altafan: let's validate these fields and make sure to return an InvalidArgument error in case they are negative

internal/infrastructure/db/badger/vtxo_repo.go
bitcoin-coder-bob: is there 1 place I could make this function and share it among the 3 db vtxo_repo files this will be called? I don't see a great file to place it in.

internal/infrastructure/db/postgres/migration/20260106000000_add_vtxo_updated_at.up.sql
bitcoin-coder-bob: yes that is my understanding

internal/infrastructure/db/postgres/sqlc/query.sql
bitcoin-coder-bob: https://github.com/arkade-os/arkd/pull/879#discussion_r2669198708

internal/interface/grpc/handlers/indexer.go
bitcoin-coder-bob:
```
if len(pubkeys) > 0 {
		resp, err = e.indexerSvc.GetVtxos(
			ctx,
			pubkeys,
			spendableOnly,
			spentOnly,
			recoverableOnly,
			pendingOnly,
			request.GetAfter(),
			request.GetBefore(),
			page,
		)
	}
```

this error was never getting checked. I will add that too

internal/infrastructure/db/postgres/sqlc/query.sql
bitcoin-coder-bob: given we are making an update on the row it felt appropriate to then modify the `updated_at`

bitcoin-coder-bob: when I change it back: https://github.com/arkade-os/arkd/pull/879#discussion_r2683341147

internal/infrastructure/db/badger/vtxo_repo.go
bitcoin-coder-bob: made a `dbutil` folder inside the `db` folder, where I put a `utils.go` containing this new util function

bitcoin-coder-bob: added here: [c839c74](https://github.com/arkade-os/arkd/pull/879/commits/c839c746c6a9489639bde131d7e55e09770286f4)

internal/infrastructure/db/postgres/sqlc/query.sql
```diff
@@ -64,7 +64,8 @@ VALUES (
     swept = EXCLUDED.swept,
     preconfirmed = EXCLUDED.preconfirmed,
     expires_at = EXCLUDED.expires_at,
-    created_at = EXCLUDED.created_at;
+    created_at = EXCLUDED.created_at,
+    updated_at = EXCLUDED.updated_at;
```
altafan: i'm sorry, i suggested you to use updated_at here, but also to not use it in the other queries below so we don't need to pass this param from go.

Let's be consistent and "rollback" these changes so we keep using `(EXTRACT(EPOCH FROM NOW()) * 1000)::BIGINT` everywhere (same applies to sqlite)

internal/core/application/indexer.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/arkade-os/arkd/internal/core/domain"
 	"github.com/arkade-os/arkd/internal/core/ports"
+	dbutil "github.com/arkade-os/arkd/internal/infrastructure/db/dbutil"
```
altafan: let's not import stuff infrastructure in the application layer, as it should depend only on _domain_ and _ports_. In this case, let's just rewrite the function `as validateTimeRange(...)` in internal/core/application/utils.go

bitcoin-coder-bob: I have now removed the new dbutils and added this to `internal/core/application/utils.go` in commit: [003da6c](https://github.com/arkade-os/arkd/pull/879/commits/003da6ce0a9c333e5d6572ab31af6be994c8cdfc)

internal/infrastructure/db/postgres/sqlc/query.sql
bitcoin-coder-bob: changed in: [003da6c](https://github.com/arkade-os/arkd/pull/879/commits/003da6ce0a9c333e5d6572ab31af6be994c8cdfc)

Issue #883, opened 2026-01-13
GetIntentWithTxid API


altafan:
Let's add a new endpoint that returns the proof and message for the intent with the given txid of an existing batch. If the intent is not found in db, return an error `intent with tixd <txid> not found`.

```
rpc GetIntent(GetIntentRequest) returns (GetIntentResponse) {
    option (meshapi.gateway.http) = {
      get: "/v1/intent"
    };
}

message GetIntentWithTxidRequest {
    oneof filter {
      string txid = 1;
    }
}
message GetIntentWithTxidRequest {
    Intent intent = 1;
}
```

Note: we use `oneof` because, later, we will add support to retrieve an intent (either from an existing batch or still in the intent queue) via another intent that represents proof of ownership.

bitcoin-coder-bob:
@altafan Clarification questions. My understanding is that a batch is the same thing as a round. I see that a round has a `CommitmentTxid` field, and also has a map of intents associated with it (`Intents  map[string]Intent`), so it is unclear to me which intent to return since many intents could be associated to the round's `CommitmentTxid` and the rpc defined in this issue calls for getting just 1 intent. 

If the new rpc to build is to take in txid (which is presumably the round's commitment txid), and a round can have many intents, and the rpc is to return just 1 intent, it is ambiguous to me which intent to return as the round can have >1 intent.


However I see that there is a `vtxo` table that has a `commitment_txid` and a `txid` field. Is the idea for this new rpc to pass along a txid string that would map to the vtxo's `txid` or to its `commitment_txid` or to its `ark_txid`?  I think my questions boil down to for the txid to be passed in this new rpc call, what is this txid? I see many different types of txid related fields and am unclear which to search on, so that I can link it back to a singular intent.

bitcoin-coder-bob:
Given the vtxo table: https://github.com/arkade-os/arkd/blob/5eb9d8484794f1288da1bb3701982599bea43a7b/internal/infrastructure/db/postgres/migration/20250702115345_init.up.sql#L42


Given a txid in the new rpc, I could find all rows that match on this txid, but presumably >1 row can have the same txid since the row is not enforced to be unique. But for each row returned we would have an `intent_id` field we could use to lookup the intent to ultimately collect the proof and message field we want to return back in the new rpc. Still the same issue as the prior comment in that we can end up getting multiple vtxo rows yielding different intents, and becomes ambiguous which intent to be using.

bitcoin-coder-bob:
[...]

PR #884, opened 2026-01-14
get intent by txid

bitcoin-coder-bob:
Issue: https://github.com/arkade-os/arkd/issues/883

rpc for `GetIntent` allowing caller to pass a `txid` and receive back corresponding intent.

Currently the `intent` table does not have a column for txid. This makes the lookup for intent by txid not easy. We can derive the txid of an intent via the `proof` column. This PR includes a db migration to add this new `txid` field, and backfill existing rows by grabbing the `proof` and deriving the `txid` and storing this in the new `txid` field.

The migration is crafted into files for sqlite and postgres called `intent_txid_migration.go` which will:
1) check for existence of new `txid` column in `intent` table
2) backfill txid based off each intent row's proof
3) create an index on the txid, making efficient lookup for our new rpc to get intent by txid

I have this custom migration setup as:
```
intentTxidMigrationBegin := uint(20260114000000)
.
.
.
		if version < intentTxidMigrationBegin {
			if err := m.Migrate(intentTxidMigrationBegin); err != nil &&
.
.
.

```

(where `20260114000000` is the date of the new schema addtion migration sql file)

The idea here is that when we detect the db has not had the newest schema addition applied (which simply adds the `txid` column to the `intent` table), we apply this simple schema addition. Then we immediately backfill in this new `txid` column via the code in `intent_txid_migration.go`.

Testing was performed as follows:
1) checkout master branch, bring up arkd (with sqlite configured as arkd's db), run integration tests to generate some db data.
2) switch to this feature branch, run arkd. This will now trigger the migration code.
3) Check that the sqlite db has the new `txid` column, and that the existing `intent` rows have this `txid` populated.

NOTE: This same custom migration is not being performed for badger.

Going forward we will write intent rows with the txid.

## Summary by CodeRabbit

* **New Features**
  * GET /v1/intent to retrieve an intent by txid (returns proof and message).
  * Service API and handlers to look up intents by txid.

* **Bug Fixes / Errors**
  * New INTENT_NOT_FOUND error returned when no intent matches a queried txid.

* **Chores (Data & Migrations)**
  * Added intent.txid column, backfill tooling, and index to populate and speed txid lookups.

* **Tests**
  * Integration tests validating txid derivation and backfill.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,6 +39,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
 	github.com/lib/pq v1.10.9
 	github.com/lightningnetwork/lnd v0.18.2-beta
+	github.com/ltcsuite/ltcd v0.23.5
 	github.com/meshapi/grpc-api-gateway v0.1.0
 	github.com/nbd-wtf/go-nostr v0.40.1
 	github.com/redis/go-redis/v9 v9.10.0
@@ -157,7 +158,6 @@ require (
 	github.com/lightningnetwork/lnd/tlv v1.2.6 // indirect
 	github.com/lightningnetwork/lnd/tor v1.1.3 // indirect
 	github.com/lithammer/shortuuid/v3 v3.0.7 // indirect
-	github.com/ltcsuite/ltcd v0.23.5 // indirect
 	github.com/ltcsuite/ltcd/chaincfg/chainhash v1.0.2 // indirect
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
--- a/pkg/arkd-wallet/go.mod
+++ b/pkg/arkd-wallet/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/arkade-os/arkd/pkg/ark-lib v0.0.0-00010101000000-000000000000
 	github.com/btcsuite/btcd/btcec/v2 v2.3.5
 	github.com/dgraph-io/badger/v4 v4.8.0
+	github.com/google/uuid v1.6.0
 	github.com/grafana/pyroscope-go v1.2.7
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
 	github.com/lightningnetwork/lnd v0.18.2-beta
@@ -38,7 +39,6 @@ require (
 	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
 	github.com/dgraph-io/ristretto/v2 v2.2.0 // indirect
 	github.com/google/flatbuffers v25.2.10+incompatible // indirect
-	github.com/google/uuid v1.6.0 // indirect
 	github.com/grafana/pyroscope-go/godeltaprof v0.1.9 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.27.2 // indirect
 	github.com/onsi/ginkgo v1.16.4 // indirect


bitcoin-coder-bob:
@coderabbitai review

bitcoin-coder-bob:
@coderabbitai review

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -590,6 +590,29 @@ func (h *handler) listenToTxEvents() {
...
+) (*arkv1.GetIntentResponse, error) {
+	var err error
+	var intent domain.Intent
+
+	switch filter := req.GetFilter().(type) {
+	case *arkv1.GetIntentRequest_Txid:
+		intent, err = h.svc.GetIntentByTxid(ctx, filter.Txid)
+		if err != nil {
+			return nil, err
+		}
```
altafan: can be moved outside the switch

internal/core/application/types.go
```diff
@@ -55,9 +55,12 @@ type Service interface {
 		proof intent.Proof,
 		message intent.DeleteMessage,
 	) errors.Error
-
 	// TODO: remove when detaching the indexer svc.
 	GetIndexerTxChannel(ctx context.Context) <-chan TransactionEvent
+	GetIntentByTxid(
+		ctx context.Context,
+		txid string,
+	) (domain.Intent, errors.Error)
```
altafan:
```suggestion
	) (*domain.Intent, errors.Error)
```

internal/core/application/service.go
```diff
@@ -4096,6 +4096,20 @@ func (s *service) verifyForfeitTxsSigs(roundId string, txs []string) []domain.Co
 	return convictions
 }
 
+func (s *service) GetIntentByTxid(
+	ctx context.Context,
+	txid string,
+) (domain.Intent, errors.Error) {
+	intent, err := s.repoManager.Rounds().GetIntentByTxid(ctx, txid)
+	if err != nil {
+		return domain.Intent{}, errors.INTERNAL_ERROR.New(
```
```suggestion
		return nil, errors.INTERNAL_ERROR.New(
```

internal/core/domain/intent.go
```diff
@@ -12,14 +14,17 @@ type Intent struct {
 	Receivers []Receiver
 	Proof     string
 	Message   string
+	Txid      string
 }
 
 func NewIntent(proof, message string, inputs []Vtxo) (*Intent, error) {
+	txid, _ := deriveTxidFromProof(proof)
```
altafan: instead of this, let's change the signature of the func and expected the txid to be passed as well (we assume it's the right one)

```
func NewIntent(proofTxid, proof, message string, inputs []Vtxo) (*Intent, error) {
```

internal/core/domain/intent.go
```diff
@@ -121,3 +129,11 @@ func (t Intents) HaveOnlyOnchainOutput() bool {
...
 	return true
 }
+
+func deriveTxidFromProof(proof string) (string, error) {
+	tx, err := psbt.NewFromRawBytes(strings.NewReader(proof), true)
+	if err != nil {
+		return "", fmt.Errorf("error deriving txid from proof: %w", err)
+	}
+	return tx.UnsignedTx.TxID(), nil
+}
```
altafan: no need, can be dropped

internal/core/domain/intent_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 var (
 	pubkey  = "25a43cecfa0e1b1a4f72d64ad15f4cfa7a84d0723e8511c969aa543638ea9967"
-	proof   = "proof"
+	proof   = "cHNidP8BALACAAAAAyLh21ahYqxWy2slXY09ZoLvYWKFo7R0zPtbnJqkSUq6AAAAAAAAAAAA9PRZ58PC4lJAJ7VEYtMMUC1E+poh/Gxo+0TNqvgh27wAAAAAAAAAAABRRy0Q2j/vH4YYRfWfzzF4LBB1xX+tQdFOIWTU6ZD0CwMAAAAAAAAAAAHoAwAAAAAAACJRILnf7Ax3APvapTeZQTkWKOBDpi3RdSHKwPmm0Ts+VOa6AAAAAAABASsAAAAAAAAAACJRILnf7Ax3APvapTeZQTkWKOBDpi3RdSHKwPmm2Ds+VObKAQMEAQAAAAABASvoAwAAAAAAACJRILnf7Ax3APvapTeZQTkWKOBDpi3RdSHKwPmm2Ds+VObKAQMEAQAAAAABASushgEAAAAAACJRIJj4FAjEAkR9IYcaA8B99f3Q/Ov7/wgh3Ph5WDzWUDL6AQMEAQAAAAAA"
```
altafan: this can be reverted and we just add a dummy variable `proofTxid`

internal/core/domain/round.go
```diff
@@ -108,6 +108,8 @@ func (r *Round) RegisterIntents(intents []Intent) ([]Event, error) {
 		return nil, fmt.Errorf("missing intents to register")
 	}
 	for _, intent := range intents {
+		txid, _ := deriveTxidFromProof(intent.Proof)
+		intent.Txid = txid
```
altafan: instead of this, let's ensure in `intent.validate` that it has a non-empty Txid

internal/core/domain/round_repo.go
```diff
@@ -24,6 +24,7 @@ type RoundRepository interface {
 	GetSweptRoundsConnectorAddress(ctx context.Context) ([]string, error)
 	GetTxsWithTxids(ctx context.Context, txids []string) ([]string, error)
 	GetRoundsWithCommitmentTxids(ctx context.Context, txids []string) (map[string]any, error)
+	GetIntentByTxid(ctx context.Context, txid string) (Intent, error)
```
altafan:
```suggestion
	GetIntentByTxid(ctx context.Context, txid string) (*Intent, error)
```

internal/core/domain/intent_test.go
bitcoin-coder-bob: we need a real proof here so that when the code derives a txid from the proof, it actually can derive the txid. Using a dummy string here will not yield a txid when derived and will error

internal/core/domain/round_test.go
```diff
@@ -13,7 +13,7 @@ var (
 	intents = []domain.Intent{
 		{
 			Id:      "0",
-			Proof:   "proof",
+			Proof:   "cHNidP8BAGYCAAAAAv6deG1s5w2msi0rziQ2HoNoxC2onFnWf6XLi36HIUWbAAAAAAAAAAAA9PRZ58PC4lJAJ7VEYtMMUC1E+poh/Gxo+0TNqvgh27wAAAAAAAAAAAABAAAAAAAAAAABagAAAAAAAQErAAAAAAAAAAAiUSC53+wMdwD72qU3mUE5FijgQ6Yt0XUhysD5ptg7PlTmygEDBAEAAAAAAQEr6AMAAAAAAAAiUSC53+wMdwD72qU3mUE5FijgQ6Yt0XUhysD5ptg7PlTmygEDBAEAAAAAAA==",
```
altafan: same for these fixtures, we don't need to change them, rather we use the dummy value `Txid: proofTxid`

internal/core/domain/round.go
bitcoin-coder-bob: the intent.validate fxn has this check

internal/infrastructure/db/badger/ark_repo.go
```diff
@@ -252,6 +259,21 @@ func (r *arkRepository) findRound(
...
+		txid, err := application.DeriveTxidFromProof(v.Proof)
+		if err != nil {
+			fmt.Printf("error deriving txid from proof: %s\n", err.Error())
+			continue
+		}
+		updatedIntent := v
+		updatedIntent.Txid = txid
+		round.Intents[k] = updatedIntent
+	}
+
```
bitcoin-coder-bob: tests indicate failures without this as intents were getting added that did not have the txid

bitcoin-coder-bob: even with updated tests this would ensure that the added intent have the txid field populated in each intent

internal/infrastructure/db/badger/ark_repo.go
```diff
@@ -526,3 +557,57 @@ func (r arkRepository) findOffchainTxs(ctx context.Context, txids []string) ([]s
 	}
 	return txs, nil
 }
+
+func (r arkRepository) GetIntentByTxid(ctx context.Context, txid string) (domain.Intent, error) {
+	idx, err := r.getIntentIndexByTxid(ctx, txid)
+	if err != nil {
+		return domain.Intent{}, err
```
altafan: let's make sure to always return `nil` if there's an error.

I saw we added an ad hoc error in pkg/errors which is perfect. To easily recognize when we have to return such error, let's make sure this api returns `nil, nil` in case the intent is not found.

This way, at interface layer, we can do something like (pseudo-code):
```
intent, err := appSvc.GetIntentWithTxid(ctx, txid)
if err != nil { 
   return nil, err
}
if intent == nil {
    return nil, errors.INTENT_NOT_FOUND
}
```

internal/infrastructure/db/badger/ark_repo.go
```diff
@@ -252,6 +259,21 @@ func (r *arkRepository) findRound(
...
+	// derive txids from intent proofs and add to each intents
+	for k, v := range round.Intents {
+		if v.Txid != "" {
+			continue
+		}
+		txid, err := application.DeriveTxidFromProof(v.Proof)
+		if err != nil {
+			fmt.Printf("error deriving txid from proof: %s\n", err.Error())
+			continue
+		}
```
altafan: no let's just assume the txid is already there, this part can be dropped
```suggestion
```

internal/infrastructure/db/badger/ark_repo.go
```diff
@@ -291,6 +313,15 @@ func (r *arkRepository) addOrUpdateRound(
 		}
 		return err
 	}
+	// upsert intent index for each intent with a txid
+	for _, it := range rnd.Intents {
+		if it.Txid == "" {
+			continue
+		}
```
altafan: same here, let's assume it's there (as it will from now on)

internal/infrastructure/db/postgres/intent_txid_migration.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	// do we even need since we keep the intent id unchanged?
+	if err := fixReceiverTableFK(ctx, dbh); err != nil {
+		return fmt.Errorf("failed to fix receiver table foreign keys: %s", err)
+	}
+	// do we even need since we keep the intent id unchanged?
+	if err := fixVtxoTableFK(ctx, dbh); err != nil {
+		return fmt.Errorf("failed to fix vtxo table foreign keys: %s", err)
+	}
```
altafan: it seems overkill to me? why we need to "swap"? can't we just add the txid column to the intent table, select all records of this table, extract the txid from the proof and update those records?

internal/infrastructure/db/postgres/migration/20260114000000_add_intent_txid.up.sql
```diff
@@ -0,0 +1,6 @@
+-- update intent table to have a new column txid, make int unique and and not null
+ALTER TABLE intent
+ADD COLUMN txid TEXT;
+
+-- create an index on the new txid column for faster lookups
+CREATE INDEX IF NOT EXISTS idx_intent_txid ON intent(txid);
```
altafan: we have this index created here but also in the migration logic BackfillIntentTxd... better to create the index AFTER all intent records have been updated and drop this?

internal/infrastructure/db/postgres/round_repo.go
```diff
@@ -165,13 +166,18 @@ func (r *roundRepository) AddOrUpdateRound(ctx context.Context, round domain.Rou
 
 		if len(round.Intents) > 0 {
 			for _, intent := range round.Intents {
+				txid, err := application.DeriveTxidFromProof(intent.Proof)
+				if err != nil {
+					return fmt.Errorf("failed to derive txid from proof: %w", err)
+				}
```
altafan: drop

internal/infrastructure/db/postgres/round_repo.go
```diff
@@ -165,13 +166,18 @@ func (r *roundRepository) AddOrUpdateRound(ctx context.Context, round domain.Rou
...
 				if err := querierWithTx.UpsertIntent(
 					ctx,
 					queries.UpsertIntentParams{
 						ID:      sql.NullString{String: intent.Id, Valid: true},
 						RoundID: sql.NullString{String: round.Id, Valid: true},
 						Proof:   sql.NullString{String: intent.Proof, Valid: true},
 						Message: sql.NullString{String: intent.Message, Valid: true},
+						Txid:    txid,
```
altafan:
```suggestion
						Txid:    intent.Txid,
```

internal/infrastructure/db/postgres/round_repo.go
```diff
@@ -422,6 +428,20 @@ func (r *roundRepository) GetRoundsWithCommitmentTxids(
...
 }
 
+func (r *roundRepository) GetIntentByTxid(
+	ctx context.Context,
+	txid string,
+) (domain.Intent, error) {
+	intent, err := r.querier.SelectIntentByTxid(ctx, txid)
+	if err != nil {
+		return domain.Intent{}, fmt.Errorf("failed to get intents by txid: %w", err)
+	}
```
altafan, line 436:
```suggestion
	if err != nil {
	    if errors.Is(err, sql.ErrNoRows) {
	       return nil, nil
	    }
		return domain.Intent{}, fmt.Errorf("failed to get intents by txid: %w", err)
	}
```

internal/infrastructure/db/service_test.go
```diff
@@ -37,6 +38,8 @@ const (
 	arkTxid   = txida
 	sweepTxid = "ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss"
 	sweepTx   = "cHNidP8BADwBAAAAAauqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqAAAAAAD/////AegDAAAAAAAAAAAAAAAAAAA="
+	proof1    = "cHNidP8BALICAAAAAv/d70xPgEdsEq2xvJnXiBtJrEQAkZYVSWdwarEJCZe9AAAAAAAAAAAA9PRZ58PC4lJAJ7VEYtMMUC1E+poh/Gxo+0TNqvgh27wAAAAAAAAAAAAC6AMAAAAAAAAiUSC53+wMdwD72qU3mUE5FijgQ6Yt0XUhysD5ptE7PlTmuqyGAQAAAAAAIlEgmPgUCMQCRH0hhxoDwH31/dD86/v/CCHc+HlYPNZQMvoAAAAAAAEBKwAAAAAAAAAAIlEgud/sDHcA+9qlN5lBORYo4EOmLdF1IcrA+abYOz5U5soBAwQBAAAAAAEBK+gDAAAAAAAAIlEgud/sDHcA+9qlN5lBORYo4EOmLdF1IcrA+abYOz5U5soBAwQBAAAAAAAA"
+	proof2    = "cHNidP8BANsCAAAAA//d70xPgEdsEq2xvJnXiBtJrEQAkZYVSWdwarEJCZe9AAAAAAAAAAAA9PRZ58PC4lJAJ7VEYtMMUC1E+poh/Gxo+0TNqvgh27wAAAAAAAAAAABRRy0Q2j/vH4YYRfWfzzF4LBB1xX+tQdFOIWTU6ZD0CwMAAAAAAAAAAALoAwAAAAAAACJRILnf7Ax3APvapTeZQTkWKOBDpi3RdSHKwPmm0Ts+VOa6rIYBAAAAAAAiUSCY+BQIxAJEfSGHGgPAffX90Pzr+/8IIdz4eVg81lAy+gAAAAAAAQErAAAAAAAAAAAiUSC53+wMdwD72qU3mUE5FijgQ6Yt0XUhysD5ptg7PlTmygEDBAEAAAAAAQEr6AMAAAAAAAAiUSC53+wMdwD72qU3mUE5FijgQ6Yt0XUhysD5ptg7PlTmygEDBAEAAAAAAQErrIYBAAAAAAAiUSCY+BQIxAJEfSGHGgPAffX90Pzr+/8IIdz4eVg81lAy+gEDBAEAAAAAAAA="
```
altafan: revert

internal/infrastructure/db/service_test.go
```diff
@@ -430,7 +433,7 @@ func testRoundRepository(t *testing.T, svc ports.RepoManager) {
 				Intents: []domain.Intent{
 					{
 						Id:      uuid.New().String(),
-						Proof:   "proof",
+						Proof:   proof1,
```
altafan: revert

internal/infrastructure/db/service_test.go
```diff
@@ -450,7 +453,7 @@ func testRoundRepository(t *testing.T, svc ports.RepoManager) {
 					},
 					{
 						Id:      uuid.New().String(),
-						Proof:   "proof",
+						Proof:   proof2,
```
altafan: revert

internal/infrastructure/db/service_test.go
altafan: use dummy values instead for the txid and make use of them in the test below instead of invoking application.DeriveTxidFromProof

internal/core/application/utils.go
```diff
@@ -465,6 +465,14 @@ func waitForConfirmation(
 	}
 }
 
+func DeriveTxidFromProof(proof string) (string, error) {
+	tx, err := psbt.NewFromRawBytes(strings.NewReader(proof), true)
+	if err != nil {
+		return "", fmt.Errorf("error deriving txid from proof: %w", err)
+	}
+	return tx.UnsignedTx.TxID(), nil
+}
```
altafan: revert the changes to this file, no changes are expected here

bitcoin-coder-bob: i have multiple files depending on this util function. Where shall I put it if not here? The files that call it are currently:

`internal/core/application/service.go`
`internal/core/domain/intent_test.go`
`internal/infrastructure/db/sqlite/round_repo.go`
`internal/infrastructure/db/service_test.go`

internal/core/domain/intent_test.go
bitcoin-coder-bob: removing it. i'll grab the txid from this proof (`b17d6142a543b01aaff6063e7f4351932aaf81053e9ef0d707c332a57aff4544`) and keep the dummy proof

bitcoin-coder-bob: or rather i'll fallback to the txid var we already have set which is all zeroes

internal/core/domain/round.go
bitcoin-coder-bob: i will remove these 2 lines, and assume the intent passed has the txid, and if not the validate will catch it

internal/core/application/utils.go
bitcoin-coder-bob: update: only being used in `internal/core/application/service.go` now

internal/infrastructure/db/postgres/intent_txid_migration.go
bitcoin-coder-bob: ah i think this works only when we add a nullable field to the schema for the txid (which we are) and that this extra complexity was when I had decided to make the new txid field `NOT NULL`. I will make the simplifying changes

internal/core/domain/intent_test.go
```diff
@@ -28,7 +28,7 @@ var (
 func TestIntent(t *testing.T) {
 	t.Run("new_intent", func(t *testing.T) {
 		t.Run("valid", func(t *testing.T) {
-			intent, err := domain.NewIntent(proof, message, inputs)
+			intent, err := domain.NewIntent(txid, proof, message, inputs)
```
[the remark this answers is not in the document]
bitcoin-coder-bob: this `txid` comes from `round_test.go` so it is declared

internal/infrastructure/db/sqlite/round_repo.go
```diff
@@ -453,6 +454,24 @@ func (r *roundRepository) GetRoundsWithCommitmentTxids(
...
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get intent by txid: %w", err)
+	}
+
+	return &domain.Intent{
+		Proof:   intent.Proof.String,
+		Message: intent.Message.String,
+	}, nil
```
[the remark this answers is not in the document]
bitcoin-coder-bob: requires the query to be changed which is currently only yielding a proof and message. I dont think we need to return the id and txid. The caller will have the txid on hand already. We could have it return the id if that is useful, but I suspect the purpose of calling `GetIntentByTxid` is to grab the proof and/or message which we are currently returning.

internal/infrastructure/db/service.go
```diff
@@ -231,6 +231,30 @@ func NewService(config ServiceConfig, txDecoder ports.TxDecoder) (ports.RepoMana
...
+			if err := m.Migrate(intentTxidMigrationBegin); err != nil &&
+				!errors.Is(err, migrate.ErrNoChange) {
+				return nil, fmt.Errorf("failed to run migrations: %s", err)
+			}
+
+			err = pgdb.BackfillIntentTxid(context.Background(), db)
+			if err != nil {
+				return nil, fmt.Errorf("failed to backfill intent txid field: %w", err)
+			}
+		}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: will leave as a TODO if we want that

internal/infrastructure/db/postgres/intent_txid_migration_test.go
```diff
@@ -0,0 +1,141 @@
...
+	var exists bool
+	err = db.QueryRow(`
+        SELECT EXISTS (
+          SELECT 1
+          FROM information_schema.columns
+          WHERE table_schema = 'public'
+            AND table_name = 'intent'
+            AND column_name = 'txid'
+        );
+    `).Scan(&exists)
```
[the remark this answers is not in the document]
bitcoin-coder-bob: it does work for local devs

internal/core/application/service.go
```diff
@@ -1697,8 +1697,14 @@ func (s *service) RegisterIntent(
...
 	}
-
-	intent, err := domain.NewIntent(signedProof, encodedMessage, vtxoInputs)
+	txid, err := DeriveTxidFromProof(signedProof)
+	if err != nil {
+		return "", errors.INTERNAL_ERROR.New("failed to derive txid from proof: %w", err).
+			WithMetadata(map[string]any{
+				"proof": signedProof,
+			})
+	}
```
altafan: not needed at all, we can use `proofTxid` declared at L#1470.. this way we can completely drop DeriveTxidFromProof()

internal/infrastructure/db/postgres/intent_txid_migration.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	if err = tx.Commit(); err != nil {
+		return fmt.Errorf("commit transaction: %w", err)
+	}
+	return nil
+}
+
+func DeriveTxidFromProof(proof string) (string, error) {
```
altafan: no need to be exported
```suggestion
func deriveTxidFromProof(proof string) (string, error) {
```

internal/infrastructure/db/postgres/intent_txid_migration_test.go
```diff
@@ -0,0 +1,141 @@
+package pgdb
```
altafan: The convention is to declare this a test package (and this is the only reason why the backfill function is exported)

```suggestion
package pgdb_test
```

internal/infrastructure/db/postgres/round_repo.go
```diff
@@ -422,6 +423,23 @@ func (r *roundRepository) GetRoundsWithCommitmentTxids(
...
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get intent by txid: %w", err)
+	}
+
+	return &domain.Intent{
+		Proof:   intent.Proof.String,
+		Message: intent.Message.String,
+	}, nil
```
altafan, line 443:
```suggestion
	return &domain.Intent{
	    Txid: intent.Txid.String,
		Proof:   intent.Proof.String,
		Message: intent.Message.String,
	}, nil
```

internal/infrastructure/db/service.go
altafan: seems like this code is duplicated and can be moved to a util function `handleIntentIdMigration` (or smtn lke that)?

internal/infrastructure/db/service_test.go
```diff
@@ -502,6 +504,17 @@ func testRoundRepository(t *testing.T, svc ports.RepoManager) {
...
+		intent, err := svc.Rounds().GetIntentByTxid(ctx, txida)
+		require.NoError(t, err)
+		require.Equal(t, "proof", intent.Proof)
+		require.Equal(t, "message", intent.Message)
+
+		intent, err = svc.Rounds().GetIntentByTxid(ctx, txidb)
+		require.NoError(t, err)
+		require.Equal(t, "proof", intent.Proof)
+		require.Equal(t, "message", intent.Message)
+
```
altafan, line 527: let's add also a test we get the expected behavior for non-existing records in db

internal/infrastructure/db/sqlite/intent_txid_migration_test.go
```diff
@@ -0,0 +1,135 @@
+package sqlitedb
```
altafan:
```suggestion
package sqlitedb_test
```

internal/infrastructure/db/postgres/round_repo.go
bitcoin-coder-bob: coderabbitai suggested this as well: https://github.com/arkade-os/arkd/pull/884#discussion_r2705616768

I will modify the query so we can populate txid here

internal/infrastructure/db/sqlite/round_repo.go
altafan, line 474:
```suggestion
	return &domain.Intent{
	    Txid: intent.Txid.String,
		Proof:   intent.Proof.String,
		Message: intent.Message.String,
	}, nil
```

bitcoin-coder-bob: also added in the id field: [27c13e0](https://github.com/arkade-os/arkd/pull/884/commits/27c13e0c61f8d7b7e516084805b62f76dffadf97)

internal/infrastructure/db/sqlite/intent_txid_migration_test.go
bitcoin-coder-bob: handled in: [27c13e0](https://github.com/arkade-os/arkd/pull/884/commits/27c13e0c61f8d7b7e516084805b62f76dffadf97)

internal/infrastructure/db/service_test.go
bitcoin-coder-bob: handled in: [27c13e0](https://github.com/arkade-os/arkd/pull/884/commits/27c13e0c61f8d7b7e516084805b62f76dffadf97)

internal/infrastructure/db/service.go
bitcoin-coder-bob: handled in: [27c13e0](https://github.com/arkade-os/arkd/pull/884/commits/27c13e0c61f8d7b7e516084805b62f76dffadf97)

internal/infrastructure/db/postgres/round_repo.go
bitcoin-coder-bob: handled in: [27c13e0](https://github.com/arkade-os/arkd/pull/884/commits/27c13e0c61f8d7b7e516084805b62f76dffadf97)

internal/infrastructure/db/postgres/intent_txid_migration_test.go
bitcoin-coder-bob: handled in: [27c13e0](https://github.com/arkade-os/arkd/pull/884/commits/27c13e0c61f8d7b7e516084805b62f76dffadf97)

internal/infrastructure/db/postgres/intent_txid_migration.go
bitcoin-coder-bob: handled in: [27c13e0](https://github.com/arkade-os/arkd/pull/884/commits/27c13e0c61f8d7b7e516084805b62f76dffadf97)

internal/core/application/service.go
bitcoin-coder-bob: handled in: [9cfbb83](https://github.com/arkade-os/arkd/pull/884/commits/9cfbb83448038839a29634b72bd9c40b6bfcb727)

internal/infrastructure/db/postgres/intent_txid_migration_test.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+func TestIntentTxidMigration(t *testing.T) {
+	ctx := context.Background()
+	dsn := "postgres://root:secret@localhost:5432/event?sslmode=disable"
+	db, err := sql.Open("postgres", dsn)
+	require.NoError(t, err)
+
+	// nolint:errcheck
+	defer db.Close()
```
altafan: Use t.Cleanup to make sure the function is called once the test ends. (same for sqlite)

```suggestion
	t.Cleanup(func() {
	    db.Close()
	})
```

internal/infrastructure/db/service.go
```diff
@@ -609,3 +620,39 @@ func newBadgerRoundRepository(args ...interface{}) (domain.RoundRepository, erro
 func newBadgerOffchainTxRepository(args ...interface{}) (domain.OffchainTxRepository, error) {
 	return initBadgerArkRepository(args...)
 }
+
+// stepwise migration for intent txid field addition
+func HandleIntentTxidMigration(m *migrate.Migrate, db *sql.DB, dbType string) error {
```
altafan: no need to be an exported function as we're not going to reuse it outside this folder (same for sqlite)

```suggestion
func handleIntentTxidMigration(m *migrate.Migrate, db *sql.DB, dbType string) error {
```

bitcoin-coder-bob: fixed in: [79fe6db](https://github.com/arkade-os/arkd/pull/884/commits/79fe6db953707b591fbfd981353d97b4d7d6e737)

internal/infrastructure/db/postgres/intent_txid_migration_test.go
bitcoin-coder-bob: fixed in: [79fe6db](https://github.com/arkade-os/arkd/pull/884/commits/79fe6db953707b591fbfd981353d97b4d7d6e737)

PR #890, opened 2026-01-27
Get pending intents by proof of ownership

louisinger:
Modify `GetIntentResponse` to return a list instead of a single intent and add a new `intent` filter. If the proof is valid and signs a `get-intent` message, the server returns the list of pending intents in the queue matching at least one of the signed input.

@bitcoin-coder-bob @altafan please review

## Release Notes

**New Features**
- GetIntent endpoint now supports filtering by intent proof and message parameters in addition to transaction ID, providing alternative lookup methods
- GetIntent endpoint returns multiple matching intents in a single response instead of a single intent, enabling efficient retrieval of related intents

### DIFF
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -4300,11 +4300,7 @@ func (s *service) GetIntentsByProof(
 		}
 	}
 
-	if len(inputsMap) > 0 {
-		return nil, errors.INVALID_INTENT_PROOF.New("some inputs are not matched by any intent")
-	}
 	return matchingIntents, nil
-
 }
 
 func extractVtxoScriptFromSignedForfeitTx(tx string) (string, error) {
--- a/api-spec/protobuf/ark/v1/service.proto
+++ b/api-spec/protobuf/ark/v1/service.proto
@@ -162,7 +162,9 @@ service ArkService {
     };
   }
 
-  // GetIntent returns a list of intents matching the given txid or spending one of the given proof inputs.
+  // GetIntent returns a list of intents depending on the filter provided:
+  // by txid: the server will return intent from completed batches
+  // by proof: the server will return any pending intents matching the proof inputs
   rpc GetIntent(GetIntentRequest) returns (GetIntentResponse) {
     option (meshapi.gateway.http) = {
       get: "/v1/intent"
--- a/go.mod
+++ b/go.mod
@@ -21,13 +21,13 @@ replace github.com/arkade-os/arkd/pkg/errors => ./pkg/errors
 require (
 	github.com/ThreeDotsLabs/watermill-sql/v3 v3.1.0
 	github.com/arkade-os/arkd/api-spec v0.0.0-00010101000000-000000000000
-	github.com/arkade-os/arkd/pkg/ark-lib v0.8.1-0.20260102141947-c66f85422174
+	github.com/arkade-os/arkd/pkg/ark-lib v0.8.1-0.20260127162821-ef2685c027f9
 	github.com/arkade-os/arkd/pkg/arkd-wallet v0.0.0-00010101000000-000000000000
 	github.com/arkade-os/arkd/pkg/arkd-wallet-btcwallet v0.0.0-00010101000000-000000000000
 	github.com/arkade-os/arkd/pkg/errors v0.0.0-00010101000000-000000000000
 	github.com/arkade-os/arkd/pkg/kvdb v0.0.0-20250606113434-241d3e1ec7cb
 	github.com/arkade-os/arkd/pkg/macaroons v0.0.0-00010101000000-000000000000
-	github.com/arkade-os/go-sdk v0.8.2-0.20260121032947-a50889351829
+	github.com/arkade-os/go-sdk v0.8.2-0.20260127164301-6e1888fba7da
 	github.com/btcsuite/btcd/btcec/v2 v2.3.5
 	github.com/dgraph-io/badger/v4 v4.8.0
 	github.com/docker/docker v27.1.1+incompatible
--- a/go.sum
+++ b/go.sum
@@ -30,8 +30,8 @@ github.com/aead/siphash v1.0.1/go.mod h1:Nywa3cDsYNNK3gaciGTWPwHt0wlpNV15vwmswBA
 github.com/antihax/optional v1.0.0/go.mod h1:uupD/76wgC+ih3iEmQUL+0Ugr19nfwCT1kdvxnR2qWY=
 github.com/antlr4-go/antlr/v4 v4.13.0 h1:lxCg3LAv+EUK6t1i0y1V6/SLeUi0eKEKdhQAlS8TVTI=
 github.com/antlr4-go/antlr/v4 v4.13.0/go.mod h1:pfChB/xh/Unjila75QW7+VU4TSnWnnk9UTnmpPaOR2g=
-github.com/arkade-os/go-sdk v0.8.2-0.20260121032947-a50889351829 h1:PbocI9n7Olv1bTmlEjMAW6e1Ko/bB+qpcgAHDWaZMN0=
-github.com/arkade-os/go-sdk v0.8.2-0.20260121032947-a50889351829/go.mod h1:ckbpT6QmWDEdlHGgo60XGBqwbFSWluT4pMLBiM7WfbU=
+github.com/arkade-os/go-sdk v0.8.2-0.20260127164301-6e1888fba7da h1:iqulYGiT106QXCI27e0nLAiZ6jvOAT1K3U51UDx8WO0=
+github.com/arkade-os/go-sdk v0.8.2-0.20260127164301-6e1888fba7da/go.mod h1:KxyBigM73fzw6VK8u775GQywlZ6hIcVJrKxFtw2fJFQ=
 github.com/armon/consul-api v0.0.0-20180202201655-eb2c6b5be1b6/go.mod h1:grANhF5doyWs3UAsr3K4I6qtAmlQcZDesFNEHPZAzj8=
 github.com/benbjohnson/clock v1.1.0/go.mod h1:J11/hYXuz8f4ySSvYwY0FKfm+ezbsZBKZxNJlLklBHA=
 github.com/beorn7/perks v1.0.1 h1:VlbKKnNfV8bJzeqoa4cOKqO6bYr3WgKZxO8Z16+hsOM=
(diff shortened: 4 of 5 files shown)


pkg/ark-lib/intent/message.go
```diff
@@ -134,6 +135,34 @@ func (m *GetPendingTxMessage) Decode(data string) error {
 
 	return nil
 }
+
+type GetIntentMessage struct {
+	BaseMessage
+	// ExpireAt is the timestamp (in seconds) at which the proof should be considered invalid
```
bitcoin-coder-bob, line 141: is this a unix timestamp?

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
 
+func (s *service) GetIntentsByProof(
+	ctx context.Context,
+	proof intent.Proof,
+	message intent.GetIntentMessage,
+) ([]domain.Intent, errors.Error) {
+	// verify the proof of ownership
+	if message.ExpireAt > 0 {
+		expireAt := time.Unix(message.ExpireAt, 0)
+		if time.Now().After(expireAt) {
```
bitcoin-coder-bob, line 4126: here you use `time.Now()` but a few lines down you use `time.Now().Unix()`. I reckon we want the `time.Now().Unix()` usage here as well

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+				WithMetadata(errors.IntentTimeRangeMetadata{
+					ValidAt:  0,
+					ExpireAt: message.ExpireAt,
+					Now:      time.Now().Unix(),
+				})
+		}
+	}
+
+	outpoints := proof.GetOutpoints()
+	proofTxid := proof.UnsignedTx.TxID()
```
bitcoin-coder-bob, line 4137: `proof.UnsignedTx` could be nil, if so I believe this would runtime error

internal/core/application/service.go
altafan: time.Unix is used to parse a int64 into time.Time
later, we use time.Now().Unix() to do the opposite and parse a time.Time to int64

altafan: @louisinger store time.Now() in a variable so we're sure to use the same value for the check and the one included in the metadata errs

internal/core/application/service.go
altafan: theoretically possible, practically impossible, as we have validation checks a layer above (interface)

internal/interface/grpc/handlers/arkservice.go
```diff
@@ -664,28 +664,39 @@ func (h *handler) listenToTxEvents() {
...
 ) (*arkv1.GetIntentResponse, error) {
-	var err error
-	var intent *domain.Intent
+	intents := make([]domain.Intent, 0)
 
 	switch filter := req.GetFilter().(type) {
 	case *arkv1.GetIntentRequest_Txid:
-		intent, err = h.svc.GetIntentByTxid(ctx, filter.Txid)
-
+		intent, err := h.svc.GetIntentByTxid(ctx, filter.Txid)
```
altafan, line 671: add validation of the txid, make sure it can be parsed with chainhash

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+
+	boardingTxs := make(map[string]wire.MsgTx)
+	for i, outpoint := range outpoints {
+		psbtInput := proof.Inputs[i+1]
```
bitcoin-coder-bob, line 4141: can you help me understand or document in a comment why we offset by 1?

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+
+		vtxoOutpoint := domain.Outpoint{
+			Txid: outpoint.Hash.String(),
+			VOut: outpoint.Index,
+		}
+
+		vtxosResult, err := s.repoManager.Vtxos().GetVtxos(ctx, []domain.Outpoint{vtxoOutpoint})
+		if err != nil || len(vtxosResult) == 0 {
+			if _, ok := boardingTxs[vtxoOutpoint.Txid]; !ok {
+				txhex, err := s.wallet.GetTransaction(ctx, outpoint.Hash.String())
```
bitcoin-coder-bob, line 4156:
```suggestion
				txHex, err := s.wallet.GetTransaction(ctx, outpoint.Hash.String())
```

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+			tx := boardingTxs[vtxoOutpoint.Txid]
+			if int(vtxoOutpoint.VOut) >= len(tx.TxOut) {
+				return nil, errors.INVALID_PSBT_INPUT.New(
+					"invalid vout index %d for tx %s (tx has %d outputs)",
+					vtxoOutpoint.VOut, vtxoOutpoint.Txid, len(tx.TxOut),
+				).WithMetadata(errors.InputMetadata{Txid: proofTxid, InputIndex: i + 1})
+			}
+			prevout := tx.TxOut[vtxoOutpoint.VOut]
+
+			if !bytes.Equal(prevout.PkScript, psbtInput.WitnessUtxo.PkScript) {
```
bitcoin-coder-bob, line 4182: `prevout` could be nil yielding a runtime error here

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+				return nil, errors.INVALID_PSBT_INPUT.New(
+					"pkscript mismatch: got %x expected %x",
+					prevout.PkScript,
+					psbtInput.WitnessUtxo.PkScript,
+				).WithMetadata(errors.InputMetadata{Txid: proofTxid, InputIndex: i + 1})
+			}
+
+			if prevout.Value != int64(psbtInput.WitnessUtxo.Value) {
```
bitcoin-coder-bob, line 4190:
```suggestion
			if prevout.Value != psbtInput.WitnessUtxo.Value {
```

since it is already of type int64

internal/core/application/service.go
bitcoin-coder-bob, line 4141: i see `i + 1` usage 8 times in this loop, may make sense to define it as a named variable

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+				return nil, errors.INVALID_PSBT_INPUT.New(
+					"invalid witness utxo value: got %d expected %d",
```
bitcoin-coder-bob, line 4192: this error log is the same as one on line 4203 just below, we should have them be different so it is clear where the error occurred

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+	signedProof, err := s.signer.SignTransactionTapscript(ctx, encodedProof, nil)
+	if err != nil {
+		return nil, errors.INTERNAL_ERROR.New("failed to sign proof: %w", err).
+			WithMetadata(map[string]any{
+				"proof": proof.UnsignedTx.TxID(),
+			})
+	}
+
+	if err := intent.Verify(signedProof, encodedMessage); err != nil {
+		log.
```
bitcoin-coder-bob, line 4266: leftover log statement we can remove?

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+	}
+
+	// proof is valid, we can retrieve the intents matching the proof's inputs
+
+	allIntents, err := s.cache.Intents().ViewAll(ctx, nil)
+	if err != nil {
+		return nil, errors.INTERNAL_ERROR.New("failed to view all intents: %w", err)
+	}
+
+	inputsMap := make(map[domain.Outpoint]struct{})
```
bitcoin-coder-bob, line 4285:
```suggestion
	inputsMap := make(map[domain.Outpoint]struct{}, len(outpoints))
```

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+			return nil, errors.INVALID_INTENT_TIMERANGE.New("proof of ownership expired").
+				WithMetadata(errors.IntentTimeRangeMetadata{
+					ValidAt:  0,
+					ExpireAt: message.ExpireAt,
+					Now:      time.Now().Unix(),
+				})
+		}
+	}
+
+	outpoints := proof.GetOutpoints()
```
bitcoin-coder-bob, line 4136: if this is nil should we return right away? That is, if the length of the proof's `UnsignedTx.TxIn <= 1` we will get nil here, and wont have outpoints to iterate over in the loop shortly used after this line,  and then in the loop on line 4286 as well

internal/core/application/service.go
altafan: because in the intent tx the very first input is the one that makes the tx invalid with outpoint {txid: messageHash, Vout: 0}

internal/core/application/service.go
altafan: the check before this one makes sure the vtxo outpoint is not out of range, therefore `prevout` cannot be empty

internal/core/application/service.go
```diff
@@ -4115,6 +4115,194 @@ func (s *service) GetIntentByTxid(
...
+				"proof": proof.UnsignedTx.TxID(),
+			})
+	}
+
+	if err := intent.Verify(signedProof, encodedMessage); err != nil {
+		log.
+			WithField("unsignedProof", encodedProof).
+			WithField("signedProof", signedProof).
+			WithField("encodedMessage", encodedMessage).
+			Tracef("failed to verify intent proof: %s", err)
```
altafan, line 4270: let's not use trace log level pls.. if you think it's useful to log this err, use debug

internal/core/application/service.go
altafan: not needed for maps

internal/core/application/service.go
louisinger: it's a copy paste of what we do in other intent validation functions

louisinger:
closed in favor of https://github.com/arkade-os/arkd/pull/917

PR #896, opened 2026-02-02
Asset Transaction validation

louisinger:
@aruokhai @altafan please review


pkg/ark-lib/asset/asset_group.go
```diff
@@ -46,6 +47,27 @@ type AssetGroup struct {
...
+		outAmounts[i] = out.Amount
+	}
+	for i, in := range ag.Inputs {
+		inAmounts[i] = in.Amount
+	}
+	sumOutputs := safeSumUint64(outAmounts)
+	sumInputs := safeSumUint64(inAmounts)
+
+	return !ag.IsIssuance() && sumInputs.Cmp(sumOutputs) < 0
+}
```
altafan, line 71: in #897, i wrote the something similar (in db manager) but with more strict logic.

Maybe these methods can be exported for the packet, but i'm not really conviced, as they should become `ContainsIssuances` and `ContainsReissuances`... seems like getting back to go-elements @louisinger 🤣. I wouldn't go for this option and just keep things as they are in ark-asset unless we find something else

pkg/ark-lib/asset/tx_validation.go
```diff
@@ -0,0 +1,301 @@
...
+
+// validateIssuance validates the control asset of an issuance asset group
+// if it's present and referenced by group, it must be issued in the same transaction
+func validateIssuance(packet Packet, grp AssetGroup) errors.Error {
+	if grp.ControlAsset == nil {
+		return nil
+	}
+
+	if grp.ControlAsset.Type == AssetRefByID {
+		// by id means the control asset is an existing asset, it must be present in the packet
```
altafan: check existance of control asset in db?

louisinger: done

PR #904, opened 2026-02-06
Require Intent for GetVtxoChain/GetVirtualTxs

### LINKED ISSUE #843: Require Intent for GetVtxoChain/GetVirtualTxs
Through a config option, make get vtxochain require an intent with every  vtxo it requests the chain for. 

the config should be a string: INDEXER_TX_EXPOSURe which would be
* PUBLIC, no intent required
* WITHELD, signatures of the operator stripped from every tx  returned, unless it provided
* PRIVATE, Intent is required for any data response

As GetVtxoChain only returns a map of ids and not actual data, we should have it generate a short lived "code" that GetVirtualTxs would use as authentication.

We also have these endpoints to think about:
https://github.com/arkade-os/arkd/blob/0c49bf7ec5c09843008b10c618ccd04a3f4588ca/api-spec/protobuf/ark/v1/indexer.proto#L16-L50

bitcoin-coder-bob:
Closes #843

The general flow is that a caller calls `GetVtxoChain` which yields `[]txids` which they then pass a param to a call to `GetVirtualTxs`. The idea of this PR is to create an auth token inside `GetVtxoChain` if the user can prove ownership of the intent they passed in, then to use this auth token when calling `GetVirtualTxs`. The spirit here is that: "A public virtual mempool means griefing attacks by third party is possible. Imagine you received a vtxo and now someone can just start unilaterally exiting your vtxo and you end up back onchain. Really annoying and can cause a number of problems with smart contracts."

This auth token is a message signed by the server's pubkey along with an outpoint, resulting in stateless auth system. 

We need go-sdk changes:
A new intent param must be passed to `GetVtxoChain` and the response will now include a new auth token (can be empty string as well). Depending on the new config for arkd for the `INDEXER_TX_EXPOSURE` field (with the 3 options being `public`, `private`, or `withheld`) will determine the creation and enforcement of this auth token.

We will also want tests against `GetVtxoChain` , the new `ValidateIntentWithProof` and `GetVirtualTxs`

When `public`: no auth token created, empty auth token in response of `GetVtxoChain` and an empty auth token to be passed to `GetVirtualTxs`.

When `private`: the caller must pass `ValidateIntentWithProof` to get an auth token, and the auth token is required to call `GetVirtualTxs`

When `withheld`:  the caller of `GetVtxoChain` has the option to pass OR fail `ValidateIntentWithProof` which will determine if they do or do not get an auth token. If they get a token they can use it when calling `GetVrtualTxs` and the behavior will function the same way as if `public` were set. However if they pass an invalid or empty auth token, we must remove from each PSBT the signature of arkd so the user cannot construct the full broadcastable txn. 

The newly added `ValidateIntentWithProof` is largely the same flow as found here: https://github.com/arkade-os/arkd/blob/368bb9fe924160cc06a5cc6b7a0fd0856c4a5bd6/internal/core/application/service.go#L2048-L2200

The `indexerService` now that it needs to be able to sign a message for the auth token, and needs access to the server's pubkey (to locate inside the psbt for removal), we have given the `indexerService` access to `ports.SignerService` and `ports.WalletService` (needed for the SignMessage call). 

Requires changes in the go-sdk for `GetVtxoChain` (which now requires an Intent be passed) and to `GetVirtualTxs` (which now requires an AuthToken to be passed). Thus the e2e tests without this sdk change will only work when the exposure type is set to `public`.

Note that theres a couple small changes in the sqlite. I re-ran `sqlc generate` and it yielded couple changes to the `After` field. 

## Summary by CodeRabbit

* **New Features**
  * Message signing capability added (signing endpoint and wallet support).
  * Intent-based access control for VTXO chain retrieval with configurable exposures (public, withheld, private).
  * APIs extended to accept/return intent and auth-token fields for controlled access to virtual transaction data.

* **Documentation**
  * Added ARKD_INDEXER_TX_EXPOSURE environment variable (default: public).

* **Tests**
  * New unit and e2e tests with fixtures for message signing and auth-token flows.


api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -203,15 +204,19 @@ message GetVtxosResponse {
 message GetVtxoChainRequest {
   IndexerOutpoint outpoint = 1;
   IndexerPageRequest page = 2;
+  Intent intent = 3;
 }
```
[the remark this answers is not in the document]
bitcoin-coder-bob: i dont think the intents are that big?

internal/core/application/indexer.go
```diff
@@ -362,15 +411,192 @@ func (i *indexerService) GetVtxoChain(
...
+	case TxExposureWithheld:
+		// validate the intent proof/message to allow access to the full chain
+		err = i.ValidateIntentWithProof(ctx, vtxoKey, intentForProof)
+		if err == nil {
+			authToken, err = i.createAuthToken(ctx, vtxoKey)
+			if err != nil {
+				return nil, err
+			}
+		}
+		// if intent failed validation, we just dont supply an auth token
```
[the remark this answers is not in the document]
bitcoin-coder-bob: we could require they always pass validation, not just use it as a gate to determine if they get an auth token

internal/core/application/indexer.go
```diff
@@ -362,15 +411,192 @@ func (i *indexerService) GetVtxoChain(
...
+				txhex := vtxoKey.Txid
+				var tx wire.MsgTx
+				if err := tx.Deserialize(hex.NewDecoder(strings.NewReader(txhex))); err != nil {
+					return errors.INVALID_PSBT_INPUT.New(
+						"failed to deserialize boarding tx %s: %s", vtxoOutpoint.Txid, err,
+					).WithMetadata(errors.InputMetadata{Txid: proofTxid, InputIndex: idx + 1})
+				}
+
+				boardingTxs[vtxoOutpoint.Txid] = tx
+			}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: ill have to give the indexerSerivce access to repo `ports.WalletService` so I can call it.

pkg/arkd-wallet/core/application/wallet/service.go
```diff
@@ -727,6 +728,16 @@ func (w *wallet) LoadSignerKey(ctx context.Context, prvkey *btcec.PrivateKey) er
...
 
+func (w *wallet) SignMessage(ctx context.Context, message []byte) ([]byte, error) {
+	if w.SignerKey == nil {
+		return nil, fmt.Errorf("signer key not loaded")
+	}
+
+	msgHash := chainhash.DoubleHashB(message)
+	sig := ecdsa.Sign(w.SignerKey, msgHash)
+	return sig.Serialize(), nil
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: should be addressed in: [9cec53b](https://github.com/arkade-os/arkd/pull/904/commits/9cec53bc705df8934f928c8b4a1db8198f0709bb)

internal/core/application/indexer.go
```diff
@@ -439,3 +722,65 @@ func paginate[T any](items []T, params *Page, maxSize int32) ([]T, PageResp) {
...
+		return false, nil // Invalid signature format
+	}
+
+	pubkey, err := schnorr.ParsePubKey(i.signerPubkey)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse signer pubkey: %w", err)
+	}
+
+	return sig.Verify(msgHash, pubkey), nil
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: should be addressed in: [9cec53b](https://github.com/arkade-os/arkd/pull/904/commits/9cec53bc705df8934f928c8b4a1db8198f0709bb)

internal/core/application/indexer.go
```diff
@@ -379,6 +608,66 @@ func (i *indexerService) GetVirtualTxs(
...
+				for j := range ptx.Inputs {
+					newSigs := make([]*psbt.PartialSig, 0)
+					for _, sig := range ptx.Inputs[j].PartialSigs {
+						// if the signature is not from arkd, keep it, otherwise remove it
+						if !bytes.Equal(sig.PubKey, i.signerPubkey) {
+							newSigs = append(newSigs, sig)
+						}
+					}
+					ptx.Inputs[j].PartialSigs = newSigs
+				}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: resolved in: [0c75ae7](https://github.com/arkade-os/arkd/pull/904/commits/0c75ae7fec157e9e39730ec448356988657eb9e4)

internal/core/application/indexer_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Create mock signer
+	mockSigner, err := newMockSignerService(f.AuthTokenTests.PrivateKeyHex)
+	require.NoError(t, err)
+
+	// Create indexer service
+	indexer := &indexerService{
+		signer:       mockSigner,
+		signerPubkey: schnorr.SerializePubKey(mockSigner.publicKey),
+	}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: resolved in: [0c75ae7](https://github.com/arkade-os/arkd/pull/904/commits/0c75ae7fec157e9e39730ec448356988657eb9e4)

internal/core/application/indexer.go
```diff
@@ -362,15 +413,193 @@ func (i *indexerService) GetVtxoChain(
...
+				WithMetadata(errors.InputMetadata{Txid: proofTxid, InputIndex: idx + 1})
+		}
+
+		vtxoOutpoint := domain.Outpoint{
+			Txid: outpoint.Hash.String(),
+			VOut: outpoint.Index,
+		}
+
+		vtxosResult, err := i.repoManager.Vtxos().GetVtxos(ctx, []domain.Outpoint{vtxoOutpoint})
+		if err != nil || len(vtxosResult) == 0 {
```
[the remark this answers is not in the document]
bitcoin-coder-bob: not that this wasnt overlooked but this it how it is handled in `DeleteIntentByProof` in `internal/core/application/service.go` which this fxn is based off of

internal/core/application/indexer.go
```diff
@@ -379,6 +608,66 @@ func (i *indexerService) GetVirtualTxs(
...
+						}
+					}
+					ptx.Inputs[j].TaprootScriptSpendSig = newSigs
+				}
+
+				var b strings.Builder
+				if err := ptx.Serialize(&b); err != nil {
+					return nil, fmt.Errorf("failed to serialize virtual tx: %s", err)
+				}
+				virtualTxs[idx] = b.String()
```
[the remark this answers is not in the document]
bitcoin-coder-bob: resolved in: [3fc2966](https://github.com/arkade-os/arkd/pull/904/commits/3fc29664c3caf3ea223accd7b4fa22d46a1b4b7a)

internal/core/application/indexer.go
```diff
@@ -362,15 +413,193 @@ func (i *indexerService) GetVtxoChain(
...
+
+			if !bytes.Equal(prevout.PkScript, psbtInput.WitnessUtxo.PkScript) {
+				return errors.INVALID_PSBT_INPUT.New(
+					"pkscript mismatch: got %x expected %x",
+					prevout.PkScript,
+					psbtInput.WitnessUtxo.PkScript,
+				).WithMetadata(errors.InputMetadata{Txid: proofTxid, InputIndex: idx + 1})
+			}
+
+			if prevout.Value != int64(psbtInput.WitnessUtxo.Value) {
```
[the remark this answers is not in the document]
bitcoin-coder-bob: resolved in: [989e0f4](https://github.com/arkade-os/arkd/pull/904/commits/989e0f4d5119ce0cb18c5783782aac2b74d7a028)

internal/core/application/indexer.go
```diff
@@ -362,22 +420,274 @@ func (i *indexerService) GetVtxoChain(
...
+			authToken, err = i.createAuthToken(ctx, vtxoKey)
+			if err != nil {
+				return nil, err
+			}
+		}
+		// if intent failed validation, we just dont supply an auth token
+
+	default:
+		return nil, fmt.Errorf("invalid exposure value: %s", i.txExposure)
+	}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: fixed in: [a0d6010](https://github.com/arkade-os/arkd/pull/904/commits/a0d60107dd80ccb9f0c42a6a49236595268d6058)

internal/core/application/indexer.go
```diff
@@ -379,22 +437,280 @@ func (i *indexerService) GetVtxoChain(
...
+			if TxExposure(i.txExposure) == TxExposurePrivate {
+				return nil, err
+			}
+			// withheld: swallow error, proceed without auth token
+		} else {
+			authToken, err = i.createAuthToken(ctx, vtxoKey)
+			if err != nil {
+				return nil, err
+			}
+		}
```
altafan: Maybe this is a bit cleaner?

```suggestion
	case TxExposureWithheld:
	    // validate the intent proof/message to allow access to the full chain
		if err = i.ValidateIntentWithProof(ctx, vtxoKey, intentForProof); err != nil {
				// withheld: swallow error, proceed without auth token
				break
			}
		}
		authToken, err = i.createAuthToken(ctx, vtxoKey)
		if err != nil {
			return nil, err
		}
	case TxExposurePrivate:
		// validate the intent proof/message to allow access to the full chain
		if err = i.ValidateIntentWithProof(ctx, vtxoKey, intentForProof); err != nil {
				return nil, err
			}
		}
		authToken, err = i.createAuthToken(ctx, vtxoKey)
		if err != nil {
			return nil, err
		}
```

api-spec/protobuf/signer/v1/service.proto
```diff
@@ -28,6 +28,8 @@ service SignerService {
       body: "*"
     };
   }
+  // SignMessage is gRPC-only (no HTTP gateway) to prevent unauthenticated access
```
louisinger: we can admit this is protected, I think it's ok to add http (I mean we expose SignTransaction so if HTTP is not safe we have bigger problems). need @altafan ACK btw

internal/core/application/indexer.go
```diff
@@ -456,3 +772,65 @@ func paginate[T any](items []T, params *Page, maxSize int32) ([]T, PageResp) {
...
+	token := make([]byte, len(msg)+len(sig))
+	copy(token[0:len(msg)], msg)
+	copy(token[len(msg):], sig)
+
+	return base64.StdEncoding.EncodeToString(token), nil
+}
+
+// validateAuthToken validates a signed auth token.
+// Returns true if the signature is valid.
+func (i *indexerService) validateAuthToken(authToken string) (bool, error) {
```
altafan: shall we check that the outpoint txid matches the one passed in the request of GetVirtualTxs?
Also, maybe we should add an invalidation timeout for tokens? atm it seems to me that:
1. tokens can be reused indefintely
2. in private/withheld, with just one token a client can get any fully signed tx in the virtual mempool

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -210,15 +211,19 @@ message GetVtxosResponse {
...
 message GetVtxoChainResponse {
   repeated IndexerChain chain = 1;
   IndexerPageResponse page = 2;
+  string auth_token = 3;
 }
 
 message GetVirtualTxsRequest {
   repeated string txids = 1;
   IndexerPageRequest page = 2;
+  string auth_token = 3;
```
louisinger: should we also add this auth_token to GetPendingTx ? 
it seems similar behavior but not in indexer ?

internal/core/application/indexer.go
altafan: ideally we should have test cases for these scenarios

louisinger (commented):
LGTM but it seems we'll reuse that process to authenticate several RPCs (now GetPendingTxs, GetVtxoChain etc..). I have the feeling this logic could be an interface layer middleware. wdyt ?

altafan (changes requested):
can you please open a dedicated pr for the changes related to the Signer (SignMessage rpc)? This way we can merge those changes already as they are just good

api-spec/protobuf/signer/v1/service.proto
altafan: i missed this. For now let's add the notations for http. Very soon we'll decide to keep the REST mapping also for the signer or if we can drop it entirely from all rpc in this proto

api-spec/protobuf/ark/v1/indexer.proto
altafan: GetPendingTx is not a method of the indexer, and no, no token for it, it's already guarded by the intent

api-spec/protobuf/signer/v1/service.proto
bitcoin-coder-bob: FWIW code rabbit said: https://github.com/arkade-os/arkd/pull/904#discussion_r2774904071

bitcoin-coder-bob: added in: [1d7480d](https://github.com/arkade-os/arkd/pull/904/commits/1d7480dfb72c439950300261e1d8a413e04358e7)

internal/core/application/indexer.go
bitcoin-coder-bob: added in: [1d7480d](https://github.com/arkade-os/arkd/pull/904/commits/1d7480dfb72c439950300261e1d8a413e04358e7)

internal/core/application/indexer.go
bitcoin-coder-bob: added in: [1d7480d](https://github.com/arkade-os/arkd/pull/904/commits/1d7480dfb72c439950300261e1d8a413e04358e7)

bitcoin-coder-bob:
> can you please open a dedicated pr for the changes related to the Signer (SignMessage rpc)? This way we can merge those changes already as they are just good

https://github.com/arkade-os/arkd/pull/954

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -3,6 +3,7 @@ syntax = "proto3";
 package ark.v1;
 
 import "meshapi/gateway/annotations.proto";
+import "ark/v1/types.proto";
```
altafan: please, do not import from types. Rather, add a brand new `IndexerIntent` message and duplicate the one we have in types.

internal/config/config.go
```diff
@@ -586,6 +612,17 @@ func (c *Config) Validate() error {
...
 	}
 
+	if !supportedIndexerTxExposures.supports(c.IndexerTxExposure) {
+		return fmt.Errorf(
+			"indexer txn exposure type not supported, please select one of: %s",
+			supportedIndexerTxExposures,
+		)
+	}
+
+	if c.IndexerAuthTokenExpiry <= 0 {
```
altafan: If the tx exposure is set to `public` we should skip this check as no intent would be needed to get the vtxo chain?

bitcoin-coder-bob: done in: [bf9e0a9](https://github.com/arkade-os/arkd/pull/904/commits/bf9e0a97b2fe783510d0ee3b74c8630a561d8369)

api-spec/protobuf/ark/v1/indexer.proto
bitcoin-coder-bob: done in: [bf9e0a9](https://github.com/arkade-os/arkd/pull/904/commits/bf9e0a97b2fe783510d0ee3b74c8630a561d8369)

internal/core/application/testdata/indexer_fixtures.json
```diff
@@ -0,0 +1,54 @@
...
+    "private_key_hex": "0000000000000000000000000000000000000000000000000000000000000001",
+    "public_key_hex": "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
+    "test_cases": [
+      {
+        "name": "valid_token_simple_outpoint",
+        "outpoint": {
+          "txid": "0000000000000000000000000000000000000000000000000000000000000001",
+          "vout": 0
+        },
+        "should_validate": true
```
altafan: let's replace this with "expected_token"?

internal/core/application/testdata/indexer_fixtures.json
```diff
@@ -0,0 +1,54 @@
...
+        "name": "valid_token_max_vout",
+        "outpoint": {
+          "txid": "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+          "vout": 4294967295
+        },
+        "should_validate": true
+      }
+    ]
+  },
+  "invalid_token_tests": [
```
altafan: let's merge these test cases so that we have a single object:
```
{
   "auth_token": {
      "private_key_hex": "0000000000000000000000000000000000000000000000000000000000000001",
      "public_key_hex": "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798",
      "valid": [{...}. ...], 
      "invalid": [{...}. ...], 
   }
}
```

bitcoin-coder-bob: done in: [dff1879](https://github.com/arkade-os/arkd/pull/904/commits/dff1879eab9f9e513e73b42f09e32854b9196948)

internal/core/application/testdata/indexer_fixtures.json
bitcoin-coder-bob: done in: [dff1879](https://github.com/arkade-os/arkd/pull/904/commits/dff1879eab9f9e513e73b42f09e32854b9196948)

internal/core/application/indexer_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+			token, err := indexer.createAuthTokenWithTimestamp(outpoint, fixedTimestamp)
+			require.NoError(t, err)
+			require.Equal(t, tc.ExpectedToken, token)
+
+			_, valid, err := indexer.validateAuthToken(token)
+			require.NoError(t, err)
+			require.True(t, valid, "token validation mismatch")
+		})
+	}
```
altafan: apply the same also to invalid
```suggestion
t.Run("valid", func(t * testing.T) {
	for _, tc := range f.AuthToken.Valid {
		t.Run(tc.Name, func(t *testing.T) {
			outpoint := Outpoint{
				Txid: tc.Outpoint.Txid,
				VOut: tc.Outpoint.Vout,
			}

			token, err := indexer.createAuthTokenWithTimestamp(outpoint, fixedTimestamp)
			require.NoError(t, err)
			require.Equal(t, tc.ExpectedToken, token)

			_, valid, err := indexer.validateAuthToken(token)
			require.NoError(t, err)
			require.True(t, valid, "token validation mismatch")
		})
	}
})
```

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -210,15 +209,19 @@ message GetVtxosResponse {
...
 message GetVtxoChainResponse {
   repeated IndexerChain chain = 1;
   IndexerPageResponse page = 2;
+  string auth_token = 3;
 }
 
 message GetVirtualTxsRequest {
   repeated string txids = 1;
   IndexerPageRequest page = 2;
+  string auth_token = 3;
```
altafan: we need also an `IndexerIntent intent` field: if a user wants to just fetch a tx w/ sigs, atm he's forced to call GetVtxoChain while he could just pass an intent that proves the ownership of the tx ins

internal/core/application/indexer.go
```diff
@@ -35,20 +64,61 @@ type IndexerService interface {
...
+		vtxoKey Outpoint,
+		intent Intent,
+		page *Page,
+	) (*VtxoChainResp, error)
+	GetVirtualTxs(
+		ctx context.Context,
+		authToken string,
+		txids []string,
+		page *Page,
+	) (*VirtualTxsResp, error)
```
altafan, line 74: for now, mainly for consistency, let's follow the pattern used also for other methods like `GetVtxos` / `GetVtxosByOutpoint`.

Therefore, let's have `GetVtxoChain(ctx, intent, token, page)` / `GetVtxoChainByOutpoint(ctx, outpoint, page)`, and `GetVirtualTxs(ctx, intent, token page)` / `GetVirtualTxsByIds(ctx, txids, page)`.

In both methods that require the intent, we first have to validate it (or the auth token) and extract the outpoint/txid from the proof tx. From there, things look exactly like the `GetVtxoChainByOutpoint` or `GetVirtualTxsByIds`, so I expect to see 2 internal methods `getVtxoChain(ctx, outpoint, page)` and `getVirtualTxs(ctx, txids, page)` to prevent duplication of code.

internal/core/application/indexer.go
```diff
@@ -250,10 +320,52 @@ func (i *indexerService) GetVtxosByOutpoint(
 }
 
 func (i *indexerService) GetVtxoChain(
-	ctx context.Context, vtxoKey Outpoint, page *Page,
+	ctx context.Context, vtxoKey Outpoint, intentForProof Intent, page *Page,
 ) (*VtxoChainResp, error) {
+	chain, err := i.buildVtxoChain(ctx, vtxoKey)
```
altafan: re https://github.com/arkade-os/arkd/pull/904/changes#r2974002111.

This can work only for indexer in public config. In other configurations, users pass an intent from which we have to extract the outpoint

internal/core/application/indexer.go
```diff
@@ -379,22 +491,285 @@ func (i *indexerService) GetVtxoChain(
...
+		if _, ok := chainTxids[txid]; !ok {
+			return fmt.Errorf("txid %s is not part of the authenticated vtxo chain", txid)
+		}
+	}
+
+	return nil
+}
+
+// similar flow in DeleteIntentsByProof inside internal/core/application/service.go
+func (i *indexerService) ValidateIntentWithProof(
```
altafan: no need to export this method, also what about this renaming?
```suggestion
func (i *indexerService) validateIntentProof(
```

internal/core/application/indexer.go
```diff
@@ -379,22 +491,285 @@ func (i *indexerService) GetVtxoChain(
...
+		}
+	}
+
+	signedProof, err := i.signTransactionTapscript(intentForProof.Proof)
+	if err != nil {
+		return errors.INTERNAL_ERROR.New("failed to sign proof: %w", err).
+			WithMetadata(map[string]any{
+				"proof": proof.UnsignedTx.TxID(),
+			})
+	}
```
altafan: Drop. With its key, the indexer signs only auth tokens. Intents should be signed by the signer instead, but we also don't want the indexer interact with the signer at all.

Probably we should just "partially" verify the intent, like if it's all good and only the signer sig is missing, we can consider the intent as valid.

For sure tho, making a random key sign the intent message won't make it valid, so let's remove these lines (and probably also the entire method)

api-spec/protobuf/ark/v1/indexer.proto
bitcoin-coder-bob: the user would still need an auth token though, which they can only get if they call `GetVtxoChain`, or are you suggesting they could pass this new `intent` param _in place of_ an auth token

internal/core/application/indexer.go
bitcoin-coder-bob: done in: [d5de1bf](https://github.com/arkade-os/arkd/pull/904/commits/d5de1bf8986438537c83f773b423995efc8f2f50)

internal/core/application/indexer.go
bitcoin-coder-bob: done in: [d5de1bf](https://github.com/arkade-os/arkd/pull/904/commits/d5de1bf8986438537c83f773b423995efc8f2f50)

internal/core/application/indexer_test.go
bitcoin-coder-bob: done in: [d5de1bf](https://github.com/arkade-os/arkd/pull/904/commits/d5de1bf8986438537c83f773b423995efc8f2f50)

internal/core/application/indexer.go
bitcoin-coder-bob: lgtm, will change

internal/core/application/indexer.go
bitcoin-coder-bob: ah ok I based this off of `DeleteIntentByProof` which also calls `SignTransactionTapscript` but I see that one actually calls the signer, while in my version it is not (and as you mentioned should not have indexer interact with signer). I will remove the `signTransactionTapscript` call and the following call to `Verify` and remove `signTransactionTapscript` function

internal/core/application/indexer.go
bitcoin-coder-bob: done in: [ea14596](https://github.com/arkade-os/arkd/pull/904/commits/ea14596289f215c0a0c3baa03e230dcefb2bdce6)

internal/core/application/indexer.go
bitcoin-coder-bob: done in: [f8870b7](https://github.com/arkade-os/arkd/pull/904/commits/f8870b76ea11650ce6df45737c738f68f408879e)

api-spec/protobuf/ark/v1/indexer.proto
bitcoin-coder-bob: we are going to go with the "in place of an auth token" for 1) better UX and 2) reduce calls to GetVtxo (an expensive call). Win win.

bitcoin-coder-bob: done here: [0c15516](https://github.com/arkade-os/arkd/pull/904/commits/0c15516bf8187fee8daaa4139cfdc82a9860acd4)

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -210,15 +209,24 @@ message GetVtxosResponse {
 message GetVtxoChainRequest {
   IndexerOutpoint outpoint = 1;
   IndexerPageRequest page = 2;
+  IndexerIntent intent = 3;
```
altafan: this should be:
```suggestion
  oneof auth {
     IndexerIntent intent = 3;
     token = 4;
  }
```

The reason is that if the response is paginated, we don't want the user to submit again the intent, rather he can just use the token.

api-spec/protobuf/ark/v1/indexer.proto
```diff
@@ -210,15 +209,24 @@ message GetVtxosResponse {
...
 }
 message GetVtxoChainResponse {
   repeated IndexerChain chain = 1;
   IndexerPageResponse page = 2;
+  string auth_token = 3;
```
altafan, line 225: worth adding a comment to specify this auth token can be used in others apis that require one (GetVirtualTxs)

internal/core/application/indexer.go
```diff
@@ -250,19 +324,218 @@ func (i *indexerService) GetVtxosByOutpoint(
...
+	txids := make([]string, 0, len(round.SweepTxs))
+	for txid := range round.SweepTxs {
+		txids = append(txids, txid)
+	}
+
+	return txids, nil
+}
+
+func (i *indexerService) getVtxoChain(
+	ctx context.Context, vtxoKey Outpoint, page *Page, token string,
```
bitcoin-coder-bob: we only ever call this function with `token` set to `""` should we remove it as a param and just set `AuthToken` to `""` @altafan

internal/core/application/indexer.go
```diff
@@ -374,48 +648,331 @@ func (i *indexerService) GetVtxoChain(
...
+	}
+
+	if !sig.Verify(msgHash, pubkey) {
+		return "", fmt.Errorf("signature verififcation failed")
+	}
+
+	return hex.EncodeToString(msg[:32]), nil
+}
+
+// hashOutpoints clones the given outpoints, sorts them lexycographically by txid and vout,
```
bitcoin-coder-bob: `lexicographically`

internal/core/application/token_cache.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	go c.sweep()
+	return c
+}
+
+// sweep removes expired entries on a fixed interval. Correctness does not
+// depend on this — reads already do a lazy expiry check — but it bounds
+// memory growth under high token volume.
+func (c *tokenCache) sweep() {
+	ticker := time.NewTicker(c.invalidationDuration / 2)
```
bitcoin-coder-bob, line 31: why divide by 2? @altafan

internal/core/application/indexer.go
altafan: Fixed in f7bf8d1

internal/core/application/indexer.go
altafan: Fixed in f7bf8d1

internal/core/application/token_cache.go
altafan: just to increase the frequncy

internal/core/application/indexer.go
```diff
@@ -250,19 +321,220 @@ func (i *indexerService) GetVtxosByOutpoint(
...
+			}
+
+			outpoints, ok := i.tokenCache.getOutpoints(hash)
+			if !ok {
+				return nil, fmt.Errorf("auth token not found")
+			}
+			if _, ok := outpoints[outpoint.String()]; !ok {
+				return nil, fmt.Errorf("auth token is not for outpoint %s", outpoint)
+			}
+		}
```
chris-ricketts, line 344:
```suggestion
	// Nothing to do as no signature stripping is performed on the VTXO chain
	case exposurePublic, exposureWithheld:
```

Unless signature stripping is performed, optionally validating the token in `exposureWithheld` is dead code.

internal/core/application/indexer.go
```diff
@@ -30,25 +59,67 @@ type IndexerService interface {
...
+	GetVtxoChain(
+		ctx context.Context, authToken string, vtxoKey Outpoint, page *Page,
+	) (*VtxoChainResp, error)
+	GetVtxoChainByIntent(ctx context.Context, intent Intent, page *Page) (*VtxoChainResp, error)
+	GetVirtualTxs(
+		ctx context.Context, authToken string, txids []string, page *Page,
+	) (*VirtualTxsResp, error)
+	GetVirtualTxsByIntent(ctx context.Context, intent Intent, page *Page) (*VirtualTxsResp, error)
 	GetBatchSweepTxs(ctx context.Context, batchOutpoint Outpoint) ([]string, error)
 	GetAsset(ctx context.Context, assetID string) ([]Asset, error)
```
chris-ricketts, line 77: Would it be worth adding a `RevokeToken` endpoint that deletes the cache entry keyed by the outpoint hash so the owner does not have to wait for the expiry if they know it leaked?

internal/core/application/indexer.go
```diff
@@ -250,19 +321,220 @@ func (i *indexerService) GetVtxosByOutpoint(
...
+			return nil, err
+		}
+	}
+
+	resp, allOutpoints, err := i.getVtxoChain(ctx, outpoint, page)
+	if err != nil {
+		return nil, err
+	}
+
+	token, err := i.createAuthToken(allOutpoints)
```
chris-ricketts, line 394: Should we check for a non-expired token cache entry before creating another one? An attacker with a valid proof could spam the endpoint and create millions of cache entries. This could DoS the indexer endpoints due to the mutex lock in the cache sweep task iterating through all the tokens. 

If you check the cache for a non-expired entry for the same outpoints it is a form of rate limiting on cache growth. 

Same for `GetVirtualTxsByIntent`

internal/core/application/indexer.go
altafan: we'll do this in a dedicated pr since it's an admin endpoint

internal/core/application/indexer.go
altafan: on it. I'm gonna change createAuthToken so that if token already exists, the function uses the existing expiration. This way, we make the endpoint always return the same token if it already exists.

internal/core/application/token_cache.go
```diff
@@ -0,0 +1,115 @@
...
+type tokenCache struct {
+	mu                   sync.RWMutex
+	outpointsByHash      map[string]map[Outpoint]time.Time
+	invalidationDuration time.Duration
+	stop                 chan struct{}
+}
+
+func newTokenCache(invalidationDuration time.Duration) *tokenCache {
+	c := &tokenCache{
+		outpointsByHash:      make(map[string]map[Outpoint]time.Time),
```
bitcoin-coder-bob, line 19:
```suggestion
        // stores outpoints by hash (hash of outpoints), with a shared expiration time for all outpoints of the same hash
		outpointsByHash:      make(map[string]map[Outpoint]time.Time),
```

internal/core/application/indexer.go
chris-ricketts: Resolving - this is so clients can practice handling token expiry during the transition from public to private.

internal/core/application/indexer.go
chris-ricketts: I was thinking it would be the intent issuer that would have the ability to revoke the token

internal/core/application/token_cache.go
altafan: Added in ae38f47

internal/core/application/indexer.go
altafan: Fixed in 35fc69b

internal/core/application/indexer.go
altafan: no revoking is something that only admin should be able to do

PR #908, opened 2026-02-10
Scale the DAG

### LINKED ISSUE #833: Scale the DAG
We want a schema and flow that:

- Scales to very deep DAGs (20k+ depth).
- Keeps **sweeping** as a cheap bulk operation.
- Lets us query `GetVtxoChain` and `ListVtxos (is_swept)` efficiently.

Core ideas:

1. Track a `depth` on every `vtxo` (longest-path depth).
2. Every 100 depths, create a **marker** in a `markers` table.
3. Each marker records all parent markers (multi‑parent DAG).
4. A `swept_markers` table isolates sweep state as **bulk inserts only**.
5. `vtxo` gets a `markers text[]` column (the markers that “cover” it).
6. Drop `vtxo.swept`; compute `is_swept` by joining against `swept_markers`.
7. `GetVtxoChain` uses markers to chunk recursion / traversal into healthy pieces.

#### 1.1 `vtxo` table

```sql
ALTER TABLE vtxo
    ADD COLUMN depth INT NOT NULL DEFAULT 0;

ALTER TABLE vtxo
    ADD COLUMN markers TEXT[]; -- e.g. ARRAY['txid:vout', 'txid2:vout2'];

ALTER TABLE vtxo
    DROP COLUMN swept;

CREATE INDEX CONCURRENTLY idx_vtxo_markers_gin
    ON vtxo USING GIN (markers);
```

#### 1.2 `markers` table

```sql
CREATE TABLE markers (
    -- A marker is a concrete vtxo outpoint: "txid:vout"
    marker_id      TEXT PRIMARY KEY,
    depth          INT NOT NULL,
    parent_markers TEXT[] NOT NULL  -- array of marker_ids (multi-parent DAG)
);

CREATE INDEX idx_markers_parent_gin
    ON markers USING GIN (parent_markers);
```

#### 1.3 `swept_markers` table

```sql
CREATE TABLE swept_markers (
    marker_id TEXT PRIMARY KEY,
    swept_at  BIGINT NOT NULL   -- unix timestamp
);
```

---

### 2. Ingestion Logic

**Goal:** For every new vtxo:

- Compute its `depth`.
- Assign its `markers text[]`.
- Optionally insert a new `markers` row if it sits on a 100‑depth boundary.

#### 2.1 Depth

For a new tx with inputs `parents []Vtxo`:

```go
func computeDepth(parents []Vtxo) int {
    maxDepth := 0
    for _, p := range parents {
        if p.Depth > maxDepth {
            maxDepth = p.Depth
        }
    }
    return maxDepth + 1
}
```

Store `depth` on the new vtxo.

#### 2.2 Markers and marker rows

- Interval: `MarkerInterval = 100`.
- `markerId(v) = v.Txid + ":" + strconv.Itoa(v.Vout)`.

```go
const MarkerInterval = 100

func assignMarkersAndMaybeCreateMarker(ctx context.Context, db *sql.DB, newVtxo *Vtxo, parents []Vtxo) error {
    // depth already computed
    depth := newVtxo.Depth

    // union parent markers
    parentMarkerSet := make(map[string]struct{})
    for _, p := range parents {
        for _, m := range p.Markers {
            parentMarkerSet[m] = struct{}{}
        }
    }
    parentMarkers := make([]string, 0, len(parentMarkerSet))
    for m := range parentMarkerSet {
        parentMarkers = append(parentMarkers, m)
    }

    if depth%MarkerInterval == 0 {
        // This vtxo becomes a marker
        id := newVtxo.Txid + ":" + strconv.Itoa(newVtxo.Vout)

        _, err := db.ExecContext(ctx, `
            INSERT INTO markers (marker_id, depth, parent_markers)
            VALUES ($1, $2, $3)
            ON CONFLICT (marker_id) DO NOTHING
        `, id, depth, pq.Array(parentMarkers))
        if err != nil {
            return err
        }

        newVtxo.Markers = []string{id}
        return nil
    }

    // Not a marker: inherit parent markers (multi‑parent)
    newVtxo.Markers = parentMarkers
    return nil
}
```

Insert `vtxo` and the `markers` row in the same transaction.

---

### 3. Sweeper Logic

**Goal:** Sweeping should only be **bulk inserts** into `swept_markers`. No per‑vtxo updates.

#### 3.1 Sweep from a root marker down

Given a starting `marker_id` (e.g. root or any “sweep point”), mark it and all descendant markers as swept:

```sql
WITH RECURSIVE target_markers AS (
    SELECT marker_id
    FROM markers
    WHERE marker_id = $1

    UNION ALL

    SELECT m.marker_id
    FROM markers m
    JOIN target_markers t
      ON t.marker_id = ANY(m.parent_markers)
)
INSERT INTO swept_markers (marker_id, swept_at)
SELECT marker_id, EXTRACT(EPOCH FROM NOW())::BIGINT
FROM target_markers
ON CONFLICT (marker_id) DO NOTHING;
```

For multiple starting markers:

```sql
WITH RECURSIVE target_markers AS (
    SELECT marker_id
    FROM markers
    WHERE marker_id = ANY($1::text[])

    UNION ALL

    SELECT m.marker_id
    FROM markers m
    JOIN target_markers t
      ON t.marker_id = ANY(m.parent_markers)
)
INSERT INTO swept_markers (marker_id, swept_at)
SELECT DISTINCT marker_id, EXTRACT(EPOCH FROM NOW())::BIGINT
FROM target_markers
ON CONFLICT (marker_id) DO NOTHING;
```

Sweeper service just calls this with the appropriate root marker(s).

---

### 4. `ListVtxos` / `fetchVtxo`: `is_swept` via view

**Goal:** `is_swept` must be cheap ($O(1)$) and derived from markers.

```sql
CREATE OR REPLACE VIEW vtxo_vw AS
SELECT
    v.*,
    EXISTS (
        SELECT 1
        FROM swept_markers sm
        WHERE sm.marker_id = ANY(v.markers)
    ) AS is_swept
FROM vtxo v;
```

`ListVtxos` can now select from `vtxo_vw` and expose `is_swept` without touching the sweeper logic.

Example:

```sql
SELECT txid, vout, depth, markers, is_swept, amount, expires_at
FROM vtxo_vw
WHERE pubkey = $1
ORDER BY created_at DESC
LIMIT $2;
```

---

### 5. `GetVtxoChain(outpoints)`: chunked DAG via markers

**Goal:** Given one or more outpoints, return a slim, paginated DAG of `child → parents` without recursing 20k+ steps in one go.

Two key pieces:

1. **Marker skeleton traversal** (very light, every 100 depth).
2. **Gap fill** per skeleton segment using `depth` ranges.

#### 5.1 Marker skeleton (upward)

Start from the marker(s) that cover the initial outpoint(s) and walk parents up, limited per page.

```sql
WITH RECURSIVE marker_chain AS (
    -- start marker for this page (derived from vtxo.markers of the starting outpoint)
    SELECT marker_id, parent_markers, depth, 1 AS level
    FROM markers
    WHERE marker_id = $1

    UNION ALL

    SELECT m.marker_id, m.parent_markers, m.depth, mc.level + 1
    FROM markers m
    JOIN marker_chain mc
      ON m.marker_id = ANY(mc.parent_markers)
    WHERE mc.level < $2  -- e.g. limit to 50 markers per page
)
SELECT marker_id, parent_markers, depth
FROM marker_chain;
```

Client encodes the last marker_id / level into `page_token` to request the next chunk.

#### 5.2 Gap fill between markers (0–100 depth window)

For each `[depth_lo, depth_hi]` window (e.g. `[1000, 1100]`) defined by two markers, fetch the actual tx graph:

```sql
SELECT
    t.txid,
    COALESCE(v.expires_at, 0) AS expires_at,
    t.type,
    array_remove(array_agg(DISTINCT p.txid), NULL) AS spends
FROM tx t
JOIN vtxo v
  ON v.txid = t.txid
LEFT JOIN vtxo p
  ON p.spent_by = t.txid
WHERE v.depth BETWEEN $1 AND $2   -- e.g. 1000 and 1100
GROUP BY t.txid, v.expires_at, t.type
ORDER BY v.depth DESC;            -- or ASC, depending on API semantics
```

Each row maps cleanly to `IndexerChain`:

```protobuf
message IndexerChain {
  string txid = 1;
  int64  expires_at = 2;
  IndexerChainedTxType type = 3;
  repeated string spends = 4;
}
```

The service:

- Uses markers to find depth windows per page.
- Runs the gap‑fill query for that window.
- Builds `IndexerChain` results.
- Returns a `page_token` encoding the next marker / window.

---

### 6. Why this works

- **Sweeper:** Only runs bulk inserts into a small `swept_markers` table.
- **VTXO rows:** Immutable w.r.t sweep; no 20k‑row UPDATEs.
- **`is_swept`:** Simple `EXISTS` over `markers[]` × `swept_markers`.
- **`GetVtxoChain`:** Works by:
  - traversing a 100× compressed marker graph (recursive, but light),
  - then fetching bounded depth ranges per page (no deep recursion on raw vtxos).

This keeps writes cheap, reads predictable, and makes the DAG depth effectively unbounded while still API‑friendly.

bitcoin-coder-bob:
> **Status:** #1092 (cursor-based pagination for `GetVtxoChain`) is merged and this branch is rebased on top of it. The duplicated pagination pieces (proto fields, cursor encode/decode, token-cache changes) have been dropped from this diff. What remains here is the swap of the cursor internals: #1092 encoded the cursor as an absolute offset over a full chain rebuild (O(N²) across all pages, accepted as interim); this PR replaces that with the frontier-resume walk so each page resumes the BFS at its frontier instead of rebuilding the whole chain (see the preload caveat in Notes for the remaining cross-page cost). The external contract (`page_token` / `next_page_token`, opaque tokens) is identical, so the swap is transparent to clients.

---

Closes #833

## Summary

Introduces a **marker DAG** — sparse checkpoints placed every 100 depths in the VTXO chain — that enables bulk DB operations instead of per-VTXO queries. Combined with bulk offchain tx fetching (merged from #1005), `GetVtxoChain` on a 10k-deep linear chain drops from thousands of sequential DB round-trips to ~200 bulk calls.

### What changed

1. **Marker system** — new `marker` and `swept_marker` tables, `depth` and `markers` JSONB columns on `vtxo`, with GIN indexes. Markers are created at every `MarkerInterval = 100` depths. Each VTXO carries the IDs of markers covering it.
2. **Bulk VTXO preload via marker DAG** — `preloadByMarkers()` walks the marker DAG upward from the frontier, bulk-loading all VTXO records into `vtxoCache` in ~N/100 calls instead of N individual fetches.
3. **Bulk offchain tx preload** (from #1005) — piggybacked onto the marker DAG walk, `GetOffchainTxsByTxids` bulk-fetches offchain txs per marker window into `offchainTxCache`, so the BFS loop never hits the DB for individual offchain tx lookups.
4. **In-loop bulk offchain tx fallback** (from #1005) — for VTXOs outside a marker window (race conditions, legacy VTXOs without markers), each BFS iteration collects missing txids and does one bulk fetch via `GetOffchainTxsByTxids` instead of one-by-one.
5. **Swept tracking replaces the per-VTXO swept column** — the `swept` boolean column is removed from `vtxo`; swept status is computed dynamically in the views from two append-only tables (see View changes below). Batch and checkpoint sweeps bulk-insert per-outpoint rows into `swept_vtxo`. Sweeping by marker is deliberately NOT the write path there: inherited markers can span independent subtrees when offchain txs consolidate inputs from different lineages, so a marker sweep would over-reach. `swept_marker` is written only with self-markers, by the schema migration (legacy swept VTXOs) and by dust-VTXO creation (dust outputs get their own outpoint marker and are immediately swept via `BulkSweepMarkers`). Marker-descendant sweeping was implemented to the issue's spec but never wired into the sweeper for the over-reach reason above, and has since been removed from the repo layer (see the cleanup note below). Finalized offchain txs that are already considered swept (batch swept or expired) persist their non-dust outputs into `swept_vtxo` at projection time, since the dropped column no longer carries that state. Backend nuance: the view-computed swept applies to postgres and sqlite; badger keeps storing `Swept` as an indexed field on the vtxo record and syncs it on sweeps.
6. **Pagination internals swapped to frontier-resume** — the `pageToken`/`nextPageToken` API landed on master via #1092 with an interim full-rebuild + offset implementation. This PR re-encodes the cursor as the walk frontier (base64 outpoints, HMAC-signed) so each page resumes the BFS where the previous one stopped instead of rebuilding the whole chain.
7. **`depth` field on VTXOs** — exposed via protobuf for clients to track chain position.
8. **New DB index** `idx_checkpoint_tx_offchain_txid` on `checkpoint_tx(offchain_txid)` to accelerate the bulk offchain tx query's join.

### Notes

- Existing VTXOs (pre-migration) are promoted onto the optimized path **automatically**. The schema migration first gives every VTXO a depth-0 self-marker; then a Go data migration (`BackfillVtxoMarkers`, gated at the schema migration version `20260701000000`) computes real BFS depths and assigns 100-depth-interval boundary markers to all existing chains. It runs once at startup, is idempotent (completion latch), and preserves swept status. On a production snapshot (112k VTXOs) it completes in ~30s. New VTXOs inherit markers from application code.
- Each db type ships exactly one migration pair: `20260701000000_add_vtxo_marker_dag` installs the whole marker-DAG schema plus the checkpoint-tx index, and the Go backfill gate targets the same version. The version is above the current production migration version, so golang-migrate applies it on deploy instead of silently skipping it as out-of-order.
- Preload ignores `pageSize` — peak memory scales with full chain length, and each page's preload re-climbs the remaining ancestor chain on a fresh per-request cache, so total DB work across all pages of a paginated read is superlinear. A process-level LRU cache keyed by marker ID (or a pageSize-bounded preload) would collapse that duplicate work but is out of scope here.
- `idx_checkpoint_tx_offchain_txid` adds a small write-path cost per checkpoint tx insert.
- The badger backend gets full marker support for parity, but its marker operations are built on store scans rather than indexes. Acceptable for the non-default backend at current scales; stubbing it (markers nil, walk falls back to the per-hop path, which the code already handles) is a reasonable alternative if maintainers prefer.

### Divergences from issue #833

This PR implements the issue's schema and goals but departs from its design in three places. Calling them out explicitly so review can judge each on its merits:

- **Traversal**: #833 specifies chunked traversal in SQL — a marker-skeleton walk plus depth-window gap-fill queries feeding the response directly. This PR keeps the BFS in Go and uses the marker DAG to bulk-preload per-request caches instead. The issue's gap-fill queries exist at the repo layer (`GetMarkersByDepthRange`, `GetVtxosByDepthRange`) but the traversal does not use them today.
- **Sweeping**: #833 wants sweeps to be bulk inserts into `swept_marker` only. Production sweeping is per-outpoint bulk inserts into `swept_vtxo` instead, because marker-granularity sweeping over-reaches when offchain txs consolidate inputs from independent lineages (a fund-safety issue found during development — see item 5 above). `swept_marker` only ever receives self-markers (legacy migration and dust).
- **Ingestion transactionality**: #833 wants the vtxo and its marker row inserted in the same transaction. Marker and vtxo stores are separate stores/transactions here; a partial-failure fallback (inherit parent markers) keeps VTXOs sweepable if the marker write fails.
- Minor schema detail: markers are stored as JSONB (postgres) / JSON text (sqlite) instead of the issue's `TEXT[]`.

### Background: the VTXO tree and `walkVtxoChain`

A VTXO chain is a tree of transactions. `walkVtxoChain` does a BFS — it processes all VTXOs at one depth, discovers their children via checkpoint tx inputs, then moves to the next depth.

Each VTXO is either **on-chain** (leaf of a batch tree) or **preconfirmed** (created by an offchain Ark tx). Preconfirmed VTXOs have a corresponding **OffchainTx** record containing checkpoint PSBTs whose inputs point to parent VTXOs — that's how the traversal discovers the next level.

```mermaid
graph TD
    subgraph "VTXO tree walkVtxoChain traverses"
        R["Round Commitment Tx (on-chain root)"]
        R --> A["VTXO A (depth 0)"]
        R --> B["VTXO B (depth 0)"]
        A --> C["VTXO C (preconfirmed, depth 1)"]
        A --> D["VTXO D (preconfirmed, depth 1)"]
        B --> E["VTXO E (preconfirmed, depth 1)"]
        C --> F["VTXO F (preconfirmed, depth 2)"]
        C --> G["VTXO G (preconfirmed, depth 2)"]
        D --> H["VTXO H (preconfirmed, depth 2)"]
        E --> I["VTXO I (preconfirmed, depth 2)"]
        E --> J["VTXO J (preconfirmed, depth 2)"]
    end
```

### Marker DAG: compressed traversal structure

Instead of visiting every VTXO individually, the marker DAG creates checkpoints every 100 depths. `preloadByMarkers` walks this small DAG to bulk-load all VTXOs and their offchain txs before the BFS loop starts.

```mermaid
graph TD
    subgraph "Marker DAG (small, ~N/100 nodes)"
        M0["Marker M0 (depths 0-99)"]
        M1["Marker M1 (depths 100-199)"]
        M2["Marker M2 (depths 200-299)"]
        M0 --> M1 --> M2
    end

    subgraph "VTXO tree (large, N nodes)"
        V0["~100 VTXOs (depths 0-99)"]
        V1["~100 VTXOs (depths 100-199)"]
        V2["~100 VTXOs (depths 200-299)"]
    end

    M0 -.->|covers| V0
    M1 -.->|covers| V1
    M2 -.->|covers| V2
```

### How preloadByMarkers works

The marker DAG walk bulk-fetches both VTXO records and offchain txs per marker window, so the BFS loop runs entirely from cache:

```mermaid
sequenceDiagram
    participant W as walkVtxoChain
    participant MDAG as Marker DAG walk
    participant DB as Database

    Note over W,MDAG: Before BFS loop starts
    W->>MDAG: preloadByMarkers(startVtxos)

    loop For each marker window (N/100 iterations)
        MDAG->>DB: GetVtxoChainByMarkers(window)
        DB-->>MDAG: ~100 VTXO records
        MDAG->>DB: GetOffchainTxsByTxids(preconfirmed txids)
        DB-->>MDAG: ~100 OffchainTx records
        MDAG->>DB: GetMarkersByIds(window)
        DB-->>MDAG: parent marker IDs
    end

    MDAG-->>W: vtxoCache + offchainTxCache fully populated

    Note over W,DB: BFS loop runs entirely from cache
    loop N iterations
        W->>W: vtxoCache lookup (hit)
        W->>W: offchainTxCache lookup (hit)
    end
```

### In-loop bulk fetch fallback (for VTXOs without markers)

For wide fanout trees or VTXOs without marker coverage, the BFS loop itself collects missing offchain txids per iteration and batch-fetches them:

```mermaid
graph TD
    subgraph "BFS iteration: 3 preconfirmed VTXOs at depth 1"
        C["VTXO C - txid: abc1"]
        D["VTXO D - txid: abc2"]
        E["VTXO E - txid: abc3"]
    end

    subgraph "Old: 3 individual queries"
        Q1["SELECT ... WHERE txid = 'abc1'"]
        Q2["SELECT ... WHERE txid = 'abc2'"]
        Q3["SELECT ... WHERE txid = 'abc3'"]
    end

    subgraph "New: 1 bulk query"
        QB["SELECT ... WHERE txid IN ('abc1','abc2','abc3')"]
    end

    C --> Q1
    D --> Q2
    E --> Q3
    C --> QB
    D --> QB
    E --> QB
```

### Combined effect (10k-entry linear chain)

```mermaid
graph TD
    subgraph "Before (no markers, no bulk fetch)"
        BA["VTXO fetches: ~10000 individual queries"]
        BB["OffchainTx fetches: ~10000 individual queries"]
        BC["Total: ~20000 round-trips"]
        BA --> BC
        BB --> BC
    end

    subgraph "After markers only (no offchain tx preload)"
        MA["VTXO fetches: ~100 bulk queries (markers)"]
        MB["OffchainTx fetches: ~10000 individual queries"]
        MC["Total: ~10100 round-trips"]
        MA --> MC
        MB --> MC
    end

    subgraph "After markers + offchain tx preload"
        FA["VTXO fetches: ~100 bulk queries (markers)"]
        FB["OffchainTx fetches: ~100 bulk queries (markers)"]
        FC["Total: ~200 round-trips"]
        FA --> FC
        FB --> FC
    end
```

## Benchmarks

**Bottom line: 13x faster on average, 21x on deep chains, p99 collapses from 40s to 2.3s — once the backfill migration has run on existing data (automatic on deploy).**

### Real-world: production DB snapshot (60 outpoints, max chain depth 3,266)

Probed against a snapshot of the production Postgres DB using the same locked set of 60 outpoints across all variants. Each variant ran against a live arkd process via gRPC.

Three variants were measured:

- **master** — baseline, no marker DAG
- **dag-1, view fixed** — marker DAG active, but migration backfill assigned depth=0 self-markers to all existing VTXOs so `preloadByMarkers` degenerates to per-VTXO fetches on historical data. Speedup here comes primarily from the bulk offchain tx fetch (#1005).
- **dag-1, backfilled** — same code, with existing VTXOs backfilled to proper BFS depths and 100-depth-interval markers, allowing `preloadByMarkers` to bulk-load large subtrees. The `BackfillVtxoMarkers` data migration brings all existing VTXOs to this state automatically on deploy; new VTXOs are here from day one.

| | master | dag-1 (view fixed) | dag-1 (backfilled) |
|---|---|---|---|
| p50 latency | 18.1s | 4.8s | **1.2s** |
| p90 latency | 33.8s | 7.7s | **2.0s** |
| p99 latency | 40.6s | 25.3s | **2.3s** |
| max latency | 42.6s | 53.9s | **2.3s** |
| mean latency | 16.4s | 5.4s | **1.1s** |
| 3,266-hop chain | 40.6s | 9.9s | **1.9s** |
| vs master (mean) | 1x | 3x faster | **13x faster** |
| vs master (median) | 1x | 3.6x faster | **14.8x faster** |

The p99/max tail collapses from 40-54s down to 2.3s once the marker DAG has real topology. The variance is essentially gone because `preloadByMarkers` fetches large subtrees in bulk, removing per-hop DB cost.

Note: the "view fixed" column also corrects a regression in the initial migration that caused the broken `swept` expression to do a sequential scan of the `swept_marker` table for every VTXO row in the outer query. Without that fix dag-1 was ~40% slower than master. See the View changes section below.

### Per-phase timing breakdown (10k-entry linear chain, 50µs simulated latency)

`TestVtxoChainTimingBreakdown` — in-process test with timing-decorated repos:

```
GetVtxoChain timing breakdown: linear chain n=10000, simulated repo latency=50µs
  wall clock (GetVtxoChain)        349ms
  Markers.GetMarkersByIds           92ms  (100 calls)
  Markers.GetVtxoChainByMarkers     57ms  (100 calls)
  OffchainTxs.GetOffchainTxsByTxids  53ms  (100 calls)
  Vtxos.GetVtxos                     1ms  (1 calls)
  sum of repo phases               203ms
  other (psbt parse + overhead)    145ms
```

All three repo phases (markers, VTXOs, offchain txs) make exactly **100 calls each** — one per marker window. The remaining 145ms is PSBT deserialization and in-process overhead.

### Call-counting test: 511-VTXO fanout tree

`TestBulkOffchainTxReducesDBCalls` — exercises the in-loop bulk fetch in isolation:

|  | Bulk calls | Individual calls | Total round-trips |
|---|---|---|---|
| **With bulk prefetch** | 9 | 0 | **9** |
| **Without (pre-optimization)** | 0 | 511 | **511** |

~57x reduction in DB round-trips. The 9 bulk calls correspond to 9 BFS depth levels.

### Latency benchmark: 511-VTXO fanout tree

`BenchmarkOffchainTxBulkVsSingle` — 50µs simulated DB latency:

| | Time/op |
|---|---|
| **Bulk prefetch** | 14.1ms |
| **No bulk (fallback)** | 390ms |
| **Speedup** | **~28x** |

### Test coverage

- **`TestVtxoChainTimingBreakdown`** — per-phase wall-clock breakdown for a 10k linear chain with timing-decorated repos. No prod instrumentation needed.
- **`TestBulkOffchainTxReducesDBCalls`** — asserts bulk fetch eliminates individual `GetOffchainTx` calls on a fanout tree.
- **`BenchmarkOffchainTxBulkVsSingle`** — wall-clock comparison with simulated latency.
- **`BenchmarkGetVtxoChain`** — linear chains (1k-5k), fanout trees, and diamond patterns with/without markers.
- **`TestVtxoChain -chain-length=5000`** — e2e test that builds a 5000-hop VTXO chain via self-sends and calls `GetVtxoChain` on the tip. Periodically calls `Settle` to refresh VTXO expiry when approaching the CSV locktime window.
- **Repo-level tests** — multi-txid bulk fetch, marker CRUD, sweep operations, across Postgres, SQLite, and Badger.
- **Domain tests** — marker boundary detection, inheritance, deep chains (20k+), multiple parents, deduplication.
- **Pagination e2e** — cursor-based pagination with early termination.
- **`TestOffchainTxSweptOutputsPersisted`** — end-to-end projection test, the only one that runs the real event-to-projection pipeline with a live tx decoder. Covers swept persistence for non-dust outputs of a swept/expired offchain tx.

> **Reviewer note — temporary verification tests.** Three extra tests were written while validating the migration prod-deployability against a production snapshot. The two `PROD_SCRATCH_DSN`-gated ones (`prod_migration_ordering_test.go`, `prod_backfill_verify_test.go`) have been removed from the diff; they remain reproducible from git history if needed. What ships is:
> - `vtxo_marker_backfill_demo_test.go` (`TestBackfillEnablesBulkPreload_Demo`) — self-contained SQLite demo showing the backfill promotes legacy self-marker VTXOs onto the `preloadByMarkers` fast path (before/after marker-window load counts). Runs in CI.

### New tables

- `marker(id TEXT PK, depth INTEGER, parent_markers JSONB)` — DAG checkpoints every 100 depths
- `swept_marker(marker_id TEXT PK FK→marker, swept_at BIGINT)` — append-only sweep records keyed by marker. In practice written only with self-markers (one marker = one VTXO): legacy swept VTXOs migrated by the schema migration, and dust VTXOs swept at creation. Coarse marker-subtree sweeping exists in the repo layer but is not wired into the sweeper
- `swept_vtxo(txid TEXT, vout INTEGER, swept_at BIGINT, PK(txid, vout))` — per-outpoint sweep records. This is the write path for both batch/round sweeps and checkpoint sweeps, because inherited markers can span independent subtrees and sweeping by marker would over-reach

### Modified tables

- `vtxo` — added `depth INTEGER NOT NULL DEFAULT 0`, `markers JSONB NOT NULL DEFAULT '[]'` (GIN indexed); removed `swept` column

### New indexes

- `idx_vtxo_markers` — GIN index on `vtxo.markers`
- `idx_marker_depth` — B-tree on `marker.depth`
- `idx_marker_parent_markers` — GIN on `marker.parent_markers`
- `idx_checkpoint_tx_offchain_txid` — B-tree on `checkpoint_tx.offchain_txid`

### View changes

- `vtxo_vw` rebuilt to compute `swept` dynamically from both sweep sources: `swept_marker` (self-marker records from the legacy migration and dust sweeps, probed by unnesting the per-row `markers` array) OR `swept_vtxo` (per-outpoint records written by batch and checkpoint sweeps):

```sql
(
    EXISTS (
        SELECT 1
        FROM jsonb_array_elements_text(v.markers) AS m(marker_id)
        JOIN swept_marker sm ON sm.marker_id = m.marker_id
    )
    OR EXISTS (
        SELECT 1 FROM swept_vtxo sv
        WHERE sv.txid = v.txid AND sv.vout = v.vout
    )
) AS swept
```

The earlier `@>` form (`v.markers @> jsonb_build_array(sm.marker_id)`) scanned `swept_marker` sequentially for every VTXO row — O(swept_marker) per outer row regardless of index. The unnest+join form probes `swept_marker_pkey` per marker ID, eliminating the scan.

### Migration backfill

The schema migration (`20260701000000_add_vtxo_marker_dag`) seeds a legacy shape — every VTXO gets a depth-0 self-marker:

- Creates one marker per existing VTXO (using outpoint as ID)
- Assigns that marker to each VTXO's `markers` column
- Migrates swept VTXOs to `swept_marker` table
- Drops `swept` column from `vtxo`

A follow-up Go data migration (`BackfillVtxoMarkers`, gated at the schema migration version `20260701000000`) then computes real BFS depths and boundary markers for the existing chains, promoting them onto the `preloadByMarkers` fast path. It is transactional (Postgres) / latch-guarded (SQLite, Badger), idempotent on re-run, and verifies swept status is preserved. Measured at ~30s on a 112k-VTXO production snapshot.

Additional notes:

- The gRPC probe harness used to produce the production benchmarks above (`cmd/probe-vtxo-chain`, locked outpoint sets, A/B latency runs, CSV output) has been removed from the diff to keep it focused. It lives in git history at `3b22a08f` for anyone reproducing the numbers.
- master's `indexer_pagination_test.go` (added by #1092 for the interim offset-cursor internals) is deleted. Its coverage moved to `TestGetVtxoChainPagination` in `indexer_test.go`, which tests the frontier cursor, early termination, and invalid-token errors.
- Sweeping leaf VTXOs no longer assumes the VTXO sits at vout 0. Leaf txs can carry an anchor output, so the sweeper/admin flows resolve the outpoint via `extractVtxoOutpoint` and `GetAllChildrenVtxos` now takes the full outpoint instead of a txid.
- The indexer no longer takes the app-service offchain-tx cache as a dependency. The marker-driven preload and bulk fetches replaced it.
- **Dead-code cleanup (done).** The unused marker repo methods `SweepMarker`, `SweepMarkerWithDescendants`, `SweepVtxosByMarker`, and `GetMarkersByDepth` have been removed across all three backends along with their sqlc queries, mocks, and tests, together with the probe CLI and the prod verification tests (~1,700 lines net). `GetMarkersByDepthRange` and `GetVtxosByDepthRange` are deliberately kept: they are #833's gap-fill building blocks and stay pending the decision on moving traversal into SQL (see Divergences above).

## Summary by CodeRabbit

* **New Features**
  * Token-based pagination for VTXO chain queries (pageToken / nextPageToken).
  * VTXO depth and marker metadata added to responses.

* **Improvements**
  * Marker-based caching and prefetching to speed chain traversal.
  * Swept state tracking moved to marker/swept records and per-outpoint sweep support.
  * Batch offchain transaction fetch to reduce lookup overhead.

* **Database**
  * Migrations to add depth, markers, and swept tracking.


bitcoin-coder-bob:
@coderabbitai review

internal/infrastructure/db/service.go
```diff
@@ -588,20 +741,71 @@ func getNewVtxosFromRound(round *domain.Round) []domain.Vtxo {
...
+
+	sweptAt := time.Now().Unix()
+	if err := s.markerStore.BulkSweepMarkers(ctx, markerIDs, sweptAt); err != nil {
+		log.WithError(err).Warn("failed to bulk sweep markers")
+		return 0
+	}
+
+	totalSwept := int64(len(vtxos))
+	log.Debugf("bulk swept %d markers affecting %d vtxos", len(markerIDs), totalSwept)
+	return totalSwept
```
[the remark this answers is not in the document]
bitcoin-coder-bob: the code returns 0 on error at line 801 before reaching the totalSwept assignment at line 804

internal/infrastructure/db/postgres/migration/20260210100000_add_depth_and_markers.up.sql
```diff
@@ -0,0 +1,94 @@
...
+SELECT v.*,
+    string_agg(vc.commitment_txid, ',') AS commitments,
+    EXISTS (
+        SELECT 1 FROM swept_marker sm
+        WHERE v.markers @> jsonb_build_array(sm.marker_id)
+    ) AS swept
+FROM vtxo v
+LEFT JOIN vtxo_commitment_txid vc
+ON v.txid = vc.vtxo_txid AND v.vout = vc.vtxo_vout
+GROUP BY v.txid, v.vout;
```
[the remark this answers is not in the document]
bitcoin-coder-bob: cosmetic. parseCommitments() already handles the NULL case gracefully

internal/infrastructure/db/postgres/marker_repo.go
```diff
@@ -0,0 +1,448 @@
...
+	markersJSON, err := json.Marshal(markerIDs)
+	if err != nil {
+		return fmt.Errorf("failed to marshal markers: %w", err)
+	}
+	return m.querier.UpdateVtxoMarkers(ctx, queries.UpdateVtxoMarkersParams{
+		Markers: markersJSON,
+		Txid:    outpoint.Txid,
+		Vout:    int32(outpoint.VOut),
+	})
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: wrong. json.RawMessage is defined as type RawMessage []byte so Go allows implicit assignment from []byte. The code compiles and works fine

internal/infrastructure/db/postgres/marker_repo.go
```diff
@@ -0,0 +1,448 @@
...
+
+	return m.querier.UpsertMarker(ctx, queries.UpsertMarkerParams{
+		ID:    marker.ID,
+		Depth: int32(marker.Depth),
+		ParentMarkers: pqtype.NullRawMessage{
+			RawMessage: parentMarkersJSON,
+			Valid:      true,
+		},
+	})
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: fixed in: [bd5e93c](https://github.com/arkade-os/arkd/pull/908/commits/bd5e93cc46aae59085355c5810afc6ce06c9028e)

internal/core/application/service_test.go
```diff
@@ -66,3 +69,599 @@ func parseTime(t *testing.T, value string) time.Time {
...
+		parent := domain.Vtxo{Depth: 20000, MarkerIDs: []string{"marker-20000"}}
+		newDepth := calculateMaxDepth([]domain.Vtxo{parent}) + 1
+		require.Equal(t, uint32(20001), newDepth)
+
+		// Depth 20100 should also be a boundary
+		require.True(t, domain.IsAtMarkerBoundary(20100))
+		require.True(t, domain.IsAtMarkerBoundary(20200))
+		require.False(t, domain.IsAtMarkerBoundary(20001))
+		require.False(t, domain.IsAtMarkerBoundary(20099))
+	})
```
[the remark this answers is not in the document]
bitcoin-coder-bob: fixed in: [bd5e93c](https://github.com/arkade-os/arkd/pull/908/commits/bd5e93cc46aae59085355c5810afc6ce06c9028e)

internal/infrastructure/db/postgres/marker_repo.go
```diff
@@ -0,0 +1,452 @@
...
+	// Insert the marker into swept_marker (sweep state is computed via view)
+	if err := m.querier.InsertSweptMarker(ctx, queries.InsertSweptMarkerParams{
+		MarkerID: markerID,
+		SweptAt:  time.Now().Unix(),
+	}); err != nil {
+		return 0, fmt.Errorf("failed to insert swept marker: %w", err)
+	}
+
+	return count, nil
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: meh i dont think this is valid. this is only used in tests btw

internal/infrastructure/db/service.go
```diff
@@ -602,6 +749,77 @@ func getNewVtxosFromRound(round *domain.Round) []domain.Vtxo {
...
+	if len(noMarkerVtxos) > 0 {
+		count, err := s.vtxoStore.SweepVtxos(ctx, noMarkerVtxos)
+		if err != nil {
+			log.WithError(err).Warn("failed to sweep vtxos without markers")
+		}
+		totalSwept += int64(count)
+	}
+
+	return totalSwept
+}
```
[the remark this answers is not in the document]
louisinger: @bitcoin-coder-bob I think rabbit is right ?

internal/infrastructure/db/service.go
```diff
@@ -527,6 +581,75 @@ func (s *service) updateProjectionsAfterOffchainTxEvents(events []domain.Event)
...
+				log.WithError(err).Warn("failed to create marker for chained vtxo")
+				// Continue without marker - non-fatal
+			} else {
+				log.Debugf("created marker %s at depth %d", newMarkerID, newDepth)
+				markerIDs = []string{newMarkerID}
+			}
+		} else if len(parentMarkerIDs) > 0 {
+			// Inherit ALL markers from parents at non-boundary depth
+			markerIDs = parentMarkerIDs
+		}
```
louisinger: instead of having this logic here, we should have a domain.NewMarker(txid, newDepth, parentMarkerIDs) so we can unit test it there (and let's unexport IsAtMarketBoundary)

internal/core/application/service_test.go
```diff
@@ -66,3 +69,599 @@ func parseTime(t *testing.T, value string) time.Time {
...
+			var newDepth uint32
+			if len(tc.spentVtxos) > 0 {
+				newDepth = maxDepth + 1
+			}
+			require.Equal(t, tc.expectedDepth, newDepth, tc.description)
+		})
+	}
+}
+
+func TestDepthAtMarkerBoundary(t *testing.T) {
```
louisinger: all marker-only unit test should be moved to marker domain

internal/infrastructure/db/service.go
```diff
@@ -527,6 +581,75 @@ func (s *service) updateProjectionsAfterOffchainTxEvents(events []domain.Event)
...
+			if err != nil {
+				log.WithError(err).Warn("failed to decode checkpoint tx for depth calculation")
+				continue
+			}
+			spentOutpoints = append(spentOutpoints, ins...)
+		}
+
+		// Get spent VTXOs to calculate new depth
+		var newDepth uint32
+		var parentMarkerIDs []string
```
louisinger: `newDepth` and `parentMarkerIDs` should be computed in application/service.go `SubmitOffchainTx` func. They should also be added to the `OffchainTxAccepted` domain event so that they are available here when updating the projection store.

The domain pattern is that every piece of data present in projection types (here domain.OffchainTx) should also appear in one of the associated events (here domain.OffchainTxEvent).

This also has the advantage of saving one GetVtxos call, since SubmitOffchainTx already fetches the spent vtxos from the outpoints.

internal/core/application/indexer.go
```diff
@@ -250,17 +257,51 @@ func (i *indexerService) GetVtxosByOutpoint(
...
 	chain := make([]ChainTx, 0)
-	nextVtxos := []domain.Outpoint{vtxoKey}
+	nextVtxos := frontier
 	visited := make(map[string]bool)
 
+	// Lazy cache for VTXOs loaded during this page.
+	vtxoCache := make(map[string]domain.Vtxo)
+	loadedMarkers := make(map[string]bool)
+
 	for len(nextVtxos) > 0 {
```
louisinger, line 614: Besides pagination improvements, we must get rid of this loop. It is not scalable and has already caused memory issues. I noticed you added the GetVtxoChainByMarkers method, shouldn't we leverage it?

internal/core/domain/marker.go
```diff
@@ -0,0 +1,31 @@
+package domain
+
+// MarkerInterval is the depth interval at which markers are created.
+// VTXOs at depth 0, 100, 200, etc. create new markers.
+const MarkerInterval = 100
```
louisinger, line 7: what's happening to prod if we change this value ? like we have a DB with MarkerInterval 100 and we migrate to 150, do we need a specific migration for this ?

louisinger (changes requested):

internal/core/application/indexer.go
bitcoin-coder-bob: I think youre right but this may be better suited as a follow-up PR given the scope of the change. 

  - The loop doesn't just fetch VTXOs — it also fetches offchain transactions, decodes checkpoint PSBTs, builds ChainTx with spends/types, and handles the batch tree branch case
  - Pagination would need to work completely differently with marker-based fetching
  - The response ordering (chain from leaf to root) would need to be reconstructed from a flat set of marker-fetched VTXOs


 This PR adds the marker infrastructure which is a prerequisite for this optimization which I would say we should do as a follow-up PR

internal/core/domain/marker.go
bitcoin-coder-bob: that would be problematic. `newMarker` would would create and skip at wrong intervals relative to the existing data, and parent marker linking would be inconsistent. I don't think we could have a migration to fix this, unless the migration is simply to rebuild the DAG if the constant is changed (pain). We need to treat this value as a "please do not change" constant. I could document that accordingly.

internal/infrastructure/db/service.go
bitcoin-coder-bob: worked in: [eabad6f](https://github.com/arkade-os/arkd/pull/908/commits/eabad6f5f0cce365608d2e293a82cb2d76b7458b)

internal/core/application/service_test.go
bitcoin-coder-bob: worked in: [eabad6f](https://github.com/arkade-os/arkd/pull/908/commits/eabad6f5f0cce365608d2e293a82cb2d76b7458b)

internal/infrastructure/db/service.go
bitcoin-coder-bob: worked in: [eabad6f](https://github.com/arkade-os/arkd/pull/908/commits/eabad6f5f0cce365608d2e293a82cb2d76b7458b)

ghost (commented):
## 🔍 Arkana Review — arkd#908 (Scale the DAG)

**Summary:** Introduces a marker-based DAG checkpoint system for VTXO chains. Markers are created every 100 depths, enabling bulk sweep operations (1 INSERT per 100 VTXOs vs 1 UPDATE per VTXO) and prefetch-based chain traversal via `ensureVtxosCached`. Adds token-based pagination for `GetVtxoChain`, a `depth` field on VTXOs, and corresponding migrations for both Postgres and SQLite.

This is a large PR (10.5K lines, ~4.2K non-test). I've focused on the data model, migration safety, security implications, and sweep correctness.

---

### ✅ Architecture — well designed

1. **Marker model** is clean: `MarkerInterval = 100`, markers created at depth boundaries (0, 100, 200...), parent marker IDs for DAG traversal. The `NewMarker()` function is deterministic and side-effect free.

2. **Swept marker as append-only table** — excellent choice. Instead of mutating a `swept` boolean on individual VTXOs, inserting into `swept_marker` marks all VTXOs sharing that marker as swept atomically. This is both faster (100x fewer writes) and more correct (no stale swept state).

3. **Dynamic swept computation via EXISTS subquery** in `vtxo_vw`:
```sql
EXISTS (
  SELECT 1 FROM swept_marker sm
  WHERE v.markers @> jsonb_build_array(sm.marker_id)
) AS swept
```
This is always correct by construction — no stale state possible. The GIN index on `markers` makes this efficient.

4. **Token-based pagination for `GetVtxoChain`** — `encodeChainCursor` / `decodeChainCursor` using base64-encoded JSON is correct. The backward compat handling (nil page + empty token = full chain) preserves existing client behavior.

5. **Marker window prefetch** in `ensureVtxosCached` — when a cache miss occurs, loading all VTXOs sharing the same marker pre-warms the cache for subsequent BFS iterations. The `loadedMarkers` dedup map prevents redundant marker loads.

### 🔒 Security

1. **Sweep correctness** — the bulk sweep path (`BulkSweepMarkers`) sweeps by marker ID. Since markers are tied to depth boundaries and VTXOs inherit marker IDs from their parents, sweeping a marker sweeps exactly the VTXOs in its 100-depth window. **No double-spend vector** — a VTXO can only be swept once because `swept_marker` has a PK on `marker_id`.

2. **Migration safety** — the migration adds columns with `DEFAULT 0` / `DEFAULT '[]'`, which is safe for existing data. The view recreation (`DROP VIEW IF EXISTS` + `CREATE VIEW`) is atomic within the migration transaction. ⚠️ **One concern:** the `DROP VIEW IF EXISTS vtxo_vw` will invalidate any in-flight queries using the old view. On a busy production server, this could cause brief errors. Consider deploying during low-traffic or ensuring the migration runs in a maintenance window.

3. **No exit path impact** — unilateral exit paths are not modified. Marker/depth tracking is metadata for operational efficiency, not consensus-critical.

4. **Existing VTXOs don't get markers** — explicitly noted in the PR. Pre-migration VTXOs have `depth=0` and `markers=[]`, so they fall back to per-VTXO sweep behavior. This is correct — no false swept status.

### 🟡 Observations

1. **`GetVtxosByMarker` error is swallowed in `ensureVtxosCached`** — line `if err != nil { continue }`. The test (`TestEnsureVtxosCached_GetVtxosByMarkerErrorSwallowed`) explicitly validates this behavior. This is acceptable since the VTXO itself is still cached via the direct DB lookup, but consider at least logging the error at debug level so operational issues with the marker index don't go unnoticed.

2. **`SweepMarkerWithDescendants`** — recursive CTE for bulk sweeping is powerful but could be expensive on deep marker chains. Ensure there's a depth limit or that this is only called during scheduled sweeps, not on hot paths.

3. **SQLite migration** is 138 lines — significantly more complex than Postgres due to SQLite's limited `ALTER TABLE` support. Consider adding a comment in the migration file noting which sections are SQLite workarounds.

4. **Proto field number 15 for `depth`** — leaving gaps (14 is `assets`, 15 is `depth`) is fine for proto. Just noting it's consistent across both `Vtxo` and `IndexerVtxo` messages.

5. **No down migration for swept_marker** — `20260210100000_add_depth_and_markers.down.sql` exists but I couldn't verify its contents in this diff. Ensure it drops `swept_marker`, `marker`, and removes the `depth`/`markers` columns.

### Cross-repo impact
- **SDKs** need to expose the new `depth` field on VTXOs — this is additive and non-breaking (new field, default 0).
- **Pagination** via `pageToken` / `nextPageToken` on `GetVtxoChain` is additive — existing clients that don't set these fields get the old behavior.
- The e2e test gap (noted in PR: "require changes to the SDK to support new proto field") should be tracked as a follow-up issue.

### Test coverage
Test files account for ~62% of the diff. Key scenarios covered: cursor round-trip, cache hits/misses, marker dedup, DB error propagation, pagination (first page, resume with token, short chain, page size limits), and sweeper checkpoint creation with markers. Comprehensive for a change of this scope.

**Verdict:** Strong architectural improvement with clear performance wins. The marker model is well-designed and the migration is safe. The two items worth attention are: (1) logging swallowed errors in `ensureVtxosCached`, and (2) ensuring the view recreation in migration happens during low-traffic. LGTM for the design — recommend a second human review given the scope (10K+ lines touching DB schema, sweep logic, and indexer).

internal/infrastructure/db/service.go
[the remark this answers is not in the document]
bitcoin-coder-bob: i think this is an outdated comment, the code has changed since @louisinger

internal/core/application/indexer.go
louisinger: +1 for another PR, easier to review too. But we need to do it before merging this one. 
We have performance issue using that loop. for a 5k tx chain, it takes 15s on master to respond. With this branch it's 12s thanks to cursor improvements, which is not significant. 
we need to know if the marker really improve response time to client (I was expecting less than 3s for 5k txs chain). Just to confirm it's the "right" change for this vtxo chain getter

bitcoin-coder-bob: PR: https://github.com/arkade-os/arkd/pull/973

bitcoin-coder-bob: the loop wasn't removed but I believe the PR captures the spirit of the ask which is removing the bottleneck (all the DB round trip calls). It changes the DB access pattern from O(chain_length) individual `GetVtxos` calls to O(chain_length / MarkerInterval) bulk fetches. The VTXOs are prefetched into vtxoCache and the loop reads from the cache instead of hitting the DB each iteration.

ghost (commented):
## 🔍 Arkana PR Review — arkd#908 (incremental)

**New commit since last review:** `69879d04` — bulk-fetch offchain txs in `walkVtxoChain` (#1005). Plus `40ff9f82` (merge master, no logic).

This commit eliminates the remaining O(N) bottleneck I flagged in previous reviews: per-hop `GetOffchainTx` calls during chain traversal. Now offchain txs are bulk-loaded alongside VTXOs during the marker preload phase, and lazily bulk-fetched per-iteration for cache misses. Good.

### Findings

**1. 🟡 Badger `GetOffchainTxsByTxids` — fragile error matching**
`internal/infrastructure/db/badger/ark_repo.go` — the new method uses `strings.Contains(err.Error(), "not found")` to skip missing txids. This is brittle: if the error message in `getOffchainTx` ever changes (or a wrapped error incidentally contains "not found"), behavior silently changes. Use `errors.Is` with a sentinel error, or at minimum `errors.As` with a typed error. The underlying `getOffchainTx` already checks `badgerhold.ErrNotFound` — expose that check properly rather than string-matching on the formatted message.

**2. ✅ Postgres/SQLite bulk queries — safe**
- Postgres: `txid = ANY($1::varchar[])` with `pq.Array(txids)` — parameterized, no injection risk.
- SQLite: `sqliteMaxBulkTxids = 500` correctly stays under `SQLITE_MAX_VARIABLE_NUMBER`. Batching loop is clean.
- Both filter `COALESCE(fail_reason, '') = ''` consistently with the single-fetch query. Good.

**3. ✅ New index migration**
`20260409140000_checkpoint_tx_offchain_txid_index` — adds `idx_checkpoint_tx_offchain_txid` on `checkpoint_tx(offchain_txid)`. This supports the JOIN inside `offchain_tx_vw` when bulk-fetching. Correct optimization — the view was already doing this join, just without the index to back the multi-row case.

**4. ✅ Preload error semantics**
`preloadByMarkers` hard-fails if `GetOffchainTxsByTxids` errors, while marker window errors log-and-continue. This asymmetry is actually correct: marker cache misses degrade to per-VTXO lookups (functional), but offchain tx failures would produce incomplete chain data (incorrect). Right call.

**5. ✅ Cache key consistency**
Offchain tx cache uses `tx.ArkTxid` as key, lookups use `vtxo.Txid`. These are the same field semantically (the offchain tx's ark txid is the VTXO's parent txid). Verified in domain model.

**6. ✅ Fallback path**
When bulk fetch returns empty or misses a txid, the walk loop falls back to individual `GetOffchainTx` and populates the cache (`offchainTxCache[vtxo.Txid] = offchainTx`). Correct — handles both pre-optimization repos and race conditions where a tx is inserted after the preload.

**7. ✅ Test coverage**
- `TestBulkOffchainTxReducesDBCalls` — fanout tree, proves bulk calls = O(depth) vs individual calls = O(N). Correct assertion.
- `TestVtxoChainTimingBreakdown` — 10k-chain wall-clock breakdown with simulated latency. Useful diagnostic.
- `BenchmarkOffchainTxBulkVsSingle` — direct comparison with simulated 50µs latency. Good.
- `preconfirmed chain bulk-loads offchain txs` + `falls back to single fetch on cache miss` — unit tests for both the happy and fallback paths.
- DB integration test in `service_test.go` — multi-txid fetch with cross-contamination guard. Thorough.

**8. Nit:** `docker-compose.regtest.yml` bumps nbxplorer `2.5.30` → `2.5.30-1`. Unrelated — should probably be a separate commit.

**9. Security:** No impact. This is purely a read-path optimization. No changes to exit paths, forfeit validation, transaction signing, or sweep correctness.

### Summary

Only finding #1 (badger string-matching) is worth addressing — it's a latent bug vector. Everything else is clean. The optimization is well-structured with proper fallback paths and strong test coverage.

bitcoin-coder-bob:

bitcoin-coder-bob:

bitcoin-coder-bob:
> ## Arkana Code Review — PR #908: Scale the DAG (deep re-review)
> Full re-review of the current state. Previous critical findings (unsigned cursors, missing `swept` column in down migration) have been **fixed** — nice work. This review covers what remains.
> 
> ⚠️ **This PR modifies protocol-critical swept state handling (forfeit/exit path). Requires human sign-off from @Kukks, @altafan, or @louisinger before merge.**
> 
> ### 🔴 HIGH — Unbounded memory on `GetVtxoChainByIntent`
> **`indexer.go:460-473`**: `GetVtxoChainByIntent` calls `walkVtxoChain` with `math.MaxInt32` as pageSize, loading the **entire** chain into memory before applying offset-based `paginate()`. Combined with `preloadByMarkers` eagerly loading the full marker DAG, a single request on a 100k-VTXO chain loads all 100k VTXOs + offchain txs into memory. This is a DoS vector.
> 
> Same issue at `indexer.go:381` for backward-compat unpaginated `GetVtxoChain` requests.
> 
> **Fix:** Add a hard upper bound (e.g., 50,000 items) and return an error if exceeded. Or refactor `GetVtxoChainByIntent` to use cursor-based pagination internally.
> 
> ### 🔴 HIGH — Marker over-reach in partial batch sweeps
> **`service.go:861-868`**: `sweepVtxosWithMarkers()` collects ALL markers from the VTXOs being swept and calls `BulkSweepMarkers` on all of them. For **full round** sweeps this is correct. For **partial subtree** sweeps (`sweeper.go:710-725`), if VTXOs from independent subtrees share a marker window (which happens when offchain txs consolidate inputs from multiple lineages), sweeping that marker incorrectly marks **all** VTXOs under it as swept — including VTXOs in un-swept subtrees.
> 
> This is the exact same over-reach problem that `SweepVtxoOutpoints` was created to solve for checkpoint sweeps, but it persists in the partial batch sweep path. This is a potential fund-loss scenario: VTXOs incorrectly marked swept become unspendable.
> 
> **Fix:** Either use `SweepVtxoOutpoints` for partial batch sweeps too, or prove that partial batch sweeps can never encounter shared markers across independent subtrees (I don't believe that invariant holds once offchain txs consolidate inputs from different lineages).
> 
> ### 🟡 MEDIUM — SQLite `LIKE` pattern for marker matching (false-positive risk)
> **`sqlite/sqlc/query.sql:535-537`**:
> 
> ```sql
> WHERE markers LIKE '%"' || @marker_id || '"%'
> ```
> 
> This substring match can produce false positives if one marker ID is a prefix of another (e.g., `marker-1` matching `marker-10`). Postgres correctly uses JSONB `?|` operator; Badger uses exact element matching. The 64-char hex txids in marker IDs make this practically impossible today, but it's a latent correctness bug. If marker ID format ever changes, this becomes a real money bug (wrong VTXOs returned → wrong VTXOs swept).
> 
> **Track as tech debt.** Consider a `json_each` subquery or application-level dedup.
> 
> ### 🟡 MEDIUM — Backfill migration risk on large production DBs
> **`20260210100000_add_depth_and_markers.up.sql:53-71`**: The `INSERT INTO marker ... SELECT` + `UPDATE vtxo SET markers = ...` runs as a single transaction over every VTXO. On production DBs with millions of rows, this holds a table lock for potentially minutes and could OOM. No batching, no progress reporting.
> 
> **Recommendation:** Document expected duration for operators. Consider a batched approach for production rollout.
> 
> ### 🟡 MEDIUM — Badger `SweepVtxosByMarker` returns inflated count
> **`badger/marker_repo.go:486-501`**: Counts ALL VTXOs with the marker (including already-swept), while Postgres/SQLite count only unswept VTXOs before marking. Inconsistent across backends. If this count drives any logic beyond logging, it's a bug.
> 
> ### 🟡 MEDIUM — Cross-repo SDK impact: pagination not wired
> All downstream SDKs need updates to expose cursor-based pagination:
> 
> * **go-sdk**: Generated protobuf lacks `page_token`/`next_page_token` fields. Needs protobuf regen.
> * **rust-sdk**: Same stale protobuf. `VtxoChainResponse` struct needs updating after regen.
> * **ts-sdk**: REST client at `src/providers/indexer.ts:590-611` doesn't pass `pageToken`. Interface needs updating.
> * **In-repo client-lib** (`pkg/client-lib/indexer/grpc/client.go:366-372`): `GetVtxoChain` doesn't expose `PageToken`. Used by the redemption flow.
> 
> Not breaking (old pagination still works), but the new feature is dead code for all consumers until these land. Worth tracking follow-up PRs.
> 
> ### 🟡 MEDIUM — Down migration view shape drift
> **`20260210100000_add_depth_and_markers.down.sql:29-44`**: The restored `vtxo_vw` and `intent_with_inputs_vw` drop `asset_id`, `asset_amount`, and `intent_txid` columns compared to the current schema. A rollback could break the application layer. The `20260416` down migration is better-guarded (refuses rollback if data exists).
> 
> ### 🟢 LOW — `isActive` / `sweep` map iteration fragility
> **`token_cache.go:44-49,88-103`**: The `sweep()` function uses `break` after first map entry, relying on the invariant that all entries share the same expiry. Correct today, fragile if the invariant breaks. Consider a comment or assertion.
> 
> ### 🟢 LOW — SQLite B-tree index on JSON column is useless
> **`sqlite/migration/20260210000000:111`**: `idx_vtxo_markers` is a B-tree index on the `markers` text column. The swept check uses `json_each(v.markers)` which iterates the JSON array — the B-tree index doesn't help. Adds write overhead without query benefit.
> 
> ### 🟢 LOW — Badger timestamp granularity mismatch
> **`badger/marker_repo.go:488`**: Uses `time.Now().Unix()` (seconds) while Postgres/SQLite use `time.Now().UnixMilli()` (milliseconds). Inconsistent if timestamps are ever compared across backends.
> 
> ### ✅ What's been fixed since last review
> * **Cursor HMAC signing** — correctly implemented with constant-time comparison (`indexer.go:822-870`).
> * **Down migration `swept` column** — `20260210` down migration now properly restores the column and backfills from `swept_marker`.
> * **`swept_vtxo` safety guard** — `20260416` down migration refuses rollback if data exists. Excellent pattern.
> * **Checkpoint sweep path** — correctly uses `SweepVtxoOutpoints` (per-outpoint) to avoid marker over-reach.
> 
> ### ✅ What looks good
> * Marker DAG domain model is well-designed with correct cycle protection (SQL `UNION` not `UNION ALL`, Badger visited set).
> * Marker ID generation (`txid:marker:depth`) is collision-free.
> * Cursor HMAC key derivation from server privkey is clean.
> * `validateChainAuth` pagination grace period is a good UX tradeoff.
> * Dual swept-state architecture (marker-based for batch, per-outpoint for checkpoint) is sound.
> * Test coverage is excellent — timing breakdowns, call-counting, benchmarks, pagination e2e, domain tests, repo-level tests across all 3 backends.
> * Proto changes are additive and backward-compatible.
> 
> **Summary:** Two blockers remain: (1) unbounded memory on full-chain requests and (2) marker over-reach in partial batch sweeps. The rest are medium/low items that can be tracked. The overall design is strong and the benchmarks justify the complexity.


I implemented some of your comments. Heres pushback on others:

  Backward-compat unbounded path (line 380)
                                                                                                                                                                                                           
  ▎ The math.MaxInt32 on the backward-compat path is intentional: callers that send no page and no page_token are old clients that expect the full chain back. Capping this silently breaks them. The right
  ▎  fix is a protocol version negotiation so old clients can be identified and rejected cleanly, which is out of scope here. Added the bound to GetVtxoChainByIntent where we control the full call       
  ▎ surface.                                                          
                                                                                                                                                                                                           
  ---                                                                 
  SQLite LIKE false-positive
                                                                                                                                                                                                           
  ▎ Marker IDs are {64-char-hex-txid}:marker:{depth} or {64-char-hex-txid}:{vout}. A prefix collision requires a 64-character hex prefix match, which is computationally impossible. Acknowledged as a 
  ▎ latent correctness concern if the format ever changes, but not worth a structural query change now. Happy to add a json_each alternative behind a flag when/if marker ID format is revisited.          
                                                                      
  ---                                                                                                                                                                                                      
  Cross-repo SDK impact                                               
                                                                                                                                                                                                           
  ▎ Agreed these need to land, but they can't be atomic with server-side changes anyway since the old pagination wire format still works. Tracked as follow-up PRs. Listing them here would just hold this 
  ▎ PR hostage to unrelated repo release cycles.                                                                                                                                                           
                                                                      
  ---                                                                                                                                                                                                      
  Down migration view shape drift                                     
                                                                                                                                                                                                           
  ▎ The asset_id/asset_amount/intent_txid columns in the current view come from migrations that postdate 20260210. Rolling back 20260210 must be paired with a code rollback to the app version that 
  ▎ predates those later migrations, at which point those columns aren't referenced by the app either. A down migration restores the schema to the state before its up migration ran, not to the current   
  ▎ head, which is standard. The 20260416 migration's refusal-to-roll-back pattern is the right guard for irreversible data changes; schema shape drift on a coordinated rollback is expected and not 
  ▎ guarded by that pattern.                                                                                                                                                                               
                                                                      
  ---
  isActive / sweep fragility
                                                                                                                                                                                                           
  ▎ The break in sweep() already has the comment // all entries share the same expiresAt. isActive was intentionally written to scan all entries without relying on that invariant, which is why the 
  ▎ comment there explains the tradeoff. An assertion would panic in production on a logic bug rather than log and continue, which seems like the wrong failure mode for a background cache janitor.

PR #912, opened 2026-02-17
Harden stream listener fanout under churn

bitcoin-coder-bob:
Fixes a DoS vector on both TX and batch event stream subscriptions where an attacker can crash the server by churning (rapidly opening and closing) gRPC streams.                                        

  The crash is caused by two concurrency bugs in the stream listener fanout:

  1. Unsynchronized map iteration. The fanout loops in listenToEvents and listenToTxEvents iterate the listener map without holding a read lock. A concurrent subscribe/unsubscribe mutates the map,
  triggering a "concurrent map iteration and map write" fatal error.
  2. Send on closed channel. When a client disconnects, the gRPC handler calls removeListener (deleting the listener) then close(listener.ch). The fanout goroutine can race against this: it grabs a
  reference to the listener, the handler closes the channel, and the goroutine sends on the closed channel — panicking the server.
3.  An attacker can open lots of streams, not read from them, the listener channels fill up, the send event go routines are blocked and start piling up.
 Both 1 and 2 are trivially triggerable by any unauthenticated client that opens and closes streams in a tight loop.

 Fixes:
  - Snapshot the listener map before fanout (getListenersCopy). The fanout iterates a stable copy rather than the live map, preventing the concurrent map read/write fatal.
  - Stop closing listener.ch from the handler. The channel close is removed from the disconnect path, eliminating the send-on-closed-channel race.
  - Add a done channel to each listener (closed via sync.Once in removeListener). The fanout goroutine does a select between l.done and l.ch <-, abandoning delivery immediately when the listener is
  removed.

  Resource exhaustion

  The fanout select includes a default case in both listenToEvents and listenToTxEvents, so events are dropped rather than blocking when a listener's channel is full. This prevents goroutine pile-up from
   clients that open streams without reading from them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of slow subscribers with timeout protection
  * Reduced contention during event and transaction stream broadcasts
  * Better cleanup for disconnected or canceled subscribers

* **Tests**
  * Expanded test coverage for listener lifecycle and timeout handling
  * Added end-to-end tests for rapid subscription churn scenarios

### DIFF
--- a/internal/test/e2e/e2e_test.go
+++ b/internal/test/e2e/e2e_test.go
@@ -4638,17 +4638,17 @@ func TestTxListenerChurn(t *testing.T) {
 }
 
 // TestEventListenerChurn is the event-stream counterpart of TestTxListenerChurn.
-// Instead of offchain transactions, it uses settlement rounds to generate
+// Instead of offchain transactions, clients join batches to generate
 // events. The structure mirrors the tx test:
 //
 //  1. A sentinel event stream stays open for the full duration and counts
-//     round-lifecycle events. A broken fanout will sever this stream.
+//     batch-lifecycle events. A broken fanout will sever this stream.
 //  2. N churn workers rapidly open/close event streams to stress the
 //     subscribe/unsubscribe path.
-//  3. A round producer drives Settle+NotifyIncomingFunds across multiple
+//  3. A batch producer drives Settle+NotifyIncomingFunds across multiple
 //     participants to generate a steady flow of events.
 //
-// The test passes when at least one round completes, the sentinel observes
+// The test passes when at least one batch completes, the sentinel observes
 // events, and no sentinel errors are recorded.
 func TestEventListenerChurn(t *testing.T) {
 	const (


internal/interface/grpc/handlers/arkservice.go
```diff
@@ -611,18 +609,22 @@ func (h *handler) listenToEvents() {
...
 						if l.includesAny(ev.topics) {
-							l.ch <- ev.event
-							count++
+							select {
+							case <-l.done:
+								return
+							case l.ch <- ev.event:
+							default:
+							}
 						}
```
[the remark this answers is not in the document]
chris-ricketts: I agree it might be better to explicitly close the stream rather than keep it open with stale events. 

Closing the original blocking goroutine after a timeout would kill any others that pile up during the interval.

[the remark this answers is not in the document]
bitcoin-coder-bob: agreed @chris-ricketts. Set a 5 second window before closing. Pushed in commit: [23894d6](https://github.com/arkade-os/arkd/pull/912/commits/23894d6ddb51c747beda5894e30c0eec427e4656)

internal/test/e2e/e2e_test.go
```diff
@@ -4364,3 +4366,600 @@ func TestAsset(t *testing.T) {
...
+			if err != nil {
+				if stressCtx.Err() != nil {
+					return
+				}
+				if isRetryableChurnError(err) {
+					retryableSubscribeErrors.Add(1)
+					time.Sleep(churnWorkerBackoff(workerID))
+				} else {
+					return
+				}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: addressed in: [0165dc1](https://github.com/arkade-os/arkd/pull/912/commits/0165dc1749b679393f104dafd14d74463c09137a)

internal/test/e2e/e2e_test.go
```diff
@@ -4354,3 +4356,622 @@ func TestAsset(t *testing.T) {
...
+		case runErr := <-errCh:
+			require.NoError(t, runErr)
+		default:
+			return
+		}
+	}
+}
+
+// TestEventListenerChurn is the event-stream counterpart of TestTxListenerChurn.
+// Instead of offchain transactions, it uses settlement rounds to generate
```
altafan:
```suggestion
// Instead of offchain transactions, clients join batches to generate events
```

internal/test/e2e/e2e_test.go
```diff
@@ -4354,3 +4356,622 @@ func TestAsset(t *testing.T) {
...
+// events. The structure mirrors the tx test:
+//
+//  1. A sentinel event stream stays open for the full duration and counts
+//     round-lifecycle events. A broken fanout will sever this stream.
```
altafan:
```suggestion
//     batch-lifecycle events. A broken fanout will sever this stream.
```

internal/test/e2e/e2e_test.go
```diff
@@ -4354,3 +4356,622 @@ func TestAsset(t *testing.T) {
...
+//  2. N churn workers rapidly open/close event streams to stress the
+//     subscribe/unsubscribe path.
+//  3. A round producer drives Settle+NotifyIncomingFunds across multiple
```
altafan:
```suggestion
//  3. A batch producer drives Settle+NotifyIncomingFunds across multiple
```

internal/test/e2e/e2e_test.go
```diff
@@ -4354,3 +4356,622 @@ func TestAsset(t *testing.T) {
...
+//     participants to generate a steady flow of events.
+//
+// The test passes when at least one round completes, the sentinel observes
```
altafan:
```suggestion
// The test passes when at least one batch completes, the sentinel observes
```

bitcoin-coder-bob: fixed in: [c682467](https://github.com/arkade-os/arkd/pull/912/commits/c68246782d04985bc8b35424a9fab94feb846e41)

internal/test/e2e/e2e_test.go
bitcoin-coder-bob: fixed in: [c682467](https://github.com/arkade-os/arkd/pull/912/commits/c68246782d04985bc8b35424a9fab94feb846e41)

internal/test/e2e/e2e_test.go
bitcoin-coder-bob: fixed in: [c682467](https://github.com/arkade-os/arkd/pull/912/commits/c68246782d04985bc8b35424a9fab94feb846e41)

internal/test/e2e/e2e_test.go
bitcoin-coder-bob: fixed in: [c682467](https://github.com/arkade-os/arkd/pull/912/commits/c68246782d04985bc8b35424a9fab94feb846e41)

Issue #846, opened 2025-12-01
Failed to register intent with output over `vtxoMinAmount`


luckysori:
In a regtest environment against arkd v0.8.10, I get this server info:

```json
{
  "version": "",
  "signerPubkey": "0256f810de93e500e745b7dabfcb2b798b216a70a99de7edee79bf1791379bf62d",
  "forfeitPubkey": "0346b8d971bddc965e90f242605f13b86dc3795f9c0dbf04d97c5982628835efac",
  "forfeitAddress": "bcrt1q3sclrm7jqnqc4k9ms85r860qyzzgc908scsff7",
  "checkpointTapscript": "03a80040b2752046b8d971bddc965e90f242605f13b86dc3795f9c0dbf04d97c5982628835efacac",
  "network": "regtest",
  "sessionDuration": "10",
  "unilateralExitDelay": "86016",
  "boardingExitDelay": "1024",
  "utxoMinAmount": "330",
  "utxoMaxAmount": "-1",
  "vtxoMinAmount": "1",
  "vtxoMaxAmount": "-1",
  "dust": "330",
  "fees": {
    "intentFee": {
      "offchainInput": "",
      "offchainOutput": "",
      "onchainInput": "0",
      "onchainOutput": "0"
    },
    "txFeeRate": "0"
  },
  "scheduledSession": null,
  "deprecatedSigners": [],
  "serviceStatus": {},
  "digest": "35d84abe0865119b937063a1d6933212d2331139b4c3f2a1aa34ce429dbaa143"
}
```

But trying to register an intent with an output above `vtxoMinAmount` returns this error:

```
AMOUNT_TOO_LOW (15): output 1 amount is lower than min vtxo amount: 330
```

Even the error message implies that the min VTXO amount is not what the info object indicates.

bitcoin-coder-bob:
PR: https://github.com/arkade-os/arkd/pull/923

PR #923, opened 2026-02-20
vtxoMinOffchainTxAmount clamped to dust when below it

bitcoin-coder-bob:
closes https://github.com/arkade-os/arkd/issues/846

I think we were just reporting back the wrong field?

  Before:
```                                                                                                                                                                                                  
VtxoMinAmount:        s.vtxoMinOffchainTxAmount                                                                                                                                                                                  
```
  After:
``` 
VtxoMinAmount:        s.vtxoMinSettlementAmount
```

A client reads `vtxoMinAmount` from `GetInfo`, constructs a `RegisterIntent` with an amount above that value, and gets rejected. The fix of reporting `vtxoMinSettlementAmount` directly solves this because that's the value `RegisterIntent` checks against.

Move setting `vtxoMinSettlement`, `vtxoMinOffchain`, and `utxoMin` into util function, `resolveMinAmounts` in `internal/core/application/utils.go` and added test for it in `internal/core/application/service_test.go` in a test called `TestResolveMinAmount`.

## Summary by CodeRabbit

* **Refactor**
  * Unified min-amount handling so validations, error messages, and public info use a single consolidated VtxoMinAmount derived from a dust-based resolution.

* **Tests**
  * Added tests covering min-amount resolution across dust thresholds and edge cases (negative/unset values, zeros, exact dust, values above dust).

### DIFF
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -64,6 +64,7 @@ type service struct {
 	boardingExitDelay         arklib.RelativeLocktime
 	roundMinParticipantsCount int64
 	roundMaxParticipantsCount int64
+	dustAmount                uint64
 	utxoMaxAmount             int64
 	utxoMinAmount             int64
 	vtxoMaxAmount             int64
@@ -160,14 +161,6 @@ func NewService(
 		return nil, fmt.Errorf("failed to generate ephemeral key: %s", err)
 	}
 
-	dustAmount, err := wallet.GetDustAmount(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get dust amount: %s", err)
-	}
-	vtxoMinAmount, utxoMinAmount = resolveMinAmounts(
-		vtxoMinAmount, utxoMinAmount, int64(dustAmount),
-	)
-
 	roundReportSvc := reportSvc
 	if roundReportSvc == nil {
 		roundReportSvc = roundReportUnimplemented{}
@@ -232,6 +225,7 @@ func (s *service) Start() error {
 		return fmt.Errorf("failed to get dust amount: %s", err)
 	}
 
+	s.dustAmount = dustAmount
 	s.vtxoMinAmount, s.utxoMinAmount = resolveMinAmounts(
 		s.vtxoMinAmount, s.utxoMinAmount, int64(dustAmount),
 	)
@@ -982,7 +976,7 @@ func (s *service) SubmitOffchainTx(
 				})
 			}
 		}
-		if out.Value < s.vtxoMinAmount && !script.IsSubDustScript(out.PkScript) {
+		if out.Value < s.vtxoMinAmount {
 			return nil, errors.AMOUNT_TOO_LOW.New(
 				"output #%d amount is lower than min vtxo amount: %d",
 				outIndex, s.vtxoMinAmount,
@@ -1886,14 +1880,14 @@ func (s *service) RegisterIntent(
 					})
 				}
 			}
-			if amount < uint64(s.vtxoMinAmount) {
+			if amount < s.dustAmount {
 				return "", errors.AMOUNT_TOO_LOW.New(
 					"output %d amount is lower than min vtxo amount: %d",
-					outputIndex, s.vtxoMinAmount,
+					outputIndex, s.dustAmount,
 				).WithMetadata(errors.AmountTooLowMetadata{
 					OutputIndex: outputIndex,
 					Amount:      int(amount),
-					MinAmount:   int(s.vtxoMinAmount),
+					MinAmount:   int(s.dustAmount),
 				})
 			}
 


internal/core/application/service.go
```diff
@@ -963,14 +954,14 @@ func (s *service) SubmitOffchainTx(
...
+		if out.Value < s.vtxoMinAmount {
 			return nil, errors.AMOUNT_TOO_LOW.New(
 				"output #%d amount is lower than min vtxo amount: %d",
-				outIndex, s.vtxoMinOffchainTxAmount,
+				outIndex, s.vtxoMinAmount,
 			).WithMetadata(errors.AmountTooLowMetadata{
 				OutputIndex: outIndex,
-				Amount:      int(s.vtxoMinOffchainTxAmount),
-				MinAmount:   int(s.vtxoMinOffchainTxAmount),
+				Amount:      int(s.vtxoMinAmount),
```
altafan: wrong amount?

```suggestion
				Amount:      int(out.Value),
```

bitcoin-coder-bob: good catch

internal/core/application/service.go
```diff
@@ -963,14 +954,14 @@ func (s *service) SubmitOffchainTx(
...
 				"output #%d amount is lower than min vtxo amount: %d",
-				outIndex, s.vtxoMinOffchainTxAmount,
+				outIndex, s.vtxoMinAmount,
 			).WithMetadata(errors.AmountTooLowMetadata{
 				OutputIndex: outIndex,
-				Amount:      int(s.vtxoMinOffchainTxAmount),
-				MinAmount:   int(s.vtxoMinOffchainTxAmount),
+				Amount:      int(out.Value),
+				MinAmount:   int(s.vtxoMinAmount),
 			})
```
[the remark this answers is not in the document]
bitcoin-coder-bob: wouldnt we want: 
```
if out.Value < s.vtxoMinAmount && !script.IsSubDustScript(out.PkScript) {
```

dust is always 330, so i am unsure how `if out.Value >= int64(dust) && out.Value < s.vtxoMinAmount {` would evaluate to true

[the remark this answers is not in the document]
bitcoin-coder-bob: done in: [2ff7250](https://github.com/arkade-os/arkd/pull/923/commits/2ff72508e351a82d6db58ecdd33312ae41429d23)

internal/core/application/service.go
```diff
@@ -162,6 +160,14 @@ func NewService(
 		return nil, fmt.Errorf("failed to generate ephemeral key: %s", err)
 	}
 
+	dustAmount, err := wallet.GetDustAmount(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get dust amount: %s", err)
+	}
+	vtxoMinAmount, utxoMinAmount = resolveMinAmounts(
+		vtxoMinAmount, utxoMinAmount, int64(dustAmount),
+	)
```
altafan: this must be dropped after merging #924. The changes made to Start() below are just enough

internal/core/application/service.go
```diff
@@ -985,14 +982,14 @@ func (s *service) SubmitOffchainTx(
 				})
 			}
 		}
-		if out.Value < s.vtxoMinOffchainTxAmount {
+		if out.Value < s.vtxoMinAmount && !script.IsSubDustScript(out.PkScript) {
```
```suggestion
		if out.Value < s.vtxoMinAmount {
```

internal/core/application/service.go
```diff
@@ -1889,14 +1886,14 @@ func (s *service) RegisterIntent(
 					})
 				}
 			}
-			if amount < uint64(s.vtxoMinSettlementAmount) {
+			if amount < uint64(s.vtxoMinAmount) {
```
altafan: In Start(), we should take care of storing the dust amount we get at line https://github.com/arkade-os/arkd/blob/master/internal/core/application/service.go#L227 and use it here

```suggestion
			if amount < uint64(s.dustAmount) {
```

internal/core/application/service.go
bitcoin-coder-bob: @altafan https://github.com/arkade-os/arkd/pull/923#pullrequestreview-3876157462  do you disagree or agree with the AIs assessment?

Without this additional check sub-dust OP_RETURN outputs were rejected (at line: https://github.com/arkade-os/arkd/blob/eb83ea9fd70fb20e3002ee2007fdea3a72c62fad/internal/core/application/service.go#L979) before    
  ever reaching the OP_RETURN exemption (at line: https://github.com/arkade-os/arkd/blob/eb83ea9fd70fb20e3002ee2007fdea3a72c62fad/internal/core/application/service.go#L992). 

So should the flow be:
  1. skip the min-amount check for OP_RETURN scripts  (i.e. add back in the  `&& !script.IsSubDustScript(out.PkScript)`)                                                                                                                                          
  2. enforce that sub-dust outputs must be OP_RETURN scripts

internal/core/application/service.go
bitcoin-coder-bob: we want `RegisterIntent` to enforce the same minimum that `GetInfo` advertises:  `VtxoMinAmount: s.vtxoMinAmount` https://github.com/arkade-os/arkd/blob/eb83ea9fd70fb20e3002ee2007fdea3a72c62fad/internal/core/application/service.go#L2162 which `SubmitOffchainTx` enforces https://github.com/arkade-os/arkd/blob/eb83ea9fd70fb20e3002ee2007fdea3a72c62fad/internal/core/application/service.go#L979. Previously a client checking `GetInfo` would see one minimum but `RegisterIntent` would enforce a different one.

altafan: yeah and that different one was set by default to dust amount if not specified in config. We decided to not have this value configurable basically and always allow to renew all vtxos that have amount at least dust (330 sats)

bitcoin-coder-bob: Bit confused, seeking clarification. 

Since `vtxoMinAmount` will always equal dust (330), the check at line 979 (`out.Value < s.vtxoMinAmount`) rejects sub-dust outputs before they ever reach the OP_RETURN exemption at line 990. Should we:    
   
  1. Remove the `vtxoMinAmount` check at line 979 entirely (the dust check at line 990 already handles it with the OP_RETURN exemption), or                                                                  
  2. Can sub-dust OP_RETURN outputs never occur in `SubmitOffchainTx`, so this is fine as-is? 
  3. no action needed?

altafan: `vtxoMinAmount` refers to the min allowed output amount in offchain tx. It's set to dust amount if only if it's config value is `-1`. So for example:
* if env var is unset -> min vtxo amount = -1 -> min vtxo amount = 330 -> sub-dust output are not allowed in offchain txs
* env var set to `5` -> min vtxo amount = 5 -> sub-dust outputs are allowed in offchain txs with min possible value of `5` sats

PR #924, opened 2026-02-23
grpc: gate Ark/Indexer RPCs with readiness interceptors

sekulicd:
## Summary

This PR removes the gRPC server/service swap behavior during wallet readiness transitions and replaces it with middleware-based readiness gating.

Ark and Indexer RPCs are now always registered at startup, but access is blocked until the app backend is started and the wallet is ready (initialized, unlocked, synced).

## Why

Previously, `arkd` could restart/swap gRPC internals when the wallet became ready. That made reconnect behavior surprising for SDK clients because clients had to understand `arkd` internals and sometimes replace connections manually.

This PR moves the responsibility to the server side in a way that is transparent to clients:
- keep service registration stable
- return `FailedPrecondition` while wallet/app is not ready
- let clients rely on normal gRPC reconnect/backoff behavior

### gRPC readiness gating (server-side)
- Added unary + stream readiness interceptors for protected methods:
  - `/ark.v1.ArkService/*`
  - `/ark.v1.IndexerService/*`
- Readiness checks require:
  - app backend started (`appStarted`)
  - wallet status is initialized + unlocked + synced
- Returns:
  - `FailedPrecondition` when app/wallet is not ready
  - `Unavailable` when wallet status cannot be queried

### gRPC service lifecycle simplification
- Removed `withAppSvc` / `withoutAppSvc` branching in gRPC service startup/shutdown
- `onReady()` now activates app services (`startAppServices()`) instead of stopping/restarting the gRPC servers
- `stop()` uses a snapshot-under-lock pattern for app service shutdown and updates readiness state

### Application service startup refactor
- Moved wallet-dependent initialization from `application.NewService()` to `application.(*service).Start()`:
  - dust amount derived min values
  - forfeit pubkey
  - checkpoint tapscript
  - forfeit address
- Added `registerEventHandlers()` helper and moved event handler registration to `Start()`
- `Stop()` is now nil-safe if `Start()` fails before `sweeperCancel` is initialized
- `application.Service.Start()` now returns `error` instead of `errors.Error`

### Misc
- Added `.gitignore` entries for local worktree directories:
  - `.worktrees/`
  - `.worktree/`

## Behavior change (intended)

After `arkd` restart:
- gRPC transport can reconnect normally
- Ark/Indexer RPCs remain registered
- while wallet is locked/syncing, Ark/Indexer RPCs return `FailedPrecondition`
- once wallet is ready, the same client connection can continue working (no custom conn swap required)

## Tests

Added readiness interceptor tests covering:
- non-protected methods bypass readiness checks
- app-not-started -> `FailedPrecondition`
- wallet status error -> `Unavailable`
- locked/syncing wallet -> `FailedPrecondition`
- ready wallet -> success

Also verified targeted package tests locally:
- `go test ./internal/interface/grpc/...`
- `go test ./internal/core/application`

@altafan @louisinger please review

## Summary by CodeRabbit

* **New Features**
  * Service readiness checks now block protected APIs until the app and wallet are ready.
  * Client libraries: improved stream reconnection, backoff behavior, and clearer error propagation on stream/parsing failures.

* **Refactor**
  * App service startup made lazy and decoupled from server boot; lifecycle hardened to prevent duplicate starts and ensure safe shutdown ordering.

* **Tests**
  * Added unit tests for readiness gating (protected and unprotected paths).

* **Chores**
  * Added ignore patterns for local worktree directories.

### DIFF
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -2066,7 +2066,7 @@ func (s *service) SubmitForfeitTxs(ctx context.Context, forfeitTxs []string) err
 			WithMetadata(errors.InvalidForfeitTxsMetadata{ForfeitTxs: forfeitTxs})
 	}
 
-	go s.checkForfeitsAndBoardingSigsSent(round.CommitmentTxid)
+	go s.checkForfeitsAndBoardingSigsSent(context.WithoutCancel(ctx), round.CommitmentTxid)
 
 	return nil
 }
@@ -2105,7 +2105,7 @@ func (s *service) SignCommitmentTx(ctx context.Context, signedCommitmentTx strin
 			WithMetadata(map[string]any{"signed_commitment_tx": signedCommitmentTx})
 	}
 
-	go s.checkForfeitsAndBoardingSigsSent(round.CommitmentTxid)
+	go s.checkForfeitsAndBoardingSigsSent(context.WithoutCancel(ctx), round.CommitmentTxid)
 
 	return nil
 }
@@ -3704,8 +3704,7 @@ func (s *service) scheduleSweepBatchOutput(round *domain.Round) {
 	}
 }
 
-func (s *service) checkForfeitsAndBoardingSigsSent(commitmentTxid string) {
-	ctx := context.Background()
+func (s *service) checkForfeitsAndBoardingSigsSent(ctx context.Context, commitmentTxid string) {
 	// NOTE: This assumes users submit all their signatures in one shot, and whatever
 	// we get from the cache are all required sigs to finalize the boarding inputs
 	// once we also sign them
--- a/internal/core/application/admin.go
+++ b/internal/core/application/admin.go
@@ -572,7 +572,9 @@ func (a *adminService) Sweep(
 	log.Infof("sweep transaction %s broadcasted", txid)
 
 	if len(batchInputs) > 0 {
-		go a.saveBatchSweptEvents(batchInputs, batchRounds, batchVtxoTrees, txid, txhex)
+		go a.saveBatchSweptEvents(
+			context.WithoutCancel(ctx), batchInputs, batchRounds, batchVtxoTrees, txid, txhex,
+		)
 	}
 
 	return
@@ -635,13 +637,12 @@ func (a *adminService) getScheduledSweep(
 }
 
 func (a *adminService) saveBatchSweptEvents(
+	ctx context.Context,
 	inputsByCommitmentTxid map[string][]ports.TxInput,
 	batchesByCommitmentTxid map[string]*domain.Round,
 	treesByCommitmentTxid map[string]*tree.TxTree,
 	txid, txhex string,
 ) {
-	ctx := context.Background()
-
 	for commitmentTxid, batchInputsList := range inputsByCommitmentTxid {
 		round := batchesByCommitmentTxid[commitmentTxid]
 
--- a/internal/core/application/fraud.go
+++ b/internal/core/application/fraud.go
@@ -42,9 +42,7 @@ func (s *service) reactToFraud(ctx context.Context, vtxo domain.Vtxo, mutx *sync
 			return fmt.Errorf("failed to broadcast checkpoint tx: %s", err)
 		}
 
-		go func() {
-			ctx := context.Background()
-
+		go func(ctx context.Context) {
 			blockTimestamp, err := waitForConfirmation(
 				ctx,
 				ptx.UnsignedTx.TxID(),
@@ -66,7 +64,7 @@ func (s *service) reactToFraud(ctx context.Context, vtxo domain.Vtxo, mutx *sync
 			); err != nil {
 				log.Errorf("failed to schedule checkpoint sweep: %s", err)
 			}
-		}()
+		}(context.WithoutCancel(ctx))
 		return nil
 	}
 
--- a/internal/telemetry/otel.go
+++ b/internal/telemetry/otel.go
@@ -116,7 +116,7 @@ func InitOtelSDK(
 
 	otel.SetTextMapPropagator(propagation.TraceContext{})
 
-	go collectGoRuntimeMetrics(context.Background())
+	go collectGoRuntimeMetrics(ctx)
 
 	var rrExporter *RoundReportLogExporter
 	if rrsvc != nil {


internal/core/application/service.go
```diff
@@ -65,9 +65,11 @@ type service struct {
 	utxoMaxAmount             int64
 	utxoMinAmount             int64
 	vtxoMaxAmount             int64
+	vtxoMinAmount             int64
```
altafan, line 70: please revert any change related to this, we already have #923 fixing this

internal/core/application/service.go
```diff
@@ -225,34 +196,98 @@ func NewService(
...
 	}
+
+	// if err := svc.restoreWatchingVtxos(); err != nil {
+	// 	return nil, fmt.Errorf("failed to restore watching vtxos: %s", err)
+	// }
+	return svc, nil
+}
+
+func (s *service) Start() error {
+	ctx := context.Background() //TODO manage context
```
altafan: let's not postpone this since we're "refactoring". Let's change the sig of the function to `Start(ctx context.Context) error` and pass this context to all the go routines spawned here so we can finally kill them properly at shutdown

internal/core/application/service.go
```diff
@@ -225,34 +196,98 @@ func NewService(
...
+	s.registerEventHandlers()
+
+	go s.listenToScannerNotifications()
+
+	log.Debug("starting sweeper service...")
+	ctx, cancel := context.WithCancel(ctx)
+	s.sweeperCancel = cancel
+	go func() {
+		if err := s.sweeper.start(ctx); err != nil {
+			log.WithError(err).Warn("failed to start sweeper")
```
altafan, line 282:
```suggestion
			log.WithError(err).Warn("failed to start sweeper")
			return
```

internal/core/application/service.go
```diff
@@ -225,34 +196,98 @@ func NewService(
...
 		roundReportSvc:                roundReportSvc,
 		alerts:                        alerts,
 		settlementMinExpiryGap:        time.Duration(settlementMinExpiryGap) * time.Second,
 		vtxoNoCsvValidationCutoffTime: vtxoNoCsvValidationCutoffTime,
 		feeManager:                    feeManager,
 	}
+
+	// if err := svc.restoreWatchingVtxos(); err != nil {
+	// 	return nil, fmt.Errorf("failed to restore watching vtxos: %s", err)
+	// }
```
altafan: let's move this to Start too, we MUST NOT forget about it!!

internal/core/application/service.go
altafan: EDIT: let's do it in a dedicated pr

internal/interface/grpc/interceptors/readiness.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (r *ReadinessService) Check(ctx context.Context, fullMethod string) error {
+	if r == nil || !isProtectedServiceMethod(fullMethod) {
+		return nil
+	}
+	// We need both checks:
+	// 1) appStarted gates the Ark backend lifecycle (startup/shutdown races, start failures),
+	// 2) wallet.Status() gates the current runtime condition (locked/syncing after backend started).
+	if !r.appStarted.Load() {
+		return protectedServiceUnavailableErr(fullMethod)
```
altafan: let's return unavailable err code here so client knows that it can/should retry

internal/interface/grpc/interceptors/readiness.go
```diff
@@ -0,0 +1,105 @@
...
+		return status.Error(codes.Unavailable, "wallet status unavailable")
+	}
+
+	walletStatus, err := r.wallet.Status(ctx)
+	if err != nil {
+		return status.Errorf(codes.Unavailable, "wallet status unavailable: %v", err)
+	}
+	if !walletStatus.IsInitialized() || !walletStatus.IsUnlocked() || !walletStatus.IsSynced() {
+		return protectedServiceUnavailableErr(fullMethod)
+	}
```
altafan, line 63: if the app svc started and the wallet is down, arkd should be down as well, so this is bad and in this case we may don't want clients to retry connecting that frequently. Let's return FailPrecondition err code here

internal/core/application/service.go
altafan: solved in amster, just sync and solve conflicts

internal/interface/grpc/interceptors/readiness.go
sekulicd: Done

internal/interface/grpc/interceptors/readiness.go
sekulicd: Done

internal/core/application/service.go
sekulicd: Done

internal/core/application/service.go
sekulicd: Done

internal/interface/grpc/interceptors/readiness_test.go
```diff
@@ -0,0 +1,200 @@
+package interceptors
```
altafan, line 1: please use `require` instead of `t.Fatal`

internal/interface/grpc/interceptors/readiness_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+func TestReadinessServiceCheck(t *testing.T) {
+	t.Run("ignores non protected methods", func(t *testing.T) {
+		r := NewReadinessService(nil)
+		if err := r.Check(context.Background(), "/ark.v1.WalletService/Lock"); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+	})
+
+	t.Run("app not started returns failed precondition", func(t *testing.T) {
```
altafan: Fix the name of these tests after we switched the err codes

```suggestion
	t.Run("app not started returns unavailable", func(t *testing.T) {
```

internal/interface/grpc/interceptors/readiness_test.go
sekulicd: Done

internal/interface/grpc/interceptors/readiness_test.go
sekulicd: Done

internal/core/application/service.go
```diff
@@ -227,34 +197,101 @@ func NewService(
...
+		}
+		log.Info("sweeper service started")
+	}()
 
-	repoManager.Events().RegisterEventsHandler(
+	log.Debug("starting app service...")
+	s.wg.Add(1)
+	go s.start()
+	return nil
+}
```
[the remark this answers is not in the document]
sekulicd: Done

internal/interface/grpc/service.go
```diff
@@ -48,6 +49,9 @@ type service struct {
 	adminServer       *http.Server
 	grpcServer        *grpc.Server
 	adminGrpcSrvr     *grpc.Server
+	readinessSvc      *interceptors.ReadinessService
+	appSvcStartMu     sync.Mutex
+	appSvcStarted     bool
```
altafan, line 54: let's use atomic.Bool also here?

internal/interface/grpc/interceptors/readiness.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/arkade-os/arkd/internal/core/ports"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	arkServiceMethodPrefix     = "/ark.v1.ArkService/"
+	indexerServiceMethodPrefix = "/ark.v1.IndexerService/"
```
louisinger, line 16: I think we can get this from proto stubs (we did it in permissions.go)

Issue #920, opened 2026-02-20
Raw sql errors from API


louisinger:
Some API endpoints may return raw sql errors instead of pkg/errors structured error messages.
for instance the endpoint `/v1/indexer/commitmentTx/{unknown_txid}` may return "{"code":13, "message":"sql: no rows in result set", "details":[]}"

"no rows in result set" is a kind of error the user should never see.

we must ensure nonce of these kind of errros are returned by API endpoints.

bitcoin-coder-bob:
working off this PR: https://github.com/arkade-os/arkd/pull/930

PR #930, opened 2026-02-25
fix sql.ErrNoRows leak in GetRoundStats repo methods

bitcoin-coder-bob:
Issue: https://github.com/arkade-os/arkd/issues/920

The only leaking error message I saw get thorugh was `ErrNoRows` for `GetRoundStats` (we now catch this and return nil) and the in tandem in `internal/core/application/indexer.go` I have added a catch now since we return nil, we catch on this and return `round not found` as an error.

It is possible that other errors like connection failures or constraint violations could still surface, I have not addressed them in this PR but could on request. Is that in or out of scope @altafan?

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized not-found behavior for missing rounds/batches and clarified error responses.
  * Get-round statistics now cleanly returns empty/no-data results instead of erroring for nonexistent rounds.

* **Tests**
  * Added tests covering various nonexistent-round query scenarios to ensure consistent behavior.

### DIFF
--- a/internal/core/application/indexer.go
+++ b/internal/core/application/indexer.go
@@ -59,7 +59,7 @@ func (i *indexerService) GetCommitmentTxInfo(
 		return nil, err
 	}
 	if roundStats == nil {
-		return nil, fmt.Errorf("round not found")
+		return nil, fmt.Errorf("batch not found")
 	}
 
 	batches := map[VOut]Batch{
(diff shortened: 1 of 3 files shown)


tiero:
Is "round not found" public facing error being lifted? Let's use batch then (I'm ok with codebase not being updated yet) q

PR #933, opened 2026-02-25
persist collected fees

### LINKED ISSUE #859: Persist collected fees
## Summary

The operator currently calculates collected fees on-the-fly during batch finalization for alerts, but does not persist this data to the database. This means fee revenue information is ephemeral and lost after the alert is published.

## Current Behavior

Fees are calculated in `internal/core/application/alert.go` during `getBatchStats()`:

```go
a.CollectedFees += intent.TotalInputAmount() + a.BoardingInputAmount - intent.TotalOutputAmount()
```

This value is only used in the `BatchFinalizedAlert` struct and sent to alert subscribers. It is never stored persistently.

## Problem

- No historical record of fee revenue
- Cannot query collected fees over time periods
- No way to generate revenue reports for the operator
- Fee data is lost if alerts are not captured externally

## Proposed Solution

Persist collected fees to the database, possibly:

1. Add a `collected_fees` column to the rounds table
2. Store per-round fee totals in a dedicated fees table
3. Expose via admin gRPC API for revenue tracking

## Code References

- `internal/core/ports/alerts.go:35` — `CollectedFees` field in `BatchFinalizedAlert`
- `internal/core/application/alert.go:78` — Fee calculation logic
- `internal/infrastructure/alertsmanager/service.go:225-228` — Alert formatting

## Impact

Medium priority. Operators need visibility into fee revenue for business operations, but the protocol functions correctly without this feature.

bitcoin-coder-bob:
closes https://github.com/arkade-os/arkd/issues/859

- Add GetCollectedFees admin endpoint (GET /v1/admin/fees/collected) that returns aggregate collected fees across batches, with optional time-range filtering (after/before unix timestamps)

- a `calculateCollectedFees` function that correctly accounts for boarding input amounts (previously fees were accumulated per-intent without including boarding inputs, leading to incorrect values I believe)

Bug breakdown: 

prior: we were adding the boding input amount for each intent. 
```
	for _, intent := range round.Intents {
		a.CollectedFees += intent.TotalInputAmount() + a.BoardingInputAmount - intent.TotalOutputAmount()
```

now: we add the boardingInput amount one time
```
	totalIn := boardingInputAmount
	totalOut := uint64(0)
	for _, intent := range round.Intents {
		totalIn += intent.TotalInputAmount()
		totalOut += intent.TotalOutputAmount()
	}
```

- fix boarding input detection: the `BoardingInputAmount` is now computed from the PSBT taproot leaf scripts rather than being passed incorrectly

- `GetCollectedFees` function added to round repo. Queries db using `SelectCollectedFees` which is a new query, that runs on the `round` table.

`func (r *Round) EndFinalization` in `internal/core/domain/round.go` modified to take in a new param `collectedFees uint64` and is added as a new field into the `RoundFinalized` struct.

The `Round` struct also has a new `Collected Fees uint64` field which is leveraged in the `GetCollectedFees` functions.

`e2e_test.go` additions for hitting new admin endpoint.

`alert_test.go`  additions for testing against `calculateCollectedFees`.

NOTE: All fees are expressed in sats.

I found I had to increase the timeout in the Makefile's `integrationtest` and needed to bump the `otel` package version due to a CVE in our version used causing CI to fail

## Summary by CodeRabbit

* **New Features**
  * Added admin API endpoint /v1/admin/fees/collected (GET) to query total collected fees within a time window via after/before query parameters.

* **Improvements**
  * Persist collected fees for rounds so totals reflect actual data.
  * Improved handling of boarding and fee calculation to produce accurate totals.

* **Tests**
  * Added unit and end-to-end tests covering collected-fees queries and calculation logic.


.github/workflows/unit.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
       - name: check code integrity
         uses: securego/gosec@master
         with:
-          args: '-severity high -quiet -exclude=G115 ./...'
+          args: '-severity high -quiet -exclude=G115,G118 ./...'
```
louisinger:
```suggestion
          args: '-severity high -quiet -exclude=G115 ./...'
```
isn't fixed by master ?

api-spec/protobuf/ark/v1/admin.proto
```diff
@@ -120,6 +120,11 @@ service AdminService {
       get: "/v1/admin/liquidity/recoverable"
     };
   }
+  rpc GetCollectedFees(GetCollectedFeesRequest) returns (GetCollectedFeesResponse) {
```
louisinger, line 135: not sure we need this cause it's subset of GetRounds ? we can sum the fees client side maybe ? I'd remove it to simplify except if it's a feature request from @Kukks / @altafan

internal/core/application/alert.go
```diff
@@ -117,3 +117,31 @@ func (s *service) getBatchStats(
...
+// calculateBoardingInputAmount computes the total amount (sats) of boarding inputs in a PSBT.
+func calculateBoardingInputAmount(ptx *psbt.Packet) uint64 {
+	boardingInputAmount := uint64(0)
+	for _, input := range ptx.Inputs {
+		if input.WitnessUtxo == nil {
+			continue
+		}
+		if len(input.TaprootLeafScript) > 0 {
+			boardingInputAmount += uint64(input.WitnessUtxo.Value)
+		}
```
louisinger:
```suggestion
    // TODO fragile, it may fail if arkd-wallet uses TaprootLeafScript in the future
		if len(input.TaprootLeafScript) > 0 {
			boardingInputAmount += uint64(input.WitnessUtxo.Value)
		}
```

louisinger: that's something we do everywhere, it's ok now but let's add TODO so we don't forget it's not ideal if arkd-wallet becomes a tapscript wallet

internal/core/application/alert.go
```diff
@@ -117,3 +117,31 @@ func (s *service) getBatchStats(
 	a.IntentsCount = len(round.Intents)
 	return
 }
+
+// calculateCollectedFees computes the total fees (sats) collected by the coordinator for a given round.
+func calculateCollectedFees(round *domain.Round, boardingInputAmount uint64) uint64 {
```
louisinger: this file is more for ports.Alerts methods, let's move the fee funcs to utils.go ?

internal/core/domain/round_event.go
```diff
@@ -33,6 +33,7 @@ type RoundFinalized struct {
 	RoundEvent
 	ForfeitTxs        []ForfeitTx
 	FinalCommitmentTx string
+	CollectedFees     uint64
```
louisinger:
```suggestion
	Fees     uint64
```
nit: Fees are always "collected" from arkd point of view no ?

internal/infrastructure/db/postgres/migration/20260225000000_add_collected_fees.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE round ADD COLUMN collected_fees BIGINT NOT NULL DEFAULT 0 CHECK (collected_fees >= 0);
```
louisinger:
```suggestion
ALTER TABLE round ADD COLUMN fees BIGINT NOT NULL DEFAULT 0 CHECK (collected_fees >= 0);
```
nit

internal/core/domain/round.go
```diff
@@ -53,6 +53,7 @@ type Round struct {
 	Version            uint
 	Swept              bool
 	VtxoTreeExpiration int64
+	CollectedFees      uint64
```
louisinger, line 56: should be big.Int to handle overflow

internal/test/e2e/e2e_test.go
```diff
@@ -4196,6 +4196,73 @@ func TestFee(t *testing.T) {
...
+	go func() {
+		_, incomingErr = alice.NotifyIncomingFunds(ctx, aliceOffchainAddr)
+		wg.Done()
+	}()
+
+	var settleErr error
+	go func() {
+		_, settleErr = alice.Settle(ctx)
+		wg.Done()
+	}()
```
louisinger: let's add another participant to that batch, doing a renewal (not boarding). to verify the fees equal the sum and check offchain settlement are also included

pkg/client-lib/client/grpc/client.go
```diff
@@ -268,6 +268,25 @@ func (a *grpcClient) GetEventStream(
 		defer close(eventsCh)
 		backoffDelay := utils.GrpcReconnectConfig.InitialDelay
 
+		send := func(ev client.BatchEventChannel) bool {
```
louisinger: can we open a dedicated PR for this ? it sounds like an enhancement

louisinger (changes requested):

api-spec/protobuf/ark/v1/admin.proto
bitcoin-coder-bob: was asked in the issue: `Expose via admin gRPC API for revenue tracking`

.github/workflows/unit.yaml
bitcoin-coder-bob: did something change on master? I was having tests timeout without this

internal/core/domain/round_event.go
bitcoin-coder-bob: only argument against that is to distinguish it from on chain fees?

internal/core/domain/round.go
bitcoin-coder-bob: given bitcoins supply cap is this even a concern? I think adding this type adds complexity for all the places we then end up having to use it.

pkg/client-lib/client/grpc/client.go
bitcoin-coder-bob: https://github.com/arkade-os/arkd/pull/953

.github/workflows/unit.yaml
louisinger: tbh I don't remember, but I think it's fixed. if it pass without it, let's revert

pkg/client-lib/client/grpc/client.go
louisinger: nice ! approved there but let's revert the file in that branch

internal/core/application/alert_test.go
```diff
@@ -0,0 +1,85 @@
...
+		receivers := make([]domain.Receiver, len(outputs))
+		for i, a := range outputs {
+			receivers[i] = domain.Receiver{Amount: a}
+		}
+		return domain.Intent{Inputs: vtxos, Receivers: receivers}
+	}
+
+	t.Run("no_intents_no_boarding", func(t *testing.T) {
+		round := makeRound(nil)
+		require.Equal(t, uint64(0), calculateCollectedFees(round, 0))
```
louisinger:
```suggestion
		require.Equal(t, 0, int(calculateCollectedFees(round, 0)))
```

we try to do this everywhere in unit tests because uint are printed as hex instead of base 10 numbers. so in requires statements, comparing "int" makes debugging easier.


I see several places in the files where we can compare "int" instead of "uint"

internal/core/domain/round.go
louisinger: u right, fee amount in theory is small so uint64 can fit a lot of intents. but today fees are satoshis, maybe one day it will be some shitcoins or something causing overflow. it's just an extra safety. at least in database we can use bigint ? 

wdyt @altafan ?

internal/infrastructure/db/postgres/round_repo.go
```diff
@@ -77,6 +78,9 @@ func (r *roundRepository) GetRoundIds(
 }
 
 func (r *roundRepository) AddOrUpdateRound(ctx context.Context, round domain.Round) error {
+	if round.CollectedFees > uint64(math.MaxInt64) {
+		return fmt.Errorf("collected_fees %d overflows int64", round.CollectedFees)
+	}
```
louisinger:
```suggestion
```

we need to find a way to remove this. I understand u cast to int64 later to set UpsertRound params but in theory the column is `BIGINT` so should not be int64 ? sqlc generated code problem or missing config maybe ? 

but having BIGINT column, uint64 domain field but int64 integer limit is confusing

louisinger: (same in sqlite)

internal/infrastructure/db/postgres/round_repo.go
```diff
@@ -445,6 +450,22 @@ func (r *roundRepository) GetIntentByTxid(
...
+		After:  after,
+		Before: before,
+	})
+	if err != nil {
+		return 0, err
+	}
+	if fees < 0 {
+		return 0, fmt.Errorf("data integrity issue: got negative collected_fees %d", fees)
+	}
+	return uint64(fees), nil
```
louisinger: column is BIGINT but return int64. I think we encountered this for asset supply column, worth double checking how we handle this there

internal/infrastructure/db/sqlite/round_repo.go
```diff
@@ -77,6 +78,9 @@ func (r *roundRepository) GetRoundIds(
 }
 
 func (r *roundRepository) AddOrUpdateRound(ctx context.Context, round domain.Round) error {
+	if round.CollectedFees > uint64(math.MaxInt64) {
+		return fmt.Errorf("collected_fees %d overflows int64", round.CollectedFees)
+	}
```
louisinger:
```suggestion

```

internal/infrastructure/db/sqlite/migration/20260225000000_add_collected_fees.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE round ADD COLUMN fees INTEGER NOT NULL DEFAULT 0 CHECK (fees >= 0);
```
louisinger:
```suggestion
ALTER TABLE round ADD COLUMN fees REAL NOT NULL DEFAULT 0 CHECK (fees >= 0);
```
"REAL" is BIGINT in sqlite (for consistency with pg migration)

internal/core/domain/round_test.go
```diff
@@ -487,6 +487,38 @@ func testEndFinalization(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Len(t, events, 1)
+			require.True(t, round.IsEnded())
+			require.Equal(t, uint64(42000), round.CollectedFees)
+
+			event, ok := events[0].(domain.RoundFinalized)
+			require.True(t, ok)
+			require.Equal(t, uint64(42000), event.Fees)
+		})
+
```
louisinger: move to t.Run("valid"), we can use the existing round there.

internal/infrastructure/db/service_test.go
```diff
@@ -182,6 +182,7 @@ func TestService(t *testing.T) {
 			// records are added before the asset ones, and vtxos are added after assets.
 			testEventRepository(t, svc)
 			testRoundRepository(t, svc)
+			testCollectedFeesRepository(t, svc)
```
louisinger: there is no collectedFeesRepository, the new tests should be inside testRoundRepository

louisinger (changes requested):

internal/infrastructure/db/service_test.go
bitcoin-coder-bob: done in: [d0458d8](https://github.com/arkade-os/arkd/pull/933/commits/d0458d86a944332c2d84e919eb3a169503669fa8)

internal/core/domain/round_test.go
bitcoin-coder-bob: done in: [d0458d8](https://github.com/arkade-os/arkd/pull/933/commits/d0458d86a944332c2d84e919eb3a169503669fa8) by updating `end_finalization` test

internal/infrastructure/db/sqlite/migration/20260225000000_add_collected_fees.up.sql
bitcoin-coder-bob: done in: [d0458d8](https://github.com/arkade-os/arkd/pull/933/commits/d0458d86a944332c2d84e919eb3a169503669fa8)

internal/infrastructure/db/sqlite/round_repo.go
bitcoin-coder-bob: done in: [d0458d8](https://github.com/arkade-os/arkd/pull/933/commits/d0458d86a944332c2d84e919eb3a169503669fa8)

internal/core/application/alert_test.go
bitcoin-coder-bob: done in: [d0458d8](https://github.com/arkade-os/arkd/pull/933/commits/d0458d86a944332c2d84e919eb3a169503669fa8)

internal/core/domain/round.go
bitcoin-coder-bob: Even if shitcoins enter the picture I reckon those would likely be tracked as separate asset denominations in some field that indicates the shitcoin

internal/infrastructure/db/postgres/round_repo.go
bitcoin-coder-bob: heres how we do in `internal/infrastructure/db/postgres/asset_repo.go`: https://github.com/arkade-os/arkd/blob/9ac7220f14e3fb7a7380acf5384787afec4abf37/internal/infrastructure/db/postgres/asset_repo.go#L140-L143

so appears we are getting back supply as a string from the query then we parse it. That approach completely avoids the int64 boundary but is much heavier IMO.

internal/infrastructure/db/postgres/round_repo.go
bitcoin-coder-bob: ok I think I have to change CollectedFees / Fees from uint64 to int64 across the entire stack then. Removes confusion (in theory) but at the cost of a relatively large repo wide change. Attempted to do this in commit: [c2c3447](https://github.com/arkade-os/arkd/pull/933/commits/c2c3447b8eac9b8a286ffda3e51c7883c6185410)

PR #935, opened 2026-02-26
client-lib: fee support in RedeemNote

louisinger:
related to https://github.com/arkade-os/go-sdk/issues/98

@altafan please review 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redemption fee handling: fees are now assessed per note and validated so they never exceed note value; individual note contributions and the final redeemable amount are adjusted accordingly.
  * Added off-chain address validation and output-fee calculation to ensure accurate, fee-aware final transfers.

### DIFF
--- a/pkg/client-lib/batch_session.go
+++ b/pkg/client-lib/batch_session.go
@@ -108,7 +108,7 @@ func (a *service) RedeemNotes(
 
 		feesSats := uint64(fees.ToSatoshis())
 		noteValue := uint64(v.Value)
-		if feesSats > noteValue {
+		if feesSats >= noteValue {
 			// not profitable to redeem this note
 			return "", fmt.Errorf("fees (%d) exceed note value (%d)", feesSats, noteValue)
 		}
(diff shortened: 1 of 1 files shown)


pkg/client-lib/batch_session.go
```diff
@@ -88,7 +98,22 @@ func (a *service) RedeemNotes(
...
-		amount += uint64(v.Value)
+		fees, err := feeEstimator.EvalOffchainInput(arkfee.OffchainInput{
+			Amount: uint64(v.Value),
+			Type:   arkfee.VtxoTypeNote,
+		})
+		if err != nil {
+			return "", err
+		}
+
+		feesSats := uint64(fees.ToSatoshis())
```
IsaqueFranklin, line 109: Initially, I was concerned about using uint64 here for the value, because the `feeEstimator` returns a float64 and it would just truncate the number down, which would mean that the fee would be less than what the operator needs and the redeem would not work. But I checked the `ToSatoshis()` method in `pkg/ark-lib/arkfee/types.go` and it already uses math.Ceil, which rounds the number up, so it's fine.

IsaqueFranklin (approved):
I validated the code using the E2E integration tests. Specifically, I ran `TestBatchSession/redeem_notes` and `TestFee` on this branch, and they both passed successfully.

The fee logic in `RedeemNotes` looks solid.

cACK.

PR #939, opened 2026-02-27
add Settings domain with DB persistence and admin CRUD API

### LINKED ISSUE #934: Add settings domain and repo
Atm we have only some of the config env vars persisted in storage and that can update at runtime via admin apis. 

Let's add a new Settings domain and related repo interface with usual CRUD apis so that we can change at will:
- ban threshold and duration
- delays
- batch max/min participants
- vtxo/utxo min/max amount
- settlement min expiry gap
- vtxo cutoff data to skip csv validation
- max tx weight

The domain can also house the existing domains for fee and scheduled session config (maybe better to keep distinct tables tho?)

All the relative env vars can be dropped alongside with these changes

bitcoin-coder-bob:
closes https://github.com/arkade-os/arkd/issues/934

  - Introduce a Settings domain model that moves 16 runtime-tunable config values (ban threshold/duration, exit delays, VTXO tree expiry, round participant limits, amount limits, etc.) out of environment
   variables and into the database
- Validate all settings fields (amount bounds, participant limits, exit delays, ban thresholds, update_fields field names) with descriptive errors
- Add SettingsRepository with implementations for Postgres, SQLite, and Badger, including SQL migrations
- Add three new Admin gRPC/REST endpoints: GetSettings, UpdateSettings (allows for partial updates), and ClearSettings
- On startup, seed default settings to the DB if none exist (scheduler-aware: block vs time-based defaults)
- Settings changes propagate live to the running app service via an UpdateSettings callback — no restart required
- Remove the corresponding env vars from config, docker-compose, and .env files

## Summary by CodeRabbit

* **New Features**
  * Admin API to get, update, and clear runtime settings with persisted defaults and live-apply support; runtime service now accepts live setting changes.

* **Documentation**
  * Added Admin Settings API docs and comprehensive settings table; removed several legacy environment variables from docs.

* **Chores**
  * Simplified local run-simulation startup (fixed session duration) and adjusted compose/dev envs (removed some env vars, added CSV allow flag).

* **Tests**
  * Added unit tests for settings repo/service and updated E2E setup to apply admin settings automatically.


bitcoin-coder-bob:
addressed the comments from @arkanaai (https://github.com/arkade-os/arkd/pull/939#issuecomment-4085554982) in: [5271785](https://github.com/arkade-os/arkd/pull/939/commits/5271785e94b6e3d4f527879c40a7e1dd2b01736d)

README.md
```diff
@@ -109,6 +99,27 @@ The `arkd` server can be configured using environment variables.
...
+| `round_max_participants_count`    | Maximum number of participants per round                     | `128`                          |
+| `vtxo_min_amount`                 | Minimum allowed amount for vtxos                             | `-1` (dust)                    |
+| `vtxo_max_amount`                 | Maximum allowed amount for vtxos                             | `-1` (unset)                   |
+| `utxo_min_amount`                 | Minimum allowed amount for boarding or collaborative exit    | `-1` (dust)                    |
+| `utxo_max_amount`                 | Maximum allowed amount for boarding or collaborative exit    | `-1` (unset)                   |
+| `ban_duration`                    | Ban duration in seconds                                      | `300` (5 minutes)              |
+| `ban_threshold`                   | Number of crimes to trigger a ban                            | `3`                            |
+| `max_tx_weight`                   | Max transaction weight                                       | `40000`                        |
+
```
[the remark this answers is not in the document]
bitcoin-coder-bob: done in: [a61a9f1](https://github.com/arkade-os/arkd/pull/939/commits/a61a9f1a0c0ff2c5ad65a6bf5b9421b41e46212e)

internal/core/application/service.go
```diff
@@ -292,6 +292,62 @@ func (s *service) Start() error {
...
 }
 
+const minAllowedSequence = 512
+
+func toRelativeLocktime(locktime int64) arklib.RelativeLocktime {
+	if locktime >= minAllowedSequence {
+		return arklib.RelativeLocktime{Type: arklib.LocktimeTypeSecond, Value: uint32(locktime)}
+	}
+	return arklib.RelativeLocktime{Type: arklib.LocktimeTypeBlock, Value: uint32(locktime)}
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: done in: [a61a9f1](https://github.com/arkade-os/arkd/pull/939/commits/a61a9f1a0c0ff2c5ad65a6bf5b9421b41e46212e)

internal/interface/grpc/handlers/adminservice.go
```diff
@@ -552,6 +552,84 @@ func (a *adminHandler) ClearIntentFees(
...
+		VtxoNoCsvValidationCutoffDate: s.GetVtxoNoCsvValidationCutoffDate(),
+		MaxTxWeight:                   s.GetMaxTxWeight(),
+	}
+
+	if err := a.adminService.UpdateSettings(ctx, settings); err != nil {
+		return nil, status.Errorf(codes.Internal, "%s", err.Error())
+	}
+
+	return &arkv1.UpdateSettingsResponse{}, nil
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: validation added in: [5271785](https://github.com/arkade-os/arkd/pull/939/commits/5271785e94b6e3d4f527879c40a7e1dd2b01736d)

internal/infrastructure/db/service.go
```diff
@@ -214,6 +226,10 @@ func NewService(config ServiceConfig, txDecoder ports.TxDecoder) (ports.RepoMana
 		if err != nil {
 			return nil, fmt.Errorf("failed to create intent fees store: %w", err)
 		}
+		settingsStore, err = settingsStoreFactory(config.DataStoreConfig...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create settings store: %w", err)
+		}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: pre-existing pattern, i'm not changing this

internal/infrastructure/db/badger/settings_repo.go
```diff
@@ -0,0 +1,95 @@
...
+			for errors.Is(err, badger.ErrConflict) && attempts <= maxRetries {
+				time.Sleep(100 * time.Millisecond)
+				err = r.store.Upsert(settingsKey, &settings)
+				attempts++
+			}
+		}
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: done in: https://github.com/arkade-os/arkd/pull/939#discussion_r2956293881

internal/core/application/admin.go
```diff
@@ -607,6 +613,32 @@ func (a *adminService) GetRecoverableLiquidity(ctx context.Context) (uint64, err
...
+	a.onInfoChange()
+	return nil
+}
+
+func (a *adminService) ClearSettings(ctx context.Context) error {
+	if err := a.repoManager.Settings().Clear(ctx); err != nil {
+		return err
+	}
+	a.onInfoChange()
+	return nil
```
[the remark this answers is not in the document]
bitcoin-coder-bob: done in: [5271785](https://github.com/arkade-os/arkd/pull/939/commits/5271785e94b6e3d4f527879c40a7e1dd2b01736d)

internal/core/application/admin.go
```diff
@@ -607,6 +613,32 @@ func (a *adminService) GetRecoverableLiquidity(ctx context.Context) (uint64, err
...
+	if err := a.repoManager.Settings().Upsert(ctx, settings); err != nil {
+		return err
+	}
+	if a.onSettingsUpdated != nil {
+		if err := a.onSettingsUpdated(ctx, settings); err != nil {
+			return err
+		}
+	}
+	a.onInfoChange()
+	return nil
```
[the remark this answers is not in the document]
bitcoin-coder-bob: done in: [5271785](https://github.com/arkade-os/arkd/pull/939/commits/5271785e94b6e3d4f527879c40a7e1dd2b01736d)

bitcoin-coder-bob:
for the commnets by @arkanaai in: https://github.com/arkade-os/arkd/pull/939#issuecomment-4085756580 they were addressed in commit: [557ccad](https://github.com/arkade-os/arkd/pull/939/commits/557ccad68b7d3fb8ced1bd46b6bd495067e88be6)

internal/core/domain/settings.go
```diff
@@ -0,0 +1,120 @@
...
+// Values >= MinAllowedSequence (512) are interpreted as seconds (BIP68
+// time-based relative locktimes use 512-second granularity units), while values
+// < 512 are interpreted as blocks. Config validation enforces that time-based
+// values are multiples of 512.
+func ToRelativeLocktime(locktime int64) arklib.RelativeLocktime {
+	if locktime >= MinAllowedSequence {
+		return arklib.RelativeLocktime{Type: arklib.LocktimeTypeSecond, Value: uint32(locktime)}
+	}
+	return arklib.RelativeLocktime{Type: arklib.LocktimeTypeBlock, Value: uint32(locktime)}
+}
```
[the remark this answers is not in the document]
bitcoin-coder-bob: done in: [d382b18](https://github.com/arkade-os/arkd/pull/939/commits/d382b18ea3056cfd6433fcaecde4cb30002e8072)

ghost (changes requested):
## 🔍 Arkana Review — Iterative (commits `1fc333e`, `2f660ca`)

**Delta since last review:** Remove unused `defaultAllowCSVBlockType` config var + seed DB-backed settings from env vars on first boot.

---

### 🔴 CRITICAL — Env-var seeded defaults bypass `domain.Settings.Validate()`

**`config.go:515-518`** — `loadSettings()` calls `defaultSettings()` → `Upsert()` **without** calling `settings.Validate()` first. The admin `UpdateSettings` API correctly calls `Validate()` before persisting, but the first-boot seed path does not.

`Config.Validate()` (called after `loadSettings()`) checks locktime type consistency and `BanDuration >= 1`, but **misses** these domain-level constraints:

| Constraint | Example bad env config |
|---|---|
| `PublicUnilateralExitDelay >= UnilateralExitDelay` | `PUBLIC=5, UNILATERAL=100` → persisted, fund safety issue |
| `BoardingExitDelay != UnilateralExitDelay` | Both set to `86400` → persisted, identical script paths |
| `RoundMaxParticipantsCount >= RoundMinParticipantsCount` | `MIN=128, MAX=1` → persisted |
| `CheckpointExitDelay > 0` | Set to `0` → persisted |
| uint32 overflow guards | Values > 2^32 → persisted |
| Satoshi amount bounds | `VtxoMinAmount=99999999999` → persisted |

**Fix:** Add `settings.Validate()` before the `Upsert` call:
```go
if settings == nil {
    settings = c.defaultSettings()
    if err := settings.Validate(); err != nil {
        return fmt.Errorf("invalid default settings: %w", err)
    }
    if err := c.repo.Settings().Upsert(ctx, *settings); err != nil {
        return fmt.Errorf("failed to seed default settings: %w", err)
    }
}
```

This is **protocol-critical** — invalid exit delays affect VTXO script safety and forfeit path timing. 🚨 Requires human review.

---

### 🟡 E2E test behavioral change — exit delay values differ

The removed `updateSettings()` helper used `unilateral_exit_delay: 512` and `boarding_exit_delay: 1024`. The new docker-compose env vars use `20` and `30` respectively. Since the scheduler is `block` type and these are block counts, the new values are valid and arguably more appropriate for e2e tests (shorter blocks = faster tests). But it's a behavioral change — any e2e test that depends on specific delay arithmetic (e.g., waiting for a certain number of blocks for an exit) will see different behavior. Verify the full e2e suite passes with these new values.

---

### 🟢 Good

- **First-boot env var seeding** is the right UX — operators can configure via env vars on initial deploy without needing a second admin API call.
- **`viper.IsSet()` guards** correctly distinguish "env var explicitly set" from "use scheduler-aware default." The two scheduler-dependent fields (`VtxoTreeExpiry`, `CheckpointExitDelay`) fall through to inline defaults when no env var is set.
- **Comment quality** is good — the doc comment on `defaultSettings()` clearly explains the first-boot-only semantics.
- **Removing `updateSettings()` from e2e** simplifies test setup and makes tests more representative of real deployments.

---

### Summary

One blocker: call `settings.Validate()` on seeded defaults before persisting (security-critical for exit delays). Everything else in this delta is solid.

**Reviewer:** Arkana (automated)  
**Requires:** Human sign-off (protocol-critical: exit delay handling on first boot)

bitcoin-coder-bob:
> ## 🔍 Arkana Review — Iterative (commits `1fc333e`, `2f660ca`)
> **Delta since last review:** Remove unused `defaultAllowCSVBlockType` config var + seed DB-backed settings from env vars on first boot.
> 
> ### 🔴 CRITICAL — Env-var seeded defaults bypass `domain.Settings.Validate()`
> **`config.go:515-518`** — `loadSettings()` calls `defaultSettings()` → `Upsert()` **without** calling `settings.Validate()` first. The admin `UpdateSettings` API correctly calls `Validate()` before persisting, but the first-boot seed path does not.
> 
> `Config.Validate()` (called after `loadSettings()`) checks locktime type consistency and `BanDuration >= 1`, but **misses** these domain-level constraints:
> 
> Constraint	Example bad env config
> `PublicUnilateralExitDelay >= UnilateralExitDelay`	`PUBLIC=5, UNILATERAL=100` → persisted, fund safety issue
> `BoardingExitDelay != UnilateralExitDelay`	Both set to `86400` → persisted, identical script paths
> `RoundMaxParticipantsCount >= RoundMinParticipantsCount`	`MIN=128, MAX=1` → persisted
> `CheckpointExitDelay > 0`	Set to `0` → persisted
> uint32 overflow guards	Values > 2^32 → persisted
> Satoshi amount bounds	`VtxoMinAmount=99999999999` → persisted
> **Fix:** Add `settings.Validate()` before the `Upsert` call:
> 
> ```go
> if settings == nil {
>     settings = c.defaultSettings()
>     if err := settings.Validate(); err != nil {
>         return fmt.Errorf("invalid default settings: %w", err)
>     }
>     if err := c.repo.Settings().Upsert(ctx, *settings); err != nil {
>         return fmt.Errorf("failed to seed default settings: %w", err)
>     }
> }
> ```
> 
> This is **protocol-critical** — invalid exit delays affect VTXO script safety and forfeit path timing. 🚨 Requires human review.
> 
> ### 🟡 E2E test behavioral change — exit delay values differ
> The removed `updateSettings()` helper used `unilateral_exit_delay: 512` and `boarding_exit_delay: 1024`. The new docker-compose env vars use `20` and `30` respectively. Since the scheduler is `block` type and these are block counts, the new values are valid and arguably more appropriate for e2e tests (shorter blocks = faster tests). But it's a behavioral change — any e2e test that depends on specific delay arithmetic (e.g., waiting for a certain number of blocks for an exit) will see different behavior. Verify the full e2e suite passes with these new values.
> 
> ### 🟢 Good
> * **First-boot env var seeding** is the right UX — operators can configure via env vars on initial deploy without needing a second admin API call.
> * **`viper.IsSet()` guards** correctly distinguish "env var explicitly set" from "use scheduler-aware default." The two scheduler-dependent fields (`VtxoTreeExpiry`, `CheckpointExitDelay`) fall through to inline defaults when no env var is set.
> * **Comment quality** is good — the doc comment on `defaultSettings()` clearly explains the first-boot-only semantics.
> * **Removing `updateSettings()` from e2e** simplifies test setup and makes tests more representative of real deployments.
> 
> ### Summary
> One blocker: call `settings.Validate()` on seeded defaults before persisting (security-critical for exit delays). Everything else in this delta is solid.
> 
> **Reviewer:** Arkana (automated) **Requires:** Human sign-off (protocol-critical: exit delay handling on first boot)

i added the `settings.Validate()` per suggestion. done in: [29e4f2b](https://github.com/arkade-os/arkd/pull/939/commits/29e4f2b112466d36e7227360406a6db78f1e9980)

Per the 512/1024 seconds the change i made was intentional to remove the out of band `updateSettings()` can and let env vars be single source of truth.

ghost (approved):
## 🔍 Arkana Review — Commit `29e4f2b` (validate seeded defaults)

**Delta since last review:** Single commit addressing the critical finding from my previous review.

### ✅ Fix confirmed — `config.go:519-522`

`loadSettings()` now calls `settings.Validate()` before `Upsert()` on the first-boot seed path. This closes the gap where a bad env-var combination (e.g. `PublicUnilateral < Unilateral`, `RoundMin > RoundMax`) could be silently persisted to the DB without validation.

The fix is correct:
- Same `Validate()` call used by the admin `UpdateSettings` path
- Error is returned with descriptive wrapping (`"invalid default settings: %w"`)
- Placement is right: after `defaultSettings()`, before `Upsert()`

No new findings. All prior critical/high issues have been addressed across the review iterations.

**⚠️ Requires:** Human sign-off (protocol-critical: settings control exit delays, locktime values, and round lifecycle parameters)

**Reviewer:** Arkana (automated)

ghost (commented):
## 🔍 Arkana Review — Iterative (commits `0f734ec`, `5142125`, `ece9dd1`)

**Reviewer:** Arkana (automated)

**Changes since last review (`29e4f2b`):**
- Two merge-master commits (conflict resolution)
- `5142125`: Removes 3 unused default constants (`defaultRoundMaxParticipantsCount`, `defaultRoundMinParticipantsCount`, `defaultVtxoNoCsvValidationCutoffDate`) that leaked in from the merge

**Assessment:** Clean housekeeping. Verified via `git grep` that the removed constants have zero remaining references on this branch. No new logic, no new risk surface.

**Status:** No new issues. My previous approvals and protocol-critical flag (exit delay handling requiring human sign-off) still stand.

ghost (commented):
## 🔍 Arkana Review — Iterative (commit `99e85fa`)

**Reviewer:** Arkana (automated)

**Delta since last review (`ece9dd1`):** Single merge-from-master commit bringing in ~20 master commits (connection pooling, tx filters, indexer GetSubscription, client-lib refactors, Go 1.26.3, Postgres 17.8, etc.).

### Merge conflict resolutions reviewed

Files with non-trivial conflict resolution:

| File | Verdict |
|------|---------|
| `internal/config/config.go` | ✅ Clean. `StreamConnPoolSize` (from master #1073) correctly integrated alongside settings domain env vars. Default constants, viper defaults, and struct field all present. Bounds clamping (`min/max`) preserved. |
| `internal/core/application/service.go` | ✅ No conflict (auto-merged). The `defer` block relocation from #1063 is orthogonal to settings changes. |
| `docker-compose.regtest.yml` | ✅ Trivial — Postgres 17.8 + nbxplorer 2.6.7 version bumps merged alongside settings env var removals. |
| `envs/arkd.dev.env` | ✅ `ARKD_SESSION_DURATION=10` from master added cleanly. |
| `internal/interface/grpc/permissions/permissions.go` | ✅ Settings admin endpoints (`GetSettings`, `UpdateSettings`, `ClearSettings`) still present at correct positions. |
| `internal/test/e2e/single_batch_smoke_test.go` | ✅ No settings-related conflicts. |
| `api-spec/protobuf/gen/ark/v1/indexer.pb.rgw.go` | ✅ Generated code, no hand-edited conflicts. |

### Assessment

Clean merge. No logic errors introduced by conflict resolution. All prior settings domain code (validation, env-var seeding, admin API endpoints, DB repositories) remains intact.

**Status:** No new issues. Previous approvals and protocol-critical flag (exit delay handling requiring human sign-off) still stand.

pkg/ark-lib/locktime.go
```diff
@@ -20,6 +20,10 @@ const (
 	// The value of 1 sec/block is used so that number of blocks refers to number of seconds.
 	SECONDS_PER_BLOCK = 1
 
+	// MinAllowedSequence is the discriminat above which all relative locktime value sare
```
bitcoin-coder-bob:
```suggestion
	// MinAllowedSequence is the discriminat above which all relative locktime values are
```

altafan: fixed in [06c5bc0](https://github.com/arkade-os/arkd/pull/939/commits/06c5bc07bfc80144059ae3e1aaacd32bcf90894b)

README.md
```diff
@@ -114,6 +105,35 @@ The `arkd` server can be configured using environment variables.
...
 | `ARKD_INDEXER_AUTH_TOKEN_EXPIRY`   | Auth token TTL in seconds                                                       | `300` (5 minutes)              |
 
+#### Admin Settings
+
+The following settings are persisted in the database and managed via the admin API. Default values are seeded on first startup.
+
+| Endpoint                          | Method | Description                              |
+|-----------------------------------|--------|------------------------------------------|
+| `/v1/admin/settings`              | GET    | Retrieve current settings                |
+| `/v1/admin/settings`              | POST   | Update settings (full replace)           |
```
bitcoin-coder-bob: isnt this endpoint partial update? since every field is optional?
```suggestion
| `/v1/admin/settings`              | POST   | Update settings (partial; only provided fields change)          |
```

README.md
```diff
@@ -55,7 +55,9 @@ In this documentation, you'll learn how to install and use `arkd`, a Bitcoin ser
 
 ### Configuration Options
 
-The `arkd` server can be configured using environment variables.
+The `arkd` server can be configured using environment variables and the admin settings API.
```
bitcoin-coder-bob, line 58: the env vars are only used for seeding on first boot. I know we say that on line 110 in this file, but the reader is going to read this line before line 110. I think we need to also call that out here, that these vars are for seeding, and that each susequent boot of arkd the runtime reads the settings from the db and not these env vars

bitcoin-coder-bob: whats unclear to me is that we do not surface a comprehensize list of the first boot seed variables. having a list of these would make it clear what gets set to what right out of the box

internal/core/application/service.go
```diff
@@ -461,6 +412,23 @@ func (s *service) SubmitOffchainTx(
 	}
 	txid := arkPtx.UnsignedTx.TxID()
 
+	settings, err := s.cache.Settings().Get(ctx)
```
bitcoin-coder-bob, line 414: theres a handful of cases in this file where we call `s.cache.Settings().Get(ctx)` and then set a bunch of local variables from the yielded `settings`. It _feels_ like something that should go in its own function and have a corresponding settings struct. For example in this `SubmitOffChainTx` we could call a new function:
```
settings, err := GetSubmitOffChainTxSettings(ctx)
```
```
func (s *service) GetSubmitOffChainTxSettings(ctx context.Context) (settings OffChainTxSettings, err error) {
settings, err := s.cache.Settings().Get(ctx)
	if err != nil {
		return nil, errors.INTERNAL_ERROR.New("failed to get settings: %w", err)
	}
   settings: OffChainTxSettings{
	banThreshold: settings.BanThreshold,
	minAllowedExitDelay: settings.UnilateralExitDelay,
	vtxoNoCsvValidationCutoffDate: settings.VtxoNoCsvValidationCutoffDate,
	signerPubkey: settings.SignerPubkey,
	vtxoMinAmount: settings.VtxoMinAmount,
	vtxoMaxAmount: settings.VtxoMaxAmount,
	checkpointTapscript: settings.CheckpointTapscript,
	maxOpReturnOutputs: settings.MaxOpReturnOutputs,
	maxTxWeight: settings.MaxTxWeight,
	maxAssetsPerVtxo: settings.MaxAssetsPerVtxo(),
	allowCSVBlockType: settings.AllowCSVBlockType(),
}

return settings, nil
}
```

internal/core/ports/live_store.go
```diff
@@ -114,3 +124,67 @@ type ConfirmationSessions struct {
...
+		VtxoMaxAmount:       s.VtxoMaxAmount,
+		CheckpointTapscript: hex.EncodeToString(s.CheckpointTapscript),
+		Fees: feeDigestData{
+			IntentFees: s.BatchFees,
+		},
+		MaxTxWeight:        int64(s.MaxTxWeight),
+		MaxOpReturnOutputs: int64(s.MaxOpReturnOutputs),
+	}
+	// nolint
+	buf, _ := json.Marshal(data)
```
bitcoin-coder-bob: swallowing the error, our prior digest construction code did not swallow the error. I think we need to return the error and upstream where this `Digest()` function is called we will need to handle the error

bitcoin-coder-bob:
I summoned the AI who actually had some valid concerns:

```
Blocking / High

  1. The "singleton" is enforced by nobody, and three code paths disagree on what "the row" is.
  - Get → SelectLatestSettings = ORDER BY updated_at DESC LIMIT 1 (postgres/sqlite query.sql)
  - Upsert → INSERT ... ON CONFLICT(id) DO UPDATE with id defaulting to 0 (Go never sets it)
  - InsertSettingsHistory → hardcodes WHERE s.id = 0
  
  The schema (id GENERATED BY DEFAULT AS IDENTITY / AUTOINCREMENT) places no CHECK/unique guarantee that only one row exists. If a second row ever lands (a stray insert, a future code path that sets id,
  a botched restore), Get follows updated_at while upsert/history follow id=0, and they silently diverge. updated_at is second-granularity (timeToUnix uses .Unix()), so the ORDER BY tiebreak is also
  non-deterministic within a second. This is exactly the failure class a CHECK (id = 1) (or id = 0) singleton constraint prevents, and matches the "restore the singleton CHECK" instinct from earlier.
  Recommendation: enforce one row at the schema level and make all three paths agree on a fixed id (WHERE id = 1), rather than relying on three separate conventions.

  2. Every process restart writes a no-op settings_history row and fires a spurious "settings changed" event.
  Boot path service.go:103-116: resolveMinAmounts is idempotent, but settings.Update({VtxoMinAmount, UtxoMinAmount}) appends to the changelog based purely on pointer-non-nil, never value comparison
  (settings.go:293-308). So the changelog is always ["vtxo_min_amount","utxo_min_amount"] on boot → postgres Upsert sees a non-empty changelog → InsertSettingsHistory runs (settings_repo.go:156-161) → an
  audit row on every restart, even when nothing changed. This directly defeats the stated intent in that same code ("Only record history for actual changes"). It also means the very first history entry
  is a phantom reconciliation of the seed, and any "settings changed" warning logs fire on every boot. Recommendation: gate the write/history on an actual value delta (compare before appending to
  changelog, or only persist if resolveMinAmounts changed something), or compute resolved min-amounts for the cache without persisting.

  3. UpdateSettings is an unserialized read-modify-write, and the cache refresh is a fire-and-forget goroutine whose ordering isn't tied to the DB.
  - adminService.UpdateSettings (admin.go:585-611): Get → settings.Update → Upsert with no lock (confirmed: no mutex in admin.go). Two concurrent admin updates each read the same base and the later
  Upsert clobbers the earlier — classic lost update.
  - postgres/sqlite Upsert ends with go r.dispatch(...) (settings_repo.go:168), and the registered handler (service.go:347-365) does its own unsynchronized cache Get → mutate → Upsert. The cache's
  RWMutex makes each call atomic but not the sequence, and the dispatch goroutines for two updates can run in either order. Net result: the cache can settle on an older value than the DB. DB/cache
  divergence with no reconciliation until restart. Redis doesn't save you here — its Watch wraps an unconditional Set, so it provides no read-check (see nit below).
  
  Settings changes are rare, so the probability is low, but it's real and silent. Recommendation: serialize UpdateSettings (mutex or SELECT ... FOR UPDATE), and make the cache update synchronous/ordered
  relative to the committed write (or refresh the cache from the DB rather than from the in-flight settings value).

  ---
  Medium
  
  4. The UpdateSettings doc comment is not just stale, it's inverted (admin.go:588-592). It describes a removed updateFields mechanism and states "Fields not set in the request default to 0 so callers 
  must populate all fields." The actual behavior is the opposite: pointer-based partial merge, omitted fields are preserved. A maintainer trusting this comment would assume an omitted field zeroes out.
  Fix or delete.

  5. Validation gaps vs. stated intent (settings.go:107-214). MaxSatoshis (line 14) is defined and used nowhere in the repo — the "add satoshi upper bound on amounts" intent never landed. There's no
  vtxo/utxo_max_amount >= min check (you can set max=5, min=100 and it validates), and MaxOpReturnOutputs is unvalidated. Either wire MaxSatoshis + a min/max ordering check, or drop the dead constant.

  6. ClearSettings is a half-wired/dead feature. repo.Clear() (DELETE FROM settings) is implemented in all three backends but called from no non-test code, the permission map references
  /AdminService/ClearSettings (permissions.go:408) for an RPC that doesn't exist (no admin.proto entry, no handler, no app method), and the README documents /v1/admin/settings/clear. Decide: implement
  it, or remove the repo method + permission + README row. Note Clear() is also dangerous given (1): on the current schema it empties the table but the seed only re-runs at boot, so a runtime clear would
  leave Get returning nil.

  7. InsertSettingsHistory ... WHERE s.id = 0 can silently snapshot nothing. If the row's id is ever not 0, the INSERT ... SELECT ... WHERE s.id=0 inserts zero rows: the upsert succeeds, the audit record
  is silently lost. Same root cause as (1).

  8. Inconsistent locking / data race on the handler field. dispatch reads r.updateHandler without updateHandlerMu, which RegisterUpdatesHandler holds when writing (settings_repo.go:42-46, 181-185, same
  in sqlite/badger). Benign in practice (registered once at startup), but it's a genuine race under -race and the mutex is half-used. Use the lock on read, or make it set-once/atomic.

  ---
  Low / nits

  - 9. README (as discussed): "Update settings (full replace)" is wrong (it's partial); the /clear endpoint doesn't exist; and the seed ARKD_* vars were deleted from the env table yet are still
  functional and still used in docker-compose.regtest.yml. They need a "first-boot seed only" note.
  - 10. Terminology drift: domain uses RoundMin/MaxParticipantsCount but the user-facing error strings say "batch min/max participants count" (settings.go:205,209) and the handler uses
  batchMinParticipants. Pick one vocabulary.
  - 11. sqlite AUTOINCREMENT is unnecessary (and slightly costly) for a singleton; AssetTxMaxWeightRatio round-trips float32↔float64 differently across backends (sqlite casts, postgres native). Cosmetic.
  - 12. Redis Upsert (redis/settings.go:36-57) wraps an unconditional Set in Watch/TxPipelined with retry, but never reads the key, so the optimistic-lock machinery buys nothing over a plain Set.
  Harmless, but misleading.
  - 13. inmemory settingsStore.Get returns a zero-value ports.Settings (not an error/nil) when never populated — a footgun if ever read before boot population.
```

internal/infrastructure/db/postgres/settings_seed.go
```diff
@@ -0,0 +1,159 @@
...
+		return fmt.Errorf("read legacy intent_fees: %w", feeErr)
+	}
+
+	// Read latest legacy scheduled_session, if any.
+	var (
+		startUnix, endUnix     int64
+		periodSec, durationSec int64
+		roundMin, roundMax     int64
+	)
+	ssErr := tx.QueryRowContext(ctx, `
```
bitcoin-coder-bob, line 71: can we turn this into a generated query from `query.sql`

altafan: well, actually this and other queries were still in query.sql, and i dropped them because they are queries related to tables that are not there anymore.

PR #957, opened 2026-03-10
Add opt-in digest validation header for stale parameter detection

### LINKED ISSUE #958: Add opt-in digest validation header to detect stale server parameters
## Summary                                                                                                                                                                                               
                                                                      
  `GetInfo` already returns a `digest` field (SHA256 of the JSON-marshalled response). Clients should be able to send this digest as a header on subsequent API calls so arkd can detect when they're operating with stale server parameters.

  ## Behavior

  - Clients optionally attach an `X-Ark-Digest` header (or `x-ark-digest` gRPC metadata) to any API call.
  - If arkd sees the header and the digest doesn't match the current one, it returns a structured `DIGEST_MISMATCH` error (`FailedPrecondition`, code 48) with the `current_digest` in metadata, telling
  the client to re-fetch `GetInfo`.
  - If the header is absent, no validation occurs — completely client opt-in.
  - `GetInfo` itself is always exempt from validation (clients need it to obtain the digest).
  - Only public service methods (ArkService, IndexerService) are validated — admin/wallet calls are unaffected.

bitcoin-coder-bob:
closes https://github.com/arkade-os/arkd/issues/958

## Summary by CodeRabbit

* **New Features**
  * Optional digest-based request validation: servers seed a digest on startup; clients fetch it via GetInfo and may send X-Ark-Digest/x-ark-digest for subsequent calls.
  * Interceptors enforce digest checks for public unary and streaming RPCs; GetInfo is exempt.
  * Stale digests yield a DIGEST_MISMATCH error that includes current_digest metadata.

* **Tests**
  * Added comprehensive unit and end-to-end tests covering unary/streaming validation, exempt cases, matching, and mismatch error details.


altafan:
Replaced by #1104

Issue #959, opened 2026-03-10
Add SDK version compatibility check via header


bitcoin-coder-bob:
Clients (SDKs) currently have no way to signal which arkd server version they were built against. When the server introduces breaking changes to an endpoint (e.g., request/response format changes), clients silently get wrong behavior or cryptic errors.                                                                                                                                                   

Add a version compatibility check so the server can proactively reject outdated clients with a clear, structured error telling them to update.

support both grpc and rest

PR #960, opened 2026-03-10
version in header

bitcoin-coder-bob:
---                                                                                                                                                                                       
  closes https://github.com/arkade-os/arkd/issues/959
                                                                                                                                                                                            
  Design                                                              
                                                                                                                                                                                            
  - Clients send their version via X-Build-Version (REST) / x-build-version (gRPC metadata)                                                                                             
  - Format: semver string, e.g. `1.0.0` or `v1.0.0`
  - Only the major version is compared — minor and patch are ignored                                                                                                                        
  - The server's own build version (set via`-ldflags` in `scripts/build-arkd`) determines the minimum major version                                                                            
  - No header = allow through (won't break existing clients)                                                                                                                                
  - Malformed version strings are allowed through (don't break misbehaving clients)                                                                                                         
  - If the server version can't be parsed (e.g. dev build with "unknown"), enforcement is disabled (major defaults to 0)                                                                    
  - Applies to all services (Ark, Indexer, Admin, Wallet)                                                                                                                                   
                                                                                                                                                                                            
  Changes                                                                                                                                                                                   
                                                                                                                                                                                            
  - New gRPC interceptor (`version.go`) with unary and stream handlers                                                                                                                 
  - Server build version passed through to interceptors at startup    
  - New error code BUILD_VERSION_TOO_OLD (code 48) in pkg/errors                                                                                                                              
  - Interceptor wired into both chains after logger, before auth                                                                                                                            
  - REST gateway forwards the X-Build-Version header to gRPC metadata                                                                                                                     
  - New doc: `docs/version-compat.md`                                                                                                                                                         
  - Unit tests covering: no header, same/above/below major, minor/patch ignored, v prefix, malformed version, unparsable server version, and stream path                                    
                                                                                                                                                                                            
  Usage Examples                                                                                                                                                                            
                                                                                                                                                                                            
  gRPC                                                                
```
  grpcurl -plaintext \
    -H 'x-build-version: 1.0.0' \                                                                                                                                                         
    localhost:7070 ark.v1.ArkService/GetInfo
```
                                                                                                                                                                                            
  REST                                                                
 ```                                                                                                                                                                                           
  curl -X POST http://localhost:7070/v1/batch/registerIntent \        
    -H 'Content-Type: application/json' \
    -H 'X-Build-Version: 1.0.0' \
    -d '{"intent": {...}}'                   
```                                                                                                                                               
  
  Error response (SDK too old)                                                                                                                                                              
                          
```                                            
  {                                                                                                                                                                                         
    "code": 9,                                                        
    "message": "server requires build version header >= 2",
    "details": [{                                                                                                                                                                           
      "code": 48,
      "name": "BUILD_VERSION_TOO_OLD",                                                                                                                                                        
      "metadata": {                                                   
        "client_version": "1.9.9",
        "min_version": "2.0.0"                                                                                                                                                              
      }
    }]                                                                                                                                                                                      
  }                                      
```                             

  Behavior notes

  - Opt-in: Requests without the header are always allowed through.
  - Malformed versions: Clients sending unparseable version strings are silently allowed through.
  - Major only: A server at `v2.5.0` accepts clients at `2.0.0` but rejects `1.9.9`.                                                                                                              
  - Future: Method-level (per-RPC) version constraints are planned but not yet implemented.  We need to have a discussion about how we want to go about that. My current ideas:
  1) add some sort of skill to our AI reviewers to look for function changes, and to see if they require the version to be bumped, and notify the pull requester to make sure they update accordingly.
2) more invasive but add a template into all PR descriptions for a section that specifically asks the pull requester to fill in any function changes and the corresponding version changes to be documented in the PR description. This will both make it easy to read in the PR description but also serve as a reminder to the dev that they do need to update any function version enforcement if their PR made any material changes to functions.

### DIFF
--- a/docs/version-compat.md
+++ b/docs/version-compat.md
@@ -51,7 +51,7 @@ service remains compatible with older clients. This has not been implemented yet
 When a client is rejected the gRPC status contains:
 
 - **Code**: maps to `BUILD_VERSION_TOO_OLD`
-- **Message**: human-readable, e.g. `server requires major build version >= 2`
+- **Message**: human-readable, e.g. `server requires build version header >= 2`
 - **Metadata fields**:
   - `client_version` -- the version string sent by the client
   - `min_version` -- the server's build version

### CODE REVIEW

File: internal/interface/grpc/interceptors/version_compat_test.go [hunk 1]
```diff
@@ -0,0 +1,503 @@
...
+}
+
+func withServiceMinVersion(t *testing.T, minVersion string, fn func()) {
+	t.Helper()
+	serviceMinVersions[testService] = BreakingChange{
+		MinVersion: *semver.New(minVersion),
+		Message:    "service ArkService requires SDK version >= " + minVersion,
+	}
+	defer delete(serviceMinVersions, testService)
+	fn()
```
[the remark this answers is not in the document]
bitcoin-coder-bob: `/ark.v1.ArkService/TestMethod` will never have real functionality, not a concern

internal/interface/grpc/interceptors/version_compat.go
```diff
@@ -0,0 +1,187 @@
...
+
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil
+	}
+
+	vals := md.Get(sdkVersionHeader)
+	if len(vals) == 0 {
+		return nil
+	}
```
louisinger: just to confirm: this make the feature optional right ?

bitcoin-coder-bob: correct. if not set, we return nil thus indicating no version error, making the version header optional

PR #961, opened 2026-03-11
allow unrolled VTXOs to join batches with configurable expiry margin

### LINKED ISSUE #703: Unrolled VTXO lifecycle transitions
* ~~If an unspent VTXO's ancestors are being unrolled, transition it to recoverable. this is based on a proximity config.~~
* If a VTXO was unrolled and its unilateral CSV is still "far", allow it to join a Batch via the collaborative path

bitcoin-coder-bob:
closes https://github.com/arkade-os/arkd/issues/703
                            
Allow unrolled VTXOs to rejoin a batch via the collaborative (boarding) path instead of rejecting them outright. When an unrolled VTXO still has sufficient CSV time remaining before its exit path  becomes available it is treated as a boarding input so it can be swept back into the Ark without requiring unilateral exit.


Why do we want this? An unrolled VTXO is stuck on-chain then the user's funds are out of the Ark and the only way back in was to wait for the CSV to expire, then sweep the funds into a new boarding UTXO. That's slow and costs an extra transaction. With this change, the user can immediately rejoin the Ark by presenting the unrolled VTXO as a boarding input in the next batch, skipping the CSV wait and the extra on-chain step. Funds go from on-chain back to off-chain in a single round.
                                                                                                                                                                                                           
  ### Changes                                                                                                                                                                                              
  
  - **`RegisterIntent`**: Instead of returning `VTXO_ALREADY_UNROLLED`, unrolled VTXOs are now appended as boarding inputs with `isUnrolledVtxo=true` and tracked in `vtxoInputs` for state settlement.    
  - **Boarding validation**: `processBoardingInputs` and `validateBoardingInput` use the unilateral exit delay (not boarding exit delay) when verifying unrolled VTXO scripts and expiry. Reason: Because the VTXO's script was originally built with the unilateral exit delay when it was created inside the Ark's VTXO tree. Unrolling just reveals that output on-chain, it doesn't rewrite the script. The CSV lock embedded in the tapscript is still the unilateral exit delay. The boarding exit delay is a separate value used for UTXOs that users create outside the Ark (boarding UTXOs). Those are constructed by the user's wallet with the boarding-specific delay. So when Validate() checks the script at line 3864, it needs to match what's actually in the tapscript. If we passed `boardingExitDelay`, validation would fail because the on-chain script contains `unilateralExitDelay`.
  - **Expiry safety check**: New `checkUnrolledVtxoExpiry` ensures the VTXO's CSV doesn't expire before the batch can finalize. Margin defaults to session duration but is configurable via `UNROLLED_VTXO_MIN_EXPIRY_MARGIN`.                                                                                                                                                                       
  - **Fee computation**: Unrolled VTXOs are excluded from the VTXO fee computation since they are already counted as onchain/boarding inputs. Reason: The fee manager applies different fee programs to onchain vs offchain inputs (line 36-39). Unrolled VTXOs are already added to boardingInputs (and therefore included in the `onchainInputs` list passed to `ComputeIntentFees` via `boardingInput.witnessUtxo`). If we didn't filter them out of `vtxoInputs`, they'd be double-counted — charged the onchain input fee and the offchain input fee.                                                    
  - **Forfeit skip**: `RequiresForfeit()` returns `false` for unrolled VTXOs since they have no off-chain state to forfeit.                                                                                
  - **Config**: New `UNROLLED_VTXO_MIN_EXPIRY_MARGIN` env var (seconds, default `0` = fallback to session duration). It is the minimum remaining CSV time an unrolled VTXO must have before the server accepts it into a batch. If the VTXO's exit path becomes available too soon, the batch might not finalize in time then the user could unilaterally exit mid-round, taking their funds while also being included in the new VTXO tree. When set to 0 (default), it falls back to `sessionDuration` which is the time a single round takes to complete. A custom value lets operators require a larger buffer if needed.                                                                                      
  - **Tests**: Unit tests for `checkUnrolledVtxoExpiry` and an e2e test covering the full unrolled-VTXO-rejoins-batch flow. 

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Unrolled VTXOs can re-enter collaborative settlement batches.
  * Configurable minimum expiry margin for rejoining VTXOs (UNROLLED_VTXO_MIN_EXPIRY_MARGIN).

* **Bug Fixes**
  * Exclude rejoining VTXOs from intent fee calculations.
  * Stronger expiry checks to prevent VTXOs expiring too soon during settlement.
  * Boarding processing selects correct exit-delay for rejoining VTXOs.
  * Recoverable listings exclude unrolled VTXOs.

* **Tests**
  * Unit tests for expiry-margin logic and an end-to-end test for unrolled VTXO rejoin.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


bitcoin-coder-bob:
@arakanaai regarding your double spend concerns: dual tracking is by design and there's no double-spend vector. The same outpoint enters the PSBT exactly once (as a boarding input). The `vtxoInputs` list never feeds into transaction construction, only into state management.

internal/config/config.go
```diff
@@ -419,6 +425,7 @@ func LoadConfig() (*Config, error) {
 		AllowCSVBlockType:             allowCSVBlockType,
 		RoundReportServiceEnabled:     viper.GetBool(RoundReportServiceEnabled),
 		SettlementMinExpiryGap:        viper.GetInt64(SettlementMinExpiryGap),
+		UnrolledVtxoMinExpiryMargin:   viper.GetInt64(UnrolledVtxoMinExpiryMargin),
```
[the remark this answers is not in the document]
bitcoin-coder-bob: A negative value silently falls back to sessionDuration (0) . None of the other int64 config values have negative checks either

internal/core/domain/vtxo.go
```diff
@@ -67,7 +67,7 @@ func (v Vtxo) IsNote() bool {
 }
 
 func (v Vtxo) RequiresForfeit() bool {
-	return !v.Swept && !v.IsNote()
+	return !v.Swept && !v.IsNote() && !v.Unrolled
```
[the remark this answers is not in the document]
bitcoin-coder-bob: done in: [9999fee](https://github.com/arkade-os/arkd/pull/961/commits/9999fee5e166f7e7c5a9631152e6e9657731fd2b)

internal/test/e2e/e2e_test.go
```diff
@@ -360,6 +360,90 @@ func TestUnilateralExit(t *testing.T) {
...
+
+	wg.Wait()
+	require.NoError(t, incomingErr)
+	time.Sleep(time.Second)
+
+	// Alice has offchain funds again
+	balance, err = alice.Balance(ctx)
+	require.NoError(t, err)
+	require.NotZero(t, balance.OffchainBalance.Total)
+}
```
altafan, line 462: now that we changed the unilateral exit delay to 20 blocks in the testing env (instead of 8 mins), we could try to mine 20 blocks and call `CompleteUnroll` to ensure it's no-op or returns some error like "no funds to claim"?

internal/core/application/service.go
```diff
@@ -3892,6 +3946,32 @@ func (s *service) validateBoardingInput(
...
+	// when a scheduled session window is active. We don't know which window the
+	// next round will land in, so use the larger of the two to stay safe.
+	if s.repoManager != nil {
+		if scheduledSession, _ := s.repoManager.ScheduledSession().
+			Get(ctx); scheduledSession != nil {
+			if scheduledSession.Duration > margin {
+				margin = scheduledSession.Duration
+			}
+		}
+	}
```
altafan: i think we can drop this extra query, no real need for this. Ideally, the scheduled session duration is always smaller than the default, so the latter is always bigger. Moreover, we should take in consideration the scheduled session duration only if we're in the middle of one, but adding such logic seems overkill to me tbh.

I also saw that in config, if the env is left unset we say to default to the session duration. What if, instead, we default to a value like 5 minutes and explictly error if is set to values <= 0 (or < session duration)?

internal/test/e2e/e2e_test.go
bitcoin-coder-bob: done in: [a786cf7](https://github.com/arkade-os/arkd/pull/961/commits/a786cf7dbf23ef0366faa6b7097d8fe5fdfb59bb)

internal/core/application/service.go
bitcoin-coder-bob: attempted fix in: [7785f6d](https://github.com/arkade-os/arkd/pull/961/commits/7785f6dd52173c6f2540eece240c5cec5390ffea)

removed this check, added enforcement of values in `internal/config/config.go`

altafan (approved):

PR #968, opened 2026-03-16
[client-lib] Fix returned value of batch expiry

altafan:
Fixes the returned value of batch expiration in `OnBatchStarted` impl of the default handler used by the client.

Please @louisinger @sekulicd review

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected batch session expiry calculation so block-based locktime values are converted to seconds, while seconds-based expiry values remain unchanged—ensuring accurate timeouts and consistent session behavior across expiry configuration types.

### DIFF
--- a/pkg/client-lib/batch_session_handler.go
+++ b/pkg/client-lib/batch_session_handler.go
@@ -412,7 +412,7 @@ func (h *defaultBatchEventsHandler) OnBatchStarted(
 			h.batchExpiry = getBatchExpiryLocktime(uint32(event.BatchExpiry))
 			expiry := time.Duration(event.BatchExpiry) * time.Second
 			if h.batchExpiry.Type == arklib.LocktimeTypeBlock {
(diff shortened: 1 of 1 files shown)


pkg/client-lib/batch_session_handler.go
```diff
@@ -410,7 +410,11 @@ func (h *defaultBatchEventsHandler) OnBatchStarted(
 			}
 			h.batchSessionId = event.Id
 			h.batchExpiry = getBatchExpiryLocktime(uint32(event.BatchExpiry))
-			return false, time.Duration(event.BatchExpiry) * time.Second, nil
+			expiry := time.Duration(event.BatchExpiry) * time.Second
+			if h.batchExpiry.Type == arklib.LocktimeTypeBlock {
+				expiry = time.Duration(event.BatchExpiry*arklib.SECONDS_PER_BLOCK) * time.Second
```
sekulicd, line 415: Shouldnt seconds per block be 512 ? cause of check in BIP68Sequence?

PR #982, opened 2026-03-18
[arkd]: tests & defensive guards for decodeTx util function 

chris-ricketts:
Fixed out-of-bounds range panic discovered during fuzzing - the trigger input is not possible from external sources as it stands. Adding guards with tests as a defensive measure. 

Changes:
- Add decodeTx guards for empty checkpoint inputs and too-short output scripts to prevent invalid indexing.
- Add table-driven unit tests for decodeTx covering both guard failures and a valid decode path.

## Release Notes

* **Bug Fixes**
  * Validation added to require transactions include at least one input and to enforce a minimum output script length, preventing malformed or empty transaction data from being accepted.

* **Tests**
  * New test coverage covering invalid and valid transaction decoding scenarios to ensure the new validations behave correctly and prevent regressions.

### DIFF
--- a/internal/core/application/utils_test.go
+++ b/internal/core/application/utils_test.go
@@ -50,75 +50,94 @@ func TestDecodeTx(t *testing.T) {
 		[][]byte{{0x51}},
 	))
 
-	tests := []struct {
-		name        string
-		offchainTx  domain.OffchainTx
-		errorSubstr string
-		check       func(t *testing.T, txid string, ins []domain.Outpoint, outs []domain.Vtxo)
-	}{
-		{
-			name: "rejects checkpoint with no inputs",
-			offchainTx: domain.OffchainTx{
-				ArkTx: validArkTx,
-				CheckpointTxs: map[string]string{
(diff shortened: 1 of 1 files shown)


internal/core/application/utils_test.go
```diff
@@ -33,3 +37,112 @@ func TestMaxAssetsPerVtxo(t *testing.T) {
 		)
 	}
 }
+
+func TestDecodeTx(t *testing.T) {
```
altafan, line 41: can you please split this test into 2 sub-tests `valid` and `invalid`, each with its own list of fixtures? this way we can get rid of the check function defined in the fixtures

altafan: example:
```
func TestDecodeTx(t *testing.T) {
   t.Run("valid", func(t *testing.T) {...})
   
   t.Run("invalid", func(t *testing.T) {...})
}
```

chris-ricketts: Fixed in 4870356

PR #999, opened 2026-04-01
Fix: Finalize swept pending tx

louisinger:
we need 2 test cases : 
* finalize a pending tx after it has been swept by the server
* finalize a pending tx after it has expired but was not swept by the server

@altafan please review

## Summary by CodeRabbit

* **Documentation**
  * Updated ARKD_VTXO_TREE_EXPIRY docs with regtest constraint; removed docs for deprecated env vars.

* **Bug Fixes**
  * More accurate detection and reporting of swept vtxos for finalized offchain transactions.

* **Chores**
  * Removed explicit scheduler and CSV-block config options; scheduler chosen from locktime/expiry settings.
  * Block-based relative locktimes now interpreted on a seconds scale.

* **Tests**
  * Added end-to-end tests for finalized swept transaction scenarios.

### DIFF
--- a/internal/core/domain/vtxo_test.go
+++ b/internal/core/domain/vtxo_test.go
@@ -15,7 +15,10 @@ const validVtxoPubkey = "25a43cecfa0e1b1a4f72d64ad15f4cfa7a84d0723e8511c969aa543
 func TestOutpoint(t *testing.T) {
 	t.Run("FromString", func(t *testing.T) {
 		t.Run("valid", func(t *testing.T) {
-			original := domain.Outpoint{Txid: "0123456789abcdef", VOut: 42}
+			original := domain.Outpoint{
+				Txid: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+				VOut: 42,
+			}
 			var parsed domain.Outpoint
 			err := parsed.FromString(original.String())
 			require.NoError(t, err)


bitcoin-coder-bob:
@louisinger I added the 2 test cases you outlined in the PR description

internal/infrastructure/db/service_test.go
```diff
@@ -2024,6 +2024,199 @@ func randomTx() string {
...
+}
+
+// taprootPkScript returns a 34-byte taproot pkscript (OP_1 + PUSH32 + key).
+func taprootPkScript(pubkeyHex string) []byte {
+	pk, _ := hex.DecodeString(pubkeyHex)
+	script := make([]byte, 34)
+	script[0] = 0x51 // OP_1
+	script[1] = 0x20 // PUSH32
+	copy(script[2:], pk)
+	return script
```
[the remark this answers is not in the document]
bitcoin-coder-bob: done in: 088acf5

pkg/ark-lib/locktime.go
```diff
@@ -16,7 +16,9 @@ const (
 	SECONDS_MAX                    = SEQUENCE_LOCKTIME_MASK << SEQUENCE_LOCKTIME_GRANULARITY
 	SEQUENCE_LOCKTIME_DISABLE_FLAG = 1 << 31
 
-	SECONDS_PER_BLOCK = 10 * 60 // 10 minutes
+	// SECONDS_PER_BLOCK is used to convert block-based locktimes to seconds.
+	// The value of 1 sec/block is used so that number of blocks refers to number of seconds.
+	SECONDS_PER_BLOCK = 1
```
bitcoin-coder-bob, line 21: this change makes me nervous. The comment says "number of blocks refers to number of seconds" which is only true for regtest where `VTXO_TREE_EXPIRY=20` is meant to be 20 seconds. Worth adding a stronger warning?

bitcoin-coder-bob: This is currently safe because block-based locktimes are now restricted to regtest but it's a landmine in that if that restriction is ever relaxed then any cross-type locktime comparison would be wildly off

envs/arkd.light.env
```diff
@@ -9,8 +8,9 @@ ARKD_WALLET_ADDR=127.0.0.1:6060
 ARKD_LIVE_STORE_TYPE=inmemory
 ARKD_DB_TYPE=sqlite
 ARKD_EVENT_DB_TYPE=badger
-ARKD_ALLOW_CSV_BLOCK_TYPE=true
 ARKD_SESSION_DURATION=10
 ARKD_VTXO_MIN_AMOUNT=1
 ARKD_BAN_THRESHOLD=1
 ARKD_ONCHAIN_OUTPUT_FEE=100
+ARKD_VTXO_TREE_EXPIRY=20
```
bitcoin-coder-bob: this implies that this env usage is only for regtest now. just want to double check that is the intended result

bitcoin-coder-bob: if so worth commenting this setup for for regtest only (either in this file or somewhere else that talks about these env files and their usages)

internal/infrastructure/db/service.go
```diff
@@ -533,6 +533,15 @@ func (s *service) updateProjectionsAfterOffchainTxEvents(events []domain.Event)
...
 		}
 
+		txSwept := false
+		batch, err := s.roundStore.GetRoundWithCommitmentTxid(ctx, offchainTx.RootCommitmentTxId)
+		// We consider the tx swept if:
+		// - there is an error fetching the batch (this is just fallback, should never happen)
+		// - the batch is swept
+		// - the tx expired (meaning one or all its inputs expired and are already swept or about
+		// to be swept)
+		txSwept = err != nil || (batch != nil && len(batch.SweepTxs) > 0) ||
```
bitcoin-coder-bob, line 543: the addition of `len(batch.SweepTxs) > 0` means any sweep tx in the batch results in this `txSwept` being true. is this overly broad?

internal/core/application/service.go
```diff
@@ -369,6 +369,23 @@ func (s *service) registerEventHandlers() {
 				return
 			}
 
+			// Make sure to mark new vtxos as swept if any of the spent inputs is swept as well or
+			// expired
+			sweptIns := false
+			now := time.Now()
```
bitcoin-coder-bob: the AI flagged this one I thought was worth surfacing. we do this time call here, and also in `internal/infrastructure/db/service.go` where we set `txSwept` around line 543. That is to say we are marking things as swept in two different places, and with the usage of time we could run into race conditions. Is this an alarmist take or worth a conversation?

pkg/ark-lib/locktime.go
altafan: we won't relax that restriction

envs/arkd.light.env
altafan: the envs folder contains only configs for regtest

internal/infrastructure/db/service.go
altafan: no it's not because it may be possible that N blocks are mined before N seconds passed, therefore we need this check and it's also a nice short circuit

internal/core/application/service.go
altafan: the one done in application does not cause any write, it's done only to make sure we set the right swept status to new vtxos of a finalized offchain tx that we broadcast to anybody listening to GetTransactionsStream

bitcoin-coder-bob:
approved if CI passes

internal/core/application/service.go
```diff
@@ -691,7 +710,7 @@ func (s *service) SubmitOffchainTx(
 				"%s already unrolled", vtxo.Outpoint,
 			).WithMetadata(errors.VtxoMetadata{VtxoOutpoint: vtxoOutpoint})
 		}
-		if vtxo.Swept || !s.sweeper.scheduler.AfterNow(vtxo.ExpiresAt) {
+		if vtxo.Swept || now.After(time.Unix(vtxo.ExpiresAt, 0)) {
```
louisinger:
```suggestion
		if vtxo.Swept || vtxo.IsExpired() {
```

let's move this logic to vtxo type so we can reuse it ? also cleaner to read

internal/config/config.go
```diff
@@ -861,11 +865,11 @@ func (c *Config) appService() error {
 	if err := c.txBuilderService(); err != nil {
 		return err
 	}
-
 	roundReportSvc, err := c.RoundReportService()
 	if err != nil {
 		return err
 	}
+	allowCSVBlockType := c.VtxoTreeExpiry.Type == arklib.LocktimeTypeBlock
```
louisinger: can we remove the allowCSVBlockType parameter and use network and/or VtxoTreeExpiry's type inside the application service ?

altafan: yeah sure, i just made the min required change in first place. Happy to drop an arg from the app factory function, they're a lot

internal/core/application/service.go
altafan: fixed in [14fb4c7](https://github.com/arkade-os/arkd/pull/999/commits/14fb4c7c66861585f1bda803606d28fe6691fea4)

internal/config/config.go
altafan: fixed in [f7ee039](https://github.com/arkade-os/arkd/pull/999/commits/f7ee039abcf892cdaaeecf36aae7990d0c07156d)

PR #1008, opened 2026-04-08
[client-lib] Update client wallet interface

altafan:
This replaces the methods to derive and get addresses from the client's wallet interface with raw methods to get and derive new pubkeys (the single key wallet returns always the same keypair). The methods to derive addresses became internal methods of the client.

EDIT: took the occasion to do some cleanup in the e2e tests

Please @sekulicd @bitcoin-coder-bob review.

## Summary by CodeRabbit

* **Refactor**
  * Wallet APIs moved from address-focused to key-centric lifecycle (create/get/list keys); address objects now include a key identifier and use typed address structures.
  * Address derivation and selection centralized into service-level helpers; legacy address-generation APIs removed.

* **Behavioral**
  * Transaction signing and submission flows standardized to use the main client and its derived transport/indexer services.

* **Tests**
  * End-to-end tests unified around a single client setup with automatic cleanup; wallet tests refocused on key lifecycle and lock behavior.


pkg/client-lib/wallet/wallet.go
```diff
@@ -18,17 +18,10 @@ type WalletService interface {
...
-		onchainAddr string, offchainAddr, boardingAddr *types.Address, err error,
-	)
-	NewAddresses(ctx context.Context, change bool, num int) (
-		onchainAddresses []string,
-		offchainAddresses, boardingAddresses []types.Address, err error,
-	)
+	GetKeyPair(
+		ctx context.Context, path string,
+	) (prvkey *btcec.PrivateKey, pubkey *btcec.PublicKey, err error)
+	NewKeyPair(ctx context.Context) (prvkey *btcec.PrivateKey, pubkey *btcec.PublicKey, err error)
```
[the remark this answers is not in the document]
tiero: I agree that we should at very least give out something or. Descriptor? Fingerprint of derivation? 

Imagine you derive a key for a custom script, would be useful to bind those (how to derive the key, where its being used) etc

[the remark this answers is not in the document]
tiero: Otherwise yes the DeriveNextAddress approach we used in the past is the "deterministic API" where we assign script template incrementally derive a new contracts script (hence address) encoding with a fresh new key.

Both approach can co-exist 

Fells the simpler in the current situation as we don't getNextAddress/getNextChangeAddress/accounts is

- give out a pubkey (here's derivation) 
- register a contract (or default) and signal the pubkey derivation path 
- spend from the contract passing the derivation path

tiero:
Overall I think Wallet is not anymore the right terminology if it reduces to low level signing methods and key derivation (address encoding is a pure function stateless and not changing between implementations, so irrelevant) 


In TypeScript we call it Identity and allows to be injected into Wallets (which are instead stateful and aware of how spending from scripts and transactions).

pkg/client-lib/wallet/wallet.go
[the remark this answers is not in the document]
altafan: > I agree that we should at very least give out something or. Descriptor? Fingerprint of derivation?
> 
> Imagine you derive a key for a custom script, would be useful to bind those (how to derive the key, where its being used) etc

yeah was waiting for dusan's feedback on this. Not my intention to ignore this at all

altafan:
> Overall I think Wallet is not anymore the right terminology if it reduces to low level signing methods and key derivation (address encoding is a pure function stateless and not changing between implementations, so irrelevant)
> 
> In TypeScript we call it Identity and allows to be injected into Wallets (which are instead stateful and aware of how spending from scripts and transactions).

I can go for the renaming in this PR directly

pkg/client-lib/wallet/singlekey/bitcoin_wallet.go
```diff
@@ -44,6 +44,30 @@ func NewBitcoinWallet(
 	}, nil
 }
 
+func (w *bitcoinWallet) GetKeyPair(
```
bitcoin-coder-bob: this function and `NewKeyPair` are identical currently. Intentional? For a wallet with 1 key that makes sense.  But appears that `NewKeyPair` is not actually generating a key pair but returning the one from `w` already created

pkg/client-lib/wallet/wallet.go
```diff
@@ -18,17 +18,10 @@ type WalletService interface {
 	Lock(ctx context.Context) (err error)
 	Unlock(ctx context.Context, password string) (alreadyUnlocked bool, err error)
 	IsLocked() bool
-	GetAddresses(ctx context.Context) (
```
bitcoin-coder-bob, line 21: these were removed, but in `pkg/client-lib/wallet/singlekey/bitcoin_wallet.go` the functions for them are still there implementing these now deleted interface functions and are unused. Should we remove them from `pkg/client-lib/wallet/singlekey/bitcoin_wallet.go` as well?

pkg/client-lib/wallet/singlekey/bitcoin_wallet.go
```diff
@@ -168,6 +192,14 @@ func (w *bitcoinWallet) NewAddresses(
 func (s *bitcoinWallet) SignTransaction(
 	ctx context.Context, explorerSvc explorer.Explorer, tx string,
 ) (string, error) {
+	if s.walletData == nil {
```
bitcoin-coder-bob, line 78: should we also add these 2 checks into `sinTapscriptSpend` and `signTaprootKeySpend`

pkg/client-lib/wallet/wallet.go
```diff
@@ -18,17 +18,10 @@ type WalletService interface {
...
-		offchainAddresses, boardingAddresses, redemptionAddresses []types.Address, err error,
-	)
-	NewAddress(ctx context.Context, change bool) (
-		onchainAddr string, offchainAddr, boardingAddr *types.Address, err error,
-	)
-	NewAddresses(ctx context.Context, change bool, num int) (
-		onchainAddresses []string,
-		offchainAddresses, boardingAddresses []types.Address, err error,
-	)
+	GetKeyPair(
```
bitcoin-coder-bob: these 2 functions return a private key, but the only place we currently use them is in tests. I guess idea is we will eventually use it?

pkg/client-lib/funding.go
```diff
@@ -35,26 +39,12 @@ func (a *service) Receive(ctx context.Context) (
 
 func (a *service) GetAddresses(
 	ctx context.Context,
-) ([]string, []string, []string, []string, error) {
+) ([]string, []types.Address, []types.Address, []types.Address, error) {
```
bitcoin-coder-bob, line 43: is the reason we kept the first return value as a string since we dont use the `types.Address`'s `Tapscripts` field?

bitcoin-coder-bob: does it make sense to name these return values so it becomes easier to know which address is which

pkg/client-lib/wallet/singlekey/bitcoin_wallet.go
```diff
@@ -44,6 +44,30 @@ func NewBitcoinWallet(
 	}, nil
 }
 
+func (w *bitcoinWallet) GetKeyPair(
+	ctx context.Context, _ string,
```
bitcoin-coder-bob: why pass the unused param? future use?

pkg/client-lib/funding.go
```diff
@@ -307,6 +297,155 @@ func (a *service) NotifyIncomingFunds(ctx context.Context, addr string) ([]types
 	}
 }
 
+func (a *service) newAddress(ctx context.Context) (string, *types.Address, *types.Address, error) {
```
bitcoin-coder-bob: does it make sense to name these return values so it becomes easier to know which address is which

pkg/client-lib/funding.go
```diff
@@ -307,6 +297,155 @@ func (a *service) NotifyIncomingFunds(ctx context.Context, addr string) ([]types
...
+			Address:    encodedOffchainAddr,
+		}, &types.Address{
+			Tapscripts: boardingTapscripts,
+			Address:    boardingAddr.EncodeAddress(),
+		}, nil
+}
+
+func (a *service) getAddresses(
+	ctx context.Context,
+) ([]string, []types.Address, []types.Address, []types.Address, error) {
```
bitcoin-coder-bob: does it make sense to name these return values so it becomes easier to know which address is which

pkg/client-lib/wallet/singlekey/bitcoin_wallet.go
altafan: this is the single key impl of that interface. That's why these 2 methods match, as in a single key there are no new keypairs to derive. In a HD impl these methods would be different indeed

pkg/client-lib/wallet/wallet.go
altafan: Forgot to drop'em, will do it

pkg/client-lib/wallet/singlekey/bitcoin_wallet.go
altafan: i didn't touch the existing code but i'll check

pkg/client-lib/funding.go
altafan: the first is string because they are not taproot addresses. Eventually, we'll try to remove those addresses, they are needed only in case of unroll (we want to "move it out" from the client), to pay the network fees for the txs to broadcast (CPFP).

pkg/client-lib/wallet/singlekey/bitcoin_wallet.go
altafan: in HD this param would be likely a derivation path or index (will be changed to `uint32` in this case)

pkg/client-lib/wallet/wallet.go
altafan: Fixed in 1d33066

pkg/client-lib/funding.go
altafan: Fixed in 1d33066

pkg/client-lib/funding.go
altafan: Fixed in 1d33066

pkg/client-lib/funding.go
altafan: Fixed in 1d33066

pkg/client-lib/wallet/singlekey/bitcoin_wallet.go
```diff
@@ -44,130 +41,48 @@ func NewBitcoinWallet(
...
+	if w.IsLocked() {
+		return nil, fmt.Errorf("wallet is locked")
 	}
-
-	return onchainAddr.EncodeAddress(), &types.Address{
-		Tapscripts: offchainAddr.Tapscripts,
-		Address:    encodedOffchainAddr,
-	}, boardingAddr, nil
+	return &wallet.KeyRef{PubKey: w.privateKey.PubKey()}, nil
 }
```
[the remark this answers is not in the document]
altafan: the single-key wallet is gonna be "deprecated" soon in favor of an HD impl in go-sdk. This wallet is meant to be used in the future for testing purposes only

PR #1011, opened 2026-04-13
[client-lib] Add default pagination for GetVtxos and GetVirtualTxs

altafan:
Adds default pagination to client-lib's indexer client for GetVtxos and GetVirtualTxs apis. The request is paginated by default if more than 1000 items are requests in a single api.

Please @louisinger review

## Summary by CodeRabbit

* **New Features**
  * Automatic pagination for large data requests with aggregated results across multiple pages
  * Rate limiting for paginated requests and safe termination when pagination exceeds allowed limits

* **Tests**
  * Added tests for pagination behavior: multi-page aggregation, rate-limiting enforcement, error propagation, and cancellation handling

### DIFF
--- a/pkg/client-lib/indexer/grpc/client.go
+++ b/pkg/client-lib/indexer/grpc/client.go
@@ -708,7 +708,7 @@ func paginatedFetch[T any](
 		all = append(all, items...)
 		reqCount++
 
-		if page == nil || page.GetNext() == page.GetTotal() {
+		if page == nil || page.GetNext() >= page.GetTotal() {
 			break
 		}
 		if reqCount >= maxPages {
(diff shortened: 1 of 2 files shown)


pkg/client-lib/indexer/grpc/client.go
```diff
@@ -19,6 +19,12 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+const (
+	maxPageSize   = 1000
+	maxReqsPerSec = 20
```
louisinger: should be configurable ? what's happening if the rate limit change to 10 reqs/s ?

louisinger (approved):

pkg/client-lib/indexer/grpc/client.go
altafan: Lowered the const to `10 reqs/sec`, meaning we return 10k vtxos/txs in a sec before throttling

PR #1014, opened 2026-04-15
VerifyTapscriptSigs options

### LINKED ISSUE #1013: [ark-lib] VerifyTapscriptSigs should optionally fail if input doesn't have any signatures
VerifyTapscriptSigs is often used by external services to validate signatures where several signers are involved. However, it skips any input without tapscripts signatures : 
https://github.com/arkade-os/arkd/blob/b5217030543be25fe241e2b70ddaa1d49b05f124/pkg/ark-lib/script/verify.go#L148-L151

it makes sense in some flow, but it's required sometimes : https://github.com/ArkLabsHQ/introspector/pull/61#discussion_r3086085591

louisinger:
add `WithSkipPublicKeys` and `WithoutSkipUnsignedInput` options to VerifyTapscriptSigs. It also simplify `intent.Verify`.

it closes #1013 

@altafan please review

## Summary by CodeRabbit

* **Refactor**
  * Signature verification now uses an options-based configuration for more flexible control over which signers or unsigned inputs are skipped.
  * Verification routines reworked to rely on the new configuration model.

* **Bug Fixes**
  * Validation tightened for certain input types; unsigned inputs are treated as error by default unless explicitly skipped.

* **Tests**
  * Updated and expanded tests to cover the new options behavior and unsigned-input error cases.

### DIFF
--- a/pkg/ark-lib/script/verify.go
+++ b/pkg/ark-lib/script/verify.go
@@ -14,8 +14,6 @@ import (
 	"github.com/btcsuite/btcd/wire"
 )
 
-
-
 type VerifyTapscriptOption func(*verifyTapscriptOptions)
 
 type verifyTapscriptOptions struct {
@@ -41,10 +39,11 @@ func WithSkipPublicKeys(keys ...*btcec.PublicKey) VerifyTapscriptOption {
 	}
 }
 
-// WithoutSkipUnsignedInput makes the func returns an error if any input is unsigned
-func WithoutSkipUnsignedInput() VerifyTapscriptOption {
+// WithSkipUnsignedInputs makes the func ignore unsigned inputs instead of
+// returning an error.
+func WithSkipUnsignedInputs() VerifyTapscriptOption {
 	return func(o *verifyTapscriptOptions) {
-		o.skipUnsignedInputs = false
+		o.skipUnsignedInputs = true
 	}
 }
 
@@ -53,8 +52,7 @@ func VerifyTapscriptSigs(
 	tx *psbt.Packet, prevoutFetcher txscript.PrevOutputFetcher,
 	opts ...VerifyTapscriptOption,
 ) (signedInputs []int, err error) {
-	// skip unsigned input by default
-	options := &verifyTapscriptOptions{skipUnsignedInputs: true}
+	options := &verifyTapscriptOptions{}
 	for _, opt := range opts {
 		opt(options)
 	}
--- a/pkg/ark-lib/script/verify.go
+++ b/pkg/ark-lib/script/verify.go
@@ -12,12 +12,52 @@ import (
 	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
-	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
 
-// VerifyTapscriptSigs verifies the tapscript signatures of the given tx
-// it skips inputs that are not signed or do not specify a taproot leaf script
-func VerifyTapscriptSigs(tx *psbt.Packet, prevoutFetcher txscript.PrevOutputFetcher, skip []*secp256k1.PublicKey) (signedInputs []int, err error) {
+
+
+type VerifyTapscriptOption func(*verifyTapscriptOptions)
+
+type verifyTapscriptOptions struct {
+	skipPublicKeys     []string // x-only hex encoded
+	skipUnsignedInputs bool
+}
+
+// WithSkipPublicKeys sets the key to ignore in signature validation
+func WithSkipPublicKeys(keys ...*btcec.PublicKey) VerifyTapscriptOption {
+	return func(o *verifyTapscriptOptions) {
+		encoded := make([]string, 0, len(keys))
+		for _, pubkey := range keys {
+			xonly := hex.EncodeToString(schnorr.SerializePubKey(pubkey))
+
+			if slices.Contains(encoded, xonly) {
+				continue // avoid duplication
+			}
+
+			encoded = append(encoded, xonly)
+		}
+
+		o.skipPublicKeys = encoded
+	}
+}
+
+// WithoutSkipUnsignedInput makes the func returns an error if any input is unsigned
+func WithoutSkipUnsignedInput() VerifyTapscriptOption {
+	return func(o *verifyTapscriptOptions) {
+		o.skipUnsignedInputs = false
+	}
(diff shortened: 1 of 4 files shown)


pkg/ark-lib/script/verify.go
```diff
@@ -12,12 +12,52 @@ import (
...
+	}
+}
+
+// VerifyTapscriptSigs verifies the tapscript signatures of the given tx.
+func VerifyTapscriptSigs(
+	tx *psbt.Packet, prevoutFetcher txscript.PrevOutputFetcher,
+	opts ...VerifyTapscriptOption,
+) (signedInputs []int, err error) {
+	// skip unsigned input by default
+	options := &verifyTapscriptOptions{skipUnsignedInputs: true}
```
chris-ricketts, line 57: I would have expected the default for `skipUnsignedInputs` to be false and for the option to be something like `WithSkipUnsignedInputs` (sets `skipUnsignedInputs` to true) - that makes it kind of "fail safe" if the user forgets to set the option, i.e. options only loosen constraints.

altafan: we wanted to preserve the original behavior, ie. skipping unsigned inputs because this is how the functions is expected to work in arkd's core logic

chris-ricketts: I see but we are already changing 2/2 call sites of this function in the `arkd` repo in this PR anyway:

https://github.com/louisinger/ark/blob/9ff5acc40972f2e0197e3ea4c8b0ceece6a2a319/internal/infrastructure/tx-builder/covenantless/builder.go#L1064-L1068

https://github.com/louisinger/ark/blob/9ff5acc40972f2e0197e3ea4c8b0ceece6a2a319/pkg/ark-lib/intent/proof.go#L103-L107

I feel like the API is less confusing if you don't mix `With` and `Without` in the options.

altafan: @louisinger i changed my mind, let's follow chris' suggestion

ghost (changes requested):
## Arkana Code Review — Protocol-Critical

This PR refactors `VerifyTapscriptSigs` from positional `skip` parameter to a functional options pattern (`WithSkipPublicKeys`, `WithSkipUnsignedInputs`). The default behavior is now stricter (errors on unsigned inputs unless opted in), which is a good safety improvement. However, I have concerns about the `proof.go` changes.

⚠️ **This touches signature verification on intent proofs — protocol-critical code. Requires human review.**

---

### 🔴 P0 — Signature verification gap in `proof.go` for non-standard inputs

**File:** `pkg/ark-lib/intent/proof.go:103-127` (new code)

The new `proof.go` post-loop only validates note closure inputs. Non-note inputs are assumed to be fully validated by `VerifyTapscriptSigs`. But `VerifyTapscriptSigs` silently skips two classes of inputs:

1. **Non-taproot inputs** (`verify.go:48` — `GetScriptClass != WitnessV1TaprootTy → continue`)
2. **Inputs without TaprootLeafScript** (`verify.go:36` — `len(TaprootLeafScript) != 1 → continue`)

Case 2 is now caught by the new `len(in.TaprootLeafScript) != 1` error in `proof.go:106` — good.

But Case 1 is a gap: if a malicious intent proof includes a non-taproot input at index 1+ with exactly 1 `TaprootLeafScript` entry containing a non-note script, this is what happens:
- `VerifyTapscriptSigs`: skips it (not taproot) — no error
- New `proof.go` loop: passes `len == 1` check, `!IsNoteClosureScript` → `continue` — no verification

**The old code caught this** — any input not in `signedSet` and not matching a note closure or `allSignersSkipped` would error with `"missing signature for input %d"`.

In practice this is likely unexploitable because the server validates outpoints against its VTXO database, but defense-in-depth matters. Suggested fix: either (a) add a non-taproot prevout check in the `proof.go` loop, or (b) make `VerifyTapscriptSigs` return which inputs it actually examined so the caller can detect unexamined inputs.

---

### 🟡 P1 — Behavioral widening: unsigned non-condition closures now silently accepted

**File:** `pkg/ark-lib/script/verify.go:186-201`

The old `allSignersSkipped` in `proof.go:461-490` only accepted unsigned inputs for `ConditionMultisigClosure` and `ConditionCSVMultisigClosure`. The new logic in `VerifyTapscriptSigs` checks `expectedSigners` against `skipPublicKeys` for **all closure types** (including `MultisigClosure`, `CSVMultisigClosure`, `CLTVMultisigClosure`).

This means an unsigned `MultisigClosure` input where all signers are in the skip list is now silently accepted — previously it would error. Since callers pass only the ASP key as `skip`, this would only matter for VTXOs with ONLY the ASP key (degenerate case), so likely not exploitable. But the behavioral change should be intentional and documented, not accidental.

**Question for the author:** Was this widening intentional?

---

### 🟡 P1 — Missing test coverage for critical path

**File:** `pkg/ark-lib/script/verify_test.go`

The new "all signers in skip list, unsigned input, no error" path in `verify.go:186-201` has no dedicated test. This is the path that replaced the old `allSignersSkipped` function — it's protocol-critical and needs explicit test coverage. Suggest adding:

```go
t.Run("unsigned input with all signers skipped is accepted", func(t *testing.T) {
    setup := newSingleKeySetup(t)
    packet, prevoutFetcher := buildTx(t, setup)
    // No signatures added.

    signedInputs, err := script.VerifyTapscriptSigs(
        packet, prevoutFetcher,
        script.WithSkipPublicKeys(setup.privKey.PubKey()),
    )
    require.NoError(t, err)
    require.Empty(t, signedInputs) // should NOT be in signedInputs
})
```

---

### 🟢 Minor — `WithSkipPublicKeys` overwrites rather than appends

**File:** `pkg/ark-lib/script/verify.go:27-40`

Multiple calls to `WithSkipPublicKeys` silently drop the first set. Current usage only calls it once, so not a bug today, but the API is a footgun. Consider either making it additive or documenting the overwrite behavior. (CodeRabbit also flagged this.)

---

### ✅ What looks good

- Default-strict behavior (error on unsigned) is the right call — safer by construction
- API cleanup removing `secp256k1` in favor of `btcec` is clean
- `builder.go` call site correctly opts into `WithSkipUnsignedInputs` to maintain existing behavior
- New test case `"unsigned input errors by default"` validates the default-strict path
- The `continue` after the all-skipped check in `verify.go:201` correctly avoids appending to `signedInputs` (contrary to CodeRabbit's incorrect claim)

---

**Requesting changes** for the P0 defense-in-depth gap and the missing test. The behavioral widening (P1) needs an explicit answer on intent.

PR #1022, opened 2026-04-19
Copy extension packets to batch leaf transaction

### LINKED ISSUE #1017: Intent extension packets should be set on batch leaf transaction
In RegisterIntent, only the asset packet is set to the batch leaf of the vtxo tree :
https://github.com/arkade-os/arkd/blob/65edd166071448b58602b132016e7e194dee6321/internal/core/application/service.go#L1896-L1909

however, users may also want to set other extension packets to the batch leaf. For instance, to [carry contract state accross settlements](https://github.com/ArkLabsHQ/introspector/pull/66). 

The extension output sets to intent transaction should be moved -entirely- to the batch leaf. The logic transforming the asset packet (mark input type 'intent') should be unchanged.

louisinger:
it closes #1017 

@altafan @michael1011 please review

## Release Notes

* **Refactor**
  * Improved transaction extension validation with enhanced early-detection mechanisms.
  * Restructured how extension data is serialized and processed during transaction creation.
  * Simplified extension handling in the transaction building pipeline.

* **Tests**
  * Updated test suite to align with restructured extension handling.

### DIFF
--- a/internal/core/application/asset_validation.go
+++ b/internal/core/application/asset_validation.go
@@ -63,10 +63,6 @@ func (s assetSource) AssetExists(ctx context.Context, assetID string) bool {
 }
 
 func hasIssuance(packet asset.Packet) bool {
-	if len(packet) == 0 {
-		return false
-	}
-
 	for _, group := range packet {
 		if group.IsIssuance() {
 			return true
--- a/internal/core/application/asset_validation.go
+++ b/internal/core/application/asset_validation.go
@@ -63,6 +63,10 @@ func (s assetSource) AssetExists(ctx context.Context, assetID string) bool {
 }
 
 func hasIssuance(packet asset.Packet) bool {
+	if len(packet) == 0 {
+		return false
+	}
+
 	for _, group := range packet {
 		if group.IsIssuance() {
 			return true
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -16,6 +16,7 @@ import (
 	"github.com/arkade-os/arkd/internal/core/domain"
 	"github.com/arkade-os/arkd/internal/core/ports"
 	arklib "github.com/arkade-os/arkd/pkg/ark-lib"
+	"github.com/arkade-os/arkd/pkg/ark-lib/asset"
 	"github.com/arkade-os/arkd/pkg/ark-lib/extension"
 	"github.com/arkade-os/arkd/pkg/ark-lib/intent"
 	"github.com/arkade-os/arkd/pkg/ark-lib/offchain"
@@ -1893,9 +1894,9 @@ func (s *service) RegisterIntent(
 		return "", err
 	}
 
-	leafTxPacket := ""
-	intentAssetPacket := ext.GetAssetPacket()
-	if len(intentAssetPacket) > 0 {
+	leafTxExtension := ""
+	if len(ext) > 0 {
+		intentAssetPacket := ext.GetAssetPacket()
 		// disable issuance in settlement
 		if hasIssuance(intentAssetPacket) {
 			return "", errors.INVALID_INTENT_PROOF.New("intent contains asset issuance").
@@ -1905,11 +1906,31 @@ func (s *service) RegisterIntent(
 				})
 		}
 
-		leafTxPacket = intentAssetPacket.LeafTxPacket(proof.UnsignedTx.TxHash()).String()
+		// rebuild the batch leaf extension from the intent extension packets
+		// copy all intent packets to batch leaf except the asset packet : transform it as input type "intent"
+		leafExtension := make(extension.Extension, 0, len(ext))
+		for _, pkt := range ext {
+			if ap, ok := pkt.(asset.Packet); ok {
+				leafAssetPacket := ap.LeafTxPacket(proof.UnsignedTx.TxHash())
+				leafExtension = append(leafExtension, leafAssetPacket)
+				continue
+			}
+			leafExtension = append(leafExtension, pkt)
+		}
+
+		leafExtScript, err := leafExtension.Serialize()
+		if err != nil {
+			return "", errors.INTERNAL_ERROR.New("failed to serialize leaf extension: %w", err).
+				WithMetadata(map[string]any{
+					"proof":   encodedProof,
+					"message": encodedMessage,
+				})
+		}
+		leafTxExtension = hex.EncodeToString(leafExtScript)
 	}
 
 	intent, err := domain.NewIntent(
-		proofTxid, encodedProof, encodedMessage, vtxoInputs, leafTxPacket,
+		proofTxid, encodedProof, encodedMessage, vtxoInputs, leafTxExtension,
 	)
 	if err != nil {
 		return "", errors.INTERNAL_ERROR.New("failed to create intent: %w", err).
--- a/internal/core/domain/intent.go
+++ b/internal/core/domain/intent.go
@@ -13,22 +13,22 @@ type Intent struct {
 	Proof     string
 	Message   string
 	Txid      string
-	// asset packet that should be added to vtxo tree leaf transaction (hex format)
-	LeafTxAssetPacket string
+	// hex-encoded extension OP_RETURN script to be added as output of the vtxo tree leaf transaction
+	LeafTxExtension string
 }
 
 func NewIntent(
 	proofTxid, proof, message string,
 	inputs []Vtxo,
-	leafTxPacket string,
+	leafTxExtension string,
 ) (*Intent, error) {
 	intent := &Intent{
-		Id:                uuid.New().String(),
-		Inputs:            inputs,
-		Proof:             proof,
-		Message:           message,
-		Txid:              proofTxid,
-		LeafTxAssetPacket: leafTxPacket,
+		Id:              uuid.New().String(),
+		Inputs:          inputs,
+		Proof:           proof,
+		Message:         message,
+		Txid:            proofTxid,
+		LeafTxExtension: leafTxExtension,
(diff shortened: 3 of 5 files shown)


ghost (changes requested):
## 🔴 Protocol-Critical Review — Requires Human Sign-off

This PR changes what data is embedded in VTXO tree leaf transactions. Any change to leaf outputs changes the tree merkle root, affecting commitment transactions, forfeit paths, and unilateral exit paths. **This is protocol-critical and must not merge without human review.**

---

### Architecture Summary

**Before:** Only the asset packet was extracted from the intent extension, converted to leaf form (`LeafTxPacket`), stored as a string, then re-parsed and wrapped in `Extension` at tree-build time to produce the OP_RETURN output.

**After:** ALL extension packets are iterated at registration time. Asset packets are transformed via `LeafTxPacket()`, non-asset packets are copied as-is. The full extension is serialized to an OP_RETURN script and stored as hex. At tree-build time, the hex is used directly as the leaf output script.

The refactor is clean and the final output format is equivalent for the asset-only case. The issues below concern the new non-asset packet copying behavior and missing test coverage.

---

### Findings

#### 1. 🔴 Blind copy of unknown packets to leaf transactions — `service.go:1916-1921`

```go
leafExtension = append(leafExtension, pkt)
```

Non-asset packets (currently `UnknownPacket` — the only other type implementing `extension.Packet`) are copied verbatim into the leaf transaction extension. This is dangerous:

- **No validation** that a given packet type is meaningful/safe in a leaf tx context. The intent tx and the leaf tx are different transactions with different structures.
- **Future packet types** added to the extension system will be silently copied to leaf transactions with no opt-in. This is an open extension point that defaults to "include everything" — the safe default should be "include nothing unless explicitly handled."
- Consider an allowlist approach: explicitly handle known packet types (asset → transform, type X → copy, unknown → skip or error).

#### 2. 🟡 Condition change widens scope — `service.go:1897`

**Old:** `if len(intentAssetPacket) > 0` — only runs when an asset packet exists.
**New:** `if len(ext) > 0` — runs when *any* extension packet exists.

An extension containing only non-asset packets (e.g., a single `UnknownPacket`) will now generate a leaf extension output where previously there would be none. This adds an OP_RETURN output to the leaf transaction that wasn't there before, changing the tree structure for these intents. Is this intentional?

If an extension has non-asset packets but no asset packet, the `hasIssuance` check runs on a nil/empty `Packet` — this is safe due to the new guard in `asset_validation.go:66-68`, but the semantic intent is unclear: why check for issuance when there's no asset packet?

#### 3. 🟡 Insufficient test coverage for the core change

The actual behavior change — the loop that copies/transforms extension packets — has **zero test coverage**. The test changes in `intent_test.go` are just renames. Missing tests:

- Extension with asset packet + unknown packet → verify both appear in serialized leaf extension
- Extension with only unknown packets → verify behavior (should it produce a leaf extension?)
- Extension with empty asset packet (0 groups) → verify `Serialize()` handles the empty packet bytes from `LeafTxPacket`
- Round-trip: verify the hex-encoded script in `LeafTxExtension` can be parsed back by `extension.NewExtensionFromBytes`

Also agree with CodeRabbit: the test should assert `intent.LeafTxExtension == leafTxExtension`.

#### 4. ✅ `hasIssuance` nil guard — `asset_validation.go:66-68`

The added `if len(packet) == 0 { return false }` guard is correct and necessary given the widened condition in service.go. Good.

#### 5. ✅ `Amount: 0` hardcoding — `utils.go:83`

Equivalent to old behavior (`TxOut()` returned `wire.NewTxOut(0, script)`). No issue.

#### 6. ✅ Domain rename — `intent.go`

`LeafTxAssetPacket` → `LeafTxExtension` is correct. Field is in-memory only (not persisted to DB), so no migration needed. No other consumers exist outside the changed files.

---

### Verdict

**Request changes.** The blind-copy of unknown packets (#1) is a protocol-level concern — it creates an open extension point where new packet types automatically end up in VTXO tree leaves without explicit handling. At minimum, add an allowlist or skip unknown types. The missing test coverage (#3) for the core loop logic needs to be addressed before merge.

Even after fixes, this requires human sign-off as it is protocol-critical (changes VTXO tree leaf structure).

🤖 Review by Arkana

ghost (changes requested):
## 🔴 Protocol-Critical Review — Requires Human Sign-off

This PR changes what data is embedded in VTXO tree leaf transactions. Any change to leaf outputs changes the tree merkle root, affecting commitment transactions, forfeit paths, and unilateral exit paths. **This is protocol-critical and must not merge without human review.**

---

### Architecture Summary

**Before:** Only the asset packet was extracted from the intent extension, converted to leaf form (`LeafTxPacket`), stored as a string, then re-parsed and wrapped in `Extension` at tree-build time to produce the OP_RETURN output.

**After:** ALL extension packets are iterated at registration time. Asset packets are transformed via `LeafTxPacket()`, non-asset packets are copied as-is. The full extension is serialized to an OP_RETURN script and stored as hex. At tree-build time, the hex is used directly as the leaf output script.

The refactor is clean and the final output format is equivalent for the asset-only case. The issues below concern the new non-asset packet copying behavior and missing test coverage.

---

### Findings

#### 1. 🔴 Blind copy of unknown packets to leaf transactions — `service.go:1916-1921`

```go
leafExtension = append(leafExtension, pkt)
```

Non-asset packets (currently `UnknownPacket` — the only other type implementing `extension.Packet`) are copied verbatim into the leaf transaction extension. This is dangerous:

- **No validation** that a given packet type is meaningful/safe in a leaf tx context. The intent tx and the leaf tx are different transactions with different structures.
- **Future packet types** added to the extension system will be silently copied to leaf transactions with no opt-in. This is an open extension point that defaults to "include everything" — the safe default should be "include nothing unless explicitly handled."
- Consider an allowlist approach: explicitly handle known packet types (asset → transform, type X → copy, unknown → skip or error).

#### 2. 🟡 Condition change widens scope — `service.go:1897`

**Old:** `if len(intentAssetPacket) > 0` — only runs when an asset packet exists.
**New:** `if len(ext) > 0` — runs when *any* extension packet exists.

An extension containing only non-asset packets (e.g., a single `UnknownPacket`) will now generate a leaf extension output where previously there would be none. This adds an OP_RETURN output to the leaf transaction that wasn't there before, changing the tree structure for these intents. Is this intentional?

If an extension has non-asset packets but no asset packet, the `hasIssuance` check runs on a nil/empty `Packet` — this is safe due to the new guard in `asset_validation.go:66-68`, but the semantic intent is unclear: why check for issuance when there's no asset packet?

#### 3. 🟡 Insufficient test coverage for the core change

The actual behavior change — the loop that copies/transforms extension packets — has **zero test coverage**. The test changes in `intent_test.go` are just renames. Missing tests:

- Extension with asset packet + unknown packet → verify both appear in serialized leaf extension
- Extension with only unknown packets → verify behavior (should it produce a leaf extension?)
- Extension with empty asset packet (0 groups) → verify `Serialize()` handles the empty packet bytes from `LeafTxPacket`
- Round-trip: verify the hex-encoded script in `LeafTxExtension` can be parsed back by `extension.NewExtensionFromBytes`

Also agree with CodeRabbit: the test should assert `intent.LeafTxExtension == leafTxExtension`.

#### 4. ✅ `hasIssuance` nil guard — `asset_validation.go:66-68`

The added `if len(packet) == 0 { return false }` guard is correct and necessary given the widened condition in service.go. Good.

#### 5. ✅ `Amount: 0` hardcoding — `utils.go:83`

Equivalent to old behavior (`TxOut()` returned `wire.NewTxOut(0, script)`). No issue.

#### 6. ✅ Domain rename — `intent.go`

`LeafTxAssetPacket` → `LeafTxExtension` is correct. Field is in-memory only (not persisted to DB), so no migration needed. No other consumers exist outside the changed files.

---

### Verdict

**Request changes.** The blind-copy of unknown packets (#1) is a protocol-level concern — it creates an open extension point where new packet types automatically end up in VTXO tree leaves without explicit handling. At minimum, add an allowlist or skip unknown types. The missing test coverage (#3) for the core loop logic needs to be addressed before merge.

Even after fixes, this requires human sign-off as it is protocol-critical (changes VTXO tree leaf structure).

🤖 Review by Arkana

internal/core/application/asset_validation.go
```diff
@@ -63,6 +63,10 @@ func (s assetSource) AssetExists(ctx context.Context, assetID string) bool {
 }
 
 func hasIssuance(packet asset.Packet) bool {
+	if len(packet) == 0 {
+		return false
+	}
```
altafan: not really needed?

altafan (approved):

ghost (commented):
## 🤖 Follow-up Review (new commit: `906f440`)

### New commit: ✅ `hasIssuance` guard removal — `asset_validation.go:66`

Removing the `if len(packet) == 0 { return false }` guard is correct. In Go, `for range nil` is a no-op that returns `false` naturally. Clean.

---

### Previous findings — still open

My prior concerns from the initial review remain unaddressed:

1. **🔴 Blind copy of unknown packets** (`service.go:1916-1921`) — non-asset packets are still copied verbatim into leaf transactions with no allowlist. New packet types will silently end up in VTXO tree leaves.
2. **🟡 Test coverage** — the core loop (iterate/transform/copy packets → serialize) has no test coverage. The test changes are still just renames.

@altafan has approved, which satisfies the human sign-off requirement for protocol-critical changes. Deferring to his judgment on #1 (he may have context on whether the "copy everything" design is intentional for upcoming packet types). But flagging that it remains an open extension point.

🤖 Review by Arkana

Issue #1033, opened 2026-04-22
Panic: nil pointer dereference in scheduleSweepBatchOutput during DB blip


ghost:
## Description

`arkd` crashed with a nil pointer dereference during a brief database connectivity interruption (RDS maintenance on nbxplorer DB).

## Panic trace

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x9352de]

goroutine 3973311 [running]:
github.com/arkade-os/arkd/internal/core/application.(*service).scheduleSweepBatchOutput(0x182fad9fa788, 0x182fb016d600)
        /app/internal/core/application/service.go:3480 +0x17e
created by github.com/arkade-os/arkd/internal/core/application.(*service).registerEvent
```

## Context

- arkd v0.9.3 deployed on production
- RDS maintenance was performed on the nbxplorer database, causing a brief connectivity loss
- `scheduleSweepBatchOutput` is a background goroutine that likely depends on nbxplorer — the nil dereference likely comes from an unhandled error response when nbxplorer was temporarily unreachable
- Only `arkd` restarted (14 min uptime vs other containers still up 25h+)

## Expected behavior

`arkd` should handle transient DB/nbxplorer connectivity issues gracefully without panicking. The sweep batch output scheduling should either retry or log an error and skip.

## Reported by

Discovered during production RDS maintenance on 2026-04-22. Louis is working on a fix.

PR #1034, opened 2026-04-22
fix: waitForConfirmation fallback to time.Now() in case of error

louisinger:
in case the sweeper waitForConfirmation fails, fallback to the current timestamp. it means the scheduler may schedule too early. it is not a problem because that situation is already handled by "BIP68 not final" retry loop. better to schedule the sweep earlier instead of not scheduling it at all like proposed in #1033 

@altafan please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a potential crash in batch sweep task scheduling when confirmation handling failed, ensuring proper expiration timestamp initialization.

### DIFF
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -3471,6 +3471,7 @@ func (s *service) scheduleSweepBatchOutput(round *domain.Round) {
 			"failed to wait for confirmation of commitment tx %s, schedule task time may be inaccurate",
 			round.CommitmentTxid,
 		)
+		blockTimestamp = &ports.BlockTimestamp{Time: time.Now().Unix()}
 	}
 
 	var expirationTimestamp int64
--- a/internal/core/application/sweeper.go
+++ b/internal/core/application/sweeper.go
@@ -474,6 +474,7 @@ func (s *sweeper) createBatchSweepTask(commitmentTxid, vtxoTreeRootTxid string) 
 					"failed to wait for confirmation of batch input tx %s, schedule task time "+
 						"may be inaccurate", rootInput,
 				)
+				blockTimestamp = &ports.BlockTimestamp{Time: time.Now().Unix()}
 			}
 
 			var expirationTimestamp int64


ghost (changes requested):
## Arkana Code Review — arkade-os/arkd#1034

**Verdict: Request changes (minor) + flag for human review (protocol-critical)**

### Context

Fixes a **production nil-pointer panic** (#1033) — `scheduleSweepBatchOutput` and `sweeper.createBatchSweepTask` dereference `blockTimestamp` after `waitForConfirmation` returns `(nil, err)`. The panic was triggered during an RDS maintenance window when nbxplorer was briefly unreachable. Good catch, correct root cause.

### What's correct

- The fix prevents the nil-pointer dereference in both call sites ✅
- The approach of scheduling "too early" and relying on the BIP68 retry loop is sound ✅
- Better to schedule than to silently skip the sweep ✅

### Issue: `Height` field is zero — use existing pattern from `fraud.go`

**`service.go:3474` and `sweeper.go:477`** — The fallback sets `Time: time.Now().Unix()` but leaves `Height: 0`. If `scheduler.Unit() == ports.BlockHeight` (line 3477 / line 480), the expiration computes as `0 + batchExpiry.Value` — a block height from 2009. The sweep fires immediately, BIP68 rejects it, and the retry loop burns cycles until the real expiry arrives. This works but is unnecessarily wasteful.

**`fraud.go:56` already handles this correctly** — it calls `s.wallet.GetCurrentBlockTime(ctx)` as an intermediate fallback, which returns both `Height` and `Time`. Only if *that* also fails does it bail out.

**Suggested fix:** Mirror the `fraud.go` pattern — try `GetCurrentBlockTime` first, fall back to `time.Now()` only if the wallet is truly unreachable:

```go
// service.go:3474 (and same pattern in sweeper.go:477)
blockTimestamp, err = s.wallet.GetCurrentBlockTime(ctx)
if err != nil {
    log.WithError(err).Warn("GetCurrentBlockTime also failed, falling back to time.Now()")
    blockTimestamp = &ports.BlockTimestamp{Time: time.Now().Unix()}
}
```

In `sweeper.go`, the wallet is accessible as `s.wallet` — same access pattern.

This gives the BlockHeight scheduler a realistic height and only degrades to wall-clock time when the wallet is fully down.

### No test coverage

Neither error path has unit tests. The existing code is also untested for this branch, so this PR doesn't make it worse — but a test that mocks `waitForConfirmation` returning an error and verifies the sweep is still scheduled would prevent future regressions.

### No cross-repo impact

No public API, proto, or type changes. SDK consumers are unaffected.

---

🔒 **Protocol-critical flag**: This touches sweep scheduling (VTXO lifecycle / round lifecycle). The fix is straightforward and the risk is low, but **human sign-off required** per protocol-critical review policy.

🤖 Reviewed by Arkana

altafan (approved):

PR #1038, opened 2026-04-22
Wait for projection db updates before sending events through streams

altafan:
In 90ee4eb287ba6b364d7a24ea90d097474d71c929 we had to introduce a query to the db in `updateProkectionsAfterOffchainTxEvents` to make sure we update vtxos records with the correct swept status, and we now see this latency is getting evident when clients get notified and make requests to the indexer right after, as they may ask for data that is not yet in db.

This PR mitigates the problem by making indexer and arkd send events and notify clients only AFTER the projection db has been updated, and not while the update is ongoing.

The other mitigation will be done in another pr and is related to improving the performance of the subject query (return only the sweep txs instead of the whole batch data - check #1040).

Please @louisinger review.

## Summary by CodeRabbit

* **Refactor**
  * Reworked internal event update flow to use type-specific, concurrency-safe update handlers and conditional dispatch for batch and offchain updates.

* **Tests**
  * Added tests covering handler registration and dispatch behavior to ensure single-install semantics and correct invocation.

### DIFF
--- a/internal/infrastructure/db/service.go
+++ b/internal/infrastructure/db/service.go
@@ -546,7 +546,9 @@ func (s *service) updateProjectionsAfterOffchainTxEvents(events []domain.Event) 
 				return false
 			}
 
-			batch, batchErr := s.roundStore.GetRoundWithCommitmentTxid(ctx, offchainTx.RootCommitmentTxId)
+			batch, batchErr := s.roundStore.GetRoundWithCommitmentTxid(
+				ctx, offchainTx.RootCommitmentTxId,
+			)
 			// We consider the tx swept if:
 			// - there is an error fetching the batch (this is just fallback, should never happen)
 			// - the batch is swept
--- a/internal/infrastructure/db/service.go
+++ b/internal/infrastructure/db/service.go
@@ -93,15 +93,17 @@ type ServiceConfig struct {
 }
 
 type service struct {
-	eventStore            domain.EventRepository
-	roundStore            domain.RoundRepository
-	vtxoStore             domain.VtxoRepository
-	scheduledSessionStore domain.ScheduledSessionRepo
-	offchainTxStore       domain.OffchainTxRepository
-	convictionStore       domain.ConvictionRepository
-	assetStore            domain.AssetRepository
-	intentFeesStore       domain.FeeRepository
-	txDecoder             ports.TxDecoder
+	eventStore              domain.EventRepository
+	roundStore              domain.RoundRepository
+	vtxoStore               domain.VtxoRepository
+	scheduledSessionStore   domain.ScheduledSessionRepo
+	offchainTxStore         domain.OffchainTxRepository
+	convictionStore         domain.ConvictionRepository
+	assetStore              domain.AssetRepository
+	intentFeesStore         domain.FeeRepository
+	txDecoder               ports.TxDecoder
+	batchUpdateHandler      *updateHandler[domain.Round]
+	offchainTxUpdateHandler *updateHandler[domain.OffchainTx]
 }
 
 func NewService(config ServiceConfig, txDecoder ports.TxDecoder) (ports.RepoManager, error) {
@@ -360,15 +362,17 @@ func NewService(config ServiceConfig, txDecoder ports.TxDecoder) (ports.RepoMana
 	}
 
 	svc := &service{
-		eventStore:            eventStore,
-		roundStore:            roundStore,
-		vtxoStore:             vtxoStore,
-		scheduledSessionStore: scheduledSessionStore,
-		offchainTxStore:       offchainTxStore,
-		txDecoder:             txDecoder,
-		convictionStore:       convictionStore,
-		assetStore:            assetStore,
-		intentFeesStore:       intentFeesStore,
+		eventStore:              eventStore,
+		roundStore:              roundStore,
+		vtxoStore:               vtxoStore,
+		scheduledSessionStore:   scheduledSessionStore,
+		offchainTxStore:         offchainTxStore,
+		txDecoder:               txDecoder,
+		convictionStore:         convictionStore,
+		assetStore:              assetStore,
+		intentFeesStore:         intentFeesStore,
+		batchUpdateHandler:      newUpdateHandler[domain.Round](),
+		offchainTxUpdateHandler: newUpdateHandler[domain.OffchainTx](),
 	}
 
 	// Register handlers that take care of keeping the projection store up-to-date.
@@ -414,6 +418,14 @@ func (s *service) Fees() domain.FeeRepository {
 	return s.intentFeesStore
 }
 
+func (s *service) RegisterBatchUpdateHandler(handler func(data domain.Round)) {
+	s.batchUpdateHandler.set(handler)
+}
+
+func (s *service) RegisterOffchainTxUpdateHandler(handler func(data domain.OffchainTx)) {
+	s.offchainTxUpdateHandler.set(handler)
+}
+
 func (s *service) Close() {
 	s.eventStore.Close()
 	s.roundStore.Close()
@@ -427,7 +439,14 @@ func (s *service) updateProjectionsAfterRoundEvents(events []domain.Event) {
 	ctx := context.Background()
 	round := domain.NewRoundFromEvents(events)
 
-	if err := s.roundStore.AddOrUpdateRound(ctx, *round); err != nil {
+	var err error
+	defer func() {
+		if err == nil {
+			s.batchUpdateHandler.dispatch(*round)
+		}
+	}()
+
+	if err = s.roundStore.AddOrUpdateRound(ctx, *round); err != nil {
 		log.WithError(err).Errorf("failed to add or update round %s", round.Id)
 		return
 	}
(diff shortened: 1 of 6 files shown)


ghost (changes requested):
## 🔍 Aggressive Code Review — arkd#1038

**Concept**: Sound. Moving client notifications to fire AFTER projection DB writes eliminates the race where clients query stale data. The `updateHandler` generic is clean. But there's a **protocol-critical bug** in the error-handling interaction with the `defer` dispatch pattern.

---

### 🚨 HIGH — BatchSwept dispatch silently dropped on SweepVtxos failure

**File:** `internal/infrastructure/db/service.go` — the new `updateProjectionsAfterRoundEvents` logic

The PR changes all `:=` error declarations to `=` against a shared named `err` variable, with a `defer` that only dispatches when `err == nil`:

```go
var err error
defer func() {
    if err == nil {
        s.batchUpdateHandler.dispatch(*round)
    }
}()
```

**The problem is the BatchSwept path (around line 462–471 in the new code):**

```go
if lastEvent.GetType() == domain.EventTypeBatchSwept {
    sweptCount, err = repo.SweepVtxos(ctx, allSweptVtxos)  // err = non-nil on failure
    if err != nil {
        log.WithError(err).Warn("failed to sweep vtxos")
        // NO RETURN HERE — falls through to FullySwept check
    }
    ...
    return  // defer fires with err != nil → dispatch SKIPPED
}
```

When `SweepVtxos` fails:
1. `err` is set to non-nil
2. The function hits `return` at the end of the BatchSwept block
3. The `defer` sees `err != nil` → **dispatch is skipped**
4. The application handler never fires

**What gets lost:**
- `propagateEvents()` — clients never learn the batch was swept
- `propagateTransactionEvent()` — no sweep tx event emitted
- Downstream indexers/SDKs see stale unswept state

**In the old code**, both handlers (DB projection + application notification) fired independently in parallel from `RegisterEventsHandler`. A `SweepVtxos` failure in the DB handler had **zero impact** on the application handler. This PR couples them, creating a regression.

**This is protocol-critical**: swept-batch notifications drive client state machines. Silent drops here mean clients hold stale VTXO state and may attempt to spend already-swept VTXOs.

**Fix**: Either (a) scope `err` tightly around `SweepVtxos` so it doesn't leak into the defer, or (b) dispatch unconditionally after `AddOrUpdateRound` succeeds (the round projection is the critical write — sweep/settle/add are secondary projections), or (c) reset `err = nil` after the SweepVtxos warning log since the sweep failure is already treated as non-fatal.

---

### ⚠️ MEDIUM — Shared `err` is fragile across the offchain tx handler

**File:** `internal/infrastructure/db/service.go` — `updateProjectionsAfterOffchainTxEvents`

Same pattern issue. In the `IsAccepted()` case, the checkpoint tx decode loop uses `err =`:

```go
for _, tx := range offchainTx.CheckpointTxs {
    txid, ins, _, err = s.txDecoder.DecodeTx(tx)
    if err != nil {
        continue  // err left non-nil
    }
}
```

If the **last** checkpoint tx fails to decode, `err` is non-nil entering `SpendVtxos`. If `SpendVtxos` succeeds, it clears `err` and dispatch happens — so this works **by accident**. A future refactor that reorders operations or adds an early return could silently break dispatch.

In the `IsFinalized()` case, `GetRoundWithCommitmentTxid` intentionally uses error as a signal (`txSwept = err != nil || ...`) but doesn't clear `err` afterward. Dispatch only works because `AddVtxos` is always called last and clears `err` on success.

**Recommendation**: Consider scoping intermediate errors locally (with `:=`) and only assigning to the shared `err` for operations that should gate dispatch. This makes the dispatch contract explicit rather than implicit.

---

### ⚠️ MEDIUM — No integration test for the dispatch-after-projection guarantee

**File:** `internal/infrastructure/db/update_handler_test.go`

The unit tests cover `updateHandler` in isolation (set, dispatch, write-once semantics) — good. But there's **no test** verifying the actual guarantee this PR exists to provide: that the application handler fires AFTER projection writes complete. An integration test that:
1. Saves round events
2. Asserts the projection DB is updated
3. Asserts the handler received the round data

...would catch regressions like the BatchSwept bug above.

---

### 📝 LOW — Nit: `set()` silently ignores duplicate registration

**File:** `internal/infrastructure/db/update_handler.go:21-24`

```go
func (u *updateHandler[T]) set(handler func(data T)) {
    if u.handler != nil {
        return  // silent no-op
    }
```

Write-once is fine, but a silent no-op is a debugging footgun. If someone accidentally calls `RegisterBatchUpdateHandler` twice (e.g., in a test or refactor), the second handler is silently dropped with no log or error. Consider at minimum a `log.Warn` or returning an error.

---

### ✅ What looks good

- The `updateHandler` generic is clean and well-typed
- Pointer→value receiver changes (`*domain.Round` → `domain.Round`) are correct — these are read-only consumers
- The `go u.handler(data)` in dispatch correctly avoids blocking the projection write path
- Test coverage for the handler primitive itself is solid
- The `decodeTx(offchainTx)` signature change (removing the dereference) is consistent

---

### Summary

The BatchSwept dispatch bug is a **must-fix** before merge. The shared-err fragility is a strong recommendation. The concept is right — please fix the error scoping and this is good to go.

⚡ **Requesting human review** — this touches round lifecycle and VTXO sweep propagation (protocol-critical paths).

internal/infrastructure/db/service.go
```diff
@@ -360,15 +362,17 @@ func NewService(config ServiceConfig, txDecoder ports.TxDecoder) (ports.RepoMana
...
+		eventStore:              eventStore,
+		roundStore:              roundStore,
+		vtxoStore:               vtxoStore,
+		scheduledSessionStore:   scheduledSessionStore,
+		offchainTxStore:         offchainTxStore,
+		txDecoder:               txDecoder,
+		convictionStore:         convictionStore,
+		assetStore:              assetStore,
+		intentFeesStore:         intentFeesStore,
+		batchUpdateHandler:      newUpdateHandler[domain.Round](),
```
louisinger, line 374:
```suggestion
		roundUpdateHandler:      newUpdateHandler[domain.Round](),
```
nit: weird to mix round and batch. either we rename all or we keep "round" for clarity ?

internal/infrastructure/db/update_handler.go
```diff
@@ -0,0 +1,34 @@
...
+	u.handler = handler
+}
+
+func (u *updateHandler[T]) dispatch(data T) {
+	u.lock.RLock()
+	defer u.lock.RUnlock()
+	if u.handler == nil {
+		return
+	}
+	go u.handler(data)
```
louisinger: dispatch seems to be called only in background update tasks running in go routines. can we avoid spawning a new one ?

internal/infrastructure/db/service.go
```diff
@@ -493,7 +513,14 @@ func (s *service) updateProjectionsAfterOffchainTxEvents(events []domain.Event)
 	ctx := context.Background()
 	offchainTx := domain.NewOffchainTxFromEvents(events)
 
-	if err := s.offchainTxStore.AddOrUpdateOffchainTx(ctx, offchainTx); err != nil {
+	var err error
+	defer func() {
+		if err == nil {
+			s.offchainTxUpdateHandler.dispatch(*offchainTx)
+		}
+	}()
```
louisinger: not clear to me why should be in defer instead of at the end ? 
IMO this pattern is : 
* not easy to read
* prone to error if u forget to properly assign the err (we saw this in application layer)

the update functions are running in their own go routines, isn't ok to sync the dispatch ?

louisinger: also what I'm worried is dispatch running in go routines : so it means if u have 2 call to `updateProjections...`: `U0` and `U1`. it would be possible that the dispatch of U1 completes before dispatch of U0. maybe it's not a problem now but if one day we need to preserve the order of dispatch (first called, first executed), we may be fucked up by this ?

internal/infrastructure/db/service.go
```diff
@@ -503,7 +530,9 @@ func (s *service) updateProjectionsAfterOffchainTxEvents(events []domain.Event)
 	case offchainTx.IsAccepted():
 		spentVtxos := make(map[domain.Outpoint]string)
 		for _, tx := range offchainTx.CheckpointTxs {
-			txid, ins, _, err := s.txDecoder.DecodeTx(tx)
+			var txid string
+			var ins []ports.TxIn
```
louisinger: that's what I don't like using the defer pattern

internal/infrastructure/db/service.go
altafan: batch and offchain events are handled by 2 different handlers that don't compete with each other. If a batch is moved to the next stage (like finalized) and at the same time an offchain tx is done, events will be sent through streams and their  order doesn't really matter because they're 2 different processes.

altafan: re defer pattern: i went for it because this way there's a single place where we call it, otherwise i would have need to add it in at least 2 different places.

I think the right tradeoff is using a `shouldDispatch bool` var instead of the err, this way the changes are much less invasive

louisinger (approved):

PR #1041, opened 2026-04-23
Add: configurable postgres connection pool limits

chris-ricketts:
## Summary
- set explicit Postgres connection pool limits for the main and event stores
- add env vars for tuning pool size and connection lifetime
- document the new settings in the README and dev env examples
## Why
This makes Postgres connection management more predictable across deployments and gives operators a simple way to tune pool behavior for their environment.
## Configuration
New environment variables:
- `ARKD_PG_DB_MAX_OPEN_CONN`
- `ARKD_PG_DB_MAX_IDLE_CONN`
- `ARKD_PG_DB_CONN_MAX_IDLE_MINS`
- `ARKD_PG_DB_CONN_MAX_LIFE_MINS`

Defaults:
- max open connections: `50`
- max idle connections: `50`
- max idle time: `5` minutes
- max lifetime: `30` minutes

## Summary by CodeRabbit

* **New Features**
  * New environment variables enable fine-grained control over PostgreSQL connection pooling, including limits on open and idle connections plus connection lifetime parameters.

* **Documentation**
  * Configuration documentation expanded with details on PostgreSQL connection pool settings and their default values.

PR #1042, opened 2026-04-23
Fix: discard interrupted SQLite connections and make asset reads stable under churn

chris-ricketts:
## Summary
- Replace transactional asset read with single query
- Separate SQLite read and write paths so reads can stay responsive while writes continue through a single writer
- Enable WAL mode so reads do not block writes
- Ensure canceled or interrupted SQLite operations do not return interrupt tainted connections back into the pool, which keeps later queries from failing unpredictably due to racing SQLite clearing the interrupted flag on the connection.

## Testing
- Add e2e coverage for repeated get-asset query churn
- Add sqlite tests for canceled read interruption

## Summary by CodeRabbit

* **Performance**
  * Optimized asset supply queries to execute in a single database operation instead of multiple sequential queries.
  * Improved SQLite connection pool management with separate read and write connection pools for better concurrency performance.

* **Bug Fixes**
  * Enhanced timeout and cancellation handling for concurrent database operations with improved connection cleanup.

* **Tests**
  * Added stress tests for concurrent asset queries and database operations to validate system reliability under load.

### DIFF
--- a/internal/infrastructure/db/postgres/asset_repo.go
+++ b/internal/infrastructure/db/postgres/asset_repo.go
@@ -125,30 +125,24 @@ func (r *assetRepository) GetAssets(
 	if len(assetIds) == 0 {
 		return nil, nil
 	}
-	var assets []domain.Asset
-	txBody := func(querierWithTx *queries.Queries) error {
-		rows, err := querierWithTx.SelectAssetsByIds(ctx, assetIds)
-		if err != nil {
-			return err
-		}
-		assets = make([]domain.Asset, 0, len(rows))
-		for _, row := range rows {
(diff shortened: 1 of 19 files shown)


chris-ricketts:
Integration tests with the Postgres backend are expected to fail until #1041 is merged and this branch is rebased.

chris-ricketts:
CI run is mangled - looking at the wrong code for some reason. Maybe because I force pushed and its cached something. I'll push some Rabbit fixes and see if it fixes it. All working locally.

chris-ricketts:
Closing and re-opening with new branch because CI is borked

PR #1043, opened 2026-04-23
Fix: discard interrupted SQLite connections and make asset reads stable under churn

chris-ricketts:
## Summary
- Replace transactional asset read with single query
- Separate SQLite read and write paths so reads can stay responsive while writes continue through a single writer
- Enable WAL mode so reads do not block writes
- Ensure canceled or interrupted SQLite operations do not return interrupt tainted connections back into the pool, which keeps later queries from failing unpredictably due to racing SQLite clearing the interrupted flag on the connection.

## Testing
- Add e2e coverage for repeated get-asset query churn
- Add sqlite tests for canceled read interruption

## Summary by CodeRabbit

* **Bug Fixes**
  * Assets that exist only in spent entries now report zero supply
  * Metadata parsing is now tolerant of malformed metadata and won't corrupt asset records

* **Performance**
  * Consolidated asset amount lookups into a single query to reduce query churn and improve listing efficiency

* **Reliability**
  * Safer DB read/write handling and connection discard on interrupted operations to prevent tainted connections

* **Tests**
  * Added concurrent asset-query stress test and regression test ensuring canceled reads discard connections


ghost (changes requested):
## Arkana Code Review — arkade-os/arkd#1043

**Reviewer:** Arkana (aggressive review mode)
**Verdict:** Request changes — one correctness bug on the commit-success path, plus a missing safety pragma. Protocol-critical files touched; human sign-off required.

---

### 🔴 BUG — `execTx`: successful commit returns error if `closeConn` fails

**`internal/infrastructure/db/sqlite/utils.go` — new `execTx`, around the commit path (approx lines 152-167 in the new file)**

After `tx.Commit()` succeeds, the code calls `closeConn(conn, false)`. If `closeConn` returns an error (e.g., the connection is already done, a file-descriptor issue, etc.), `execTx` returns that error to the caller:

```go
if err := tx.Commit(); err != nil { ... }

if err := closeConn(conn, false); err != nil {
    return err   // ← commit already succeeded!
}
return nil
```

The caller sees an error and may retry the write. For protocol-critical operations in `vtxo_repo.go` (SpendVtxos, SettleVtxos, SweepVtxos) and `round_repo.go` (AddOrUpdateRound), this means **a successfully committed transaction could be retried**, leading to double-spend attempts, duplicate round entries, or at minimum confusing error propagation.

**Fix:** After a successful commit, log `closeConn` errors but don't return them. The data is already persisted — the connection cleanup is best-effort.

```go
if err := tx.Commit(); err != nil { ... }

if err := closeConn(conn, false); err != nil {
    log.WithError(err).Warn("failed to close connection after successful commit")
}
return nil
```

---

### 🟡 MEDIUM — Missing `PRAGMA busy_timeout` on both pools

**`internal/infrastructure/db/sqlite/utils.go` — `OpenDb()` function**

WAL mode is correctly enabled for concurrent read/write, but there's no `busy_timeout` set on either pool. Without it, any `SQLITE_BUSY` error (e.g., during WAL checkpointing, or a brief write-lock overlap) causes an **immediate failure** rather than a short wait-and-retry.

The write path has retry logic in `execTx` (which handles `isConflictError`), but the read path via `withReadQuerier` has **no retry at all**. Under heavy load with WAL checkpointing, a read could sporadically fail with `SQLITE_BUSY`.

**Recommendation:** Add `PRAGMA busy_timeout = 5000;` (or similar) on both the read and write connections right after enabling WAL:

```go
if _, err := writeDB.Exec(`PRAGMA journal_mode = WAL;`); err != nil { ... }
if _, err := writeDB.Exec(`PRAGMA busy_timeout = 5000;`); err != nil { ... }
if _, err := readDB.Exec(`PRAGMA busy_timeout = 5000;`); err != nil { ... }
```

---

### 🟡 MEDIUM — Postgres `AssetAmount` field typed as `interface{}`

**`internal/infrastructure/db/postgres/sqlc/queries/query.sql.go` — `SelectAssetsWithUnspentAmountsByIdsRow`**

The Postgres version of the new query scans `COALESCE(v.asset_amount::TEXT, '0')` into `interface{}`:

```go
type SelectAssetsWithUnspentAmountsByIdsRow struct {
    ...
    AssetAmount    interface{}
}
```

This requires the runtime `stringifyAssetAmount` type-switch in `asset_repo.go`. The SQLite version correctly uses `sql.NullString`. Consider adding a sqlc override or casting so the Postgres generated code also gets `string` or `sql.NullString`. The `interface{}` approach is fragile — a driver upgrade could change the concrete type returned.

---

### 🟢 LOW — `isInterruptError` relies on heuristic string matching

**`internal/infrastructure/db/sqlite/utils.go` — `isInterruptError()`**

```go
errMsg := strings.ToLower(err.Error())
return strings.Contains(errMsg, "interrupted") || strings.Contains(errMsg, "sqlite_interrupt")
```

This is inherently fragile — any error message containing "interrupted" (even unrelated ones) would trigger connection discard. The `modernc.org/sqlite` driver does expose `sqlite3.Error` with a code field. Consider unwrapping to check the actual SQLite error code (`SQLITE_INTERRUPT = 9`) for a more robust check. The context-cancellation checks above are solid; this is just about the fallback heuristic.

---

### 🟢 OBSERVATION — Supply calculation semantics look correct

The old approach (per-asset `SelectAssetSupply` / `SelectAssetAmounts`) computed supply via separate queries inside a transaction. The new approach (`SelectAssetsWithUnspentAmountsByIds`) uses a single LEFT JOIN returning one row per unspent VTXO per asset, with client-side aggregation via `indexByID`. The semantics are equivalent:

- Both filter `spent = false AND asset_amount > 0`
- LEFT JOIN correctly produces a NULL row for assets with zero unspent VTXOs (handled by `Supply: *big.NewInt(0)`)
- The `testAssetRepositorySpentOnlySupply` test correctly covers the spent-only edge case

The Postgres query joins `vtxo_vw` (which itself LEFT JOINs `asset_projection`), which is semantically equivalent to the old INNER JOIN on `asset_projection` + `vtxo` since the `asset_amount > 0` filter excludes the NULL/0 rows from the LEFT JOIN.

---

### 🟢 OBSERVATION — Connection discard mechanism is sound

The `closeConn(conn, discard)` pattern using `conn.Raw(func(any) error { return driver.ErrBadConn })` to force `database/sql` to evict a tainted connection is a known, correct pattern. The `cancellation_test.go` validates this properly.

---

### ⚠️ PROTOCOL-CRITICAL — Human review required

This PR touches **vtxo_repo.go** and **round_repo.go**, which contain protocol-critical code paths:

- **vtxo_repo.go**: `AddVtxos`, `SpendVtxos`, `SettleVtxos`, `SweepVtxos`, `UnrollVtxos`, `GetVtxos`, `GetLeafVtxosForBatch` — all core VTXO lifecycle operations
- **round_repo.go**: `AddOrUpdateRound`, `GetRoundWithId`, `GetRoundForfeitTxs`, `GetRoundConnectorTree`, `GetRoundVtxoTree`, `GetSweepableRounds` — round lifecycle, forfeit paths, connector trees

The changes are mechanical (wrapping existing calls in `withReadQuerier`/`withWriteQuerier` and `execTx(ctx, r.db.Write(), ...)`), but given that **bugs here lose real money**, a human must verify each call site.

---

### Summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🔴 Bug | `execTx` returns error after successful commit if `closeConn` fails — can cause retried writes |
| 2 | 🟡 Medium | Missing `PRAGMA busy_timeout` — reads can fail under WAL checkpoint contention |
| 3 | 🟡 Medium | Postgres `AssetAmount` as `interface{}` — fragile type assertion |
| 4 | 🟢 Low | `isInterruptError` string heuristic — consider SQLite error code check |
| 5 | ⚠️ Protocol | vtxo_repo.go + round_repo.go changes require human sign-off |

Overall architecture is solid — split read/write pools with WAL and connection-discard-on-interrupt is the right approach for SQLite under concurrent load. The e2e `TestGetAssetQueryChurn` test is thorough. Fix the commit-success-path bug (#1) before merging.

🤖 Reviewed by Arkana

ghost (approved):
## Arkana Code Review — arkade-os/arkd#1043 (commit ce43d369)

**Reviewer:** Arkana (aggressive review mode)  
**Verdict:** ✅ Approve — all critical/medium findings from previous reviews are resolved.

---

### Previous findings — final status

| # | Finding | Severity | Status |
|---|---|---|---|
| 1 | `execTx` returns error after successful commit | 🔴 HIGH | ✅ Fixed — logs warning, no longer propagates error |
| 2 | Missing `PRAGMA busy_timeout` | 🟡 MEDIUM | ✅ Fixed — set to 5000ms on both read and write pools |
| 3 | `isInterruptError` string heuristic | 🟢 LOW | ✅ Fixed — proper `sqlite.Error` code check against `SQLITE_INTERRUPT` |
| 4 | Postgres metadata `err` shadowing | 🟢 LOW | ✅ Fixed — separate `parseErr` variable with `else` guard |
| 5 | SQLite `AssetAmount` NULL handling | 🟢 LOW | ✅ Fixed — `COALESCE(v.asset_amount, '0')` makes column non-nullable, silent `continue` replaced with explicit error |

### Remaining nits (non-blocking)

#### 🟢 LOW — `busy_timeout` error messages still say "failed to enable WAL"

`sqlite/utils.go` — Both `PRAGMA busy_timeout` error paths still use the message `"failed to enable WAL"` instead of `"failed to set busy_timeout"`. Copy-paste oversight. Won't cause bugs but will confuse anyone debugging a busy_timeout failure.

#### 🟢 LOW — `err == sql.ErrNoRows` vs `errors.Is` inconsistency

`sqlite/asset_repo.go`, `sqlite/conviction_repo.go`, `sqlite/intent_fees_repo.go` — Some sentinel checks use direct `==` comparison while others (round_repo, vtxo_repo, settings_repo, scheduled_session_repo) correctly use `errors.Is`. The `withReadQuerier` wrapper can join errors via `fmt.Errorf("%w: %w", err, closeErr)`, which would prevent `==` from matching. Extremely unlikely in practice (closeConn rarely fails on non-interrupt path), but `errors.Is` is the correct Go idiom. Worth a consistency pass.

#### 🟢 LOW — Postgres `AssetAmount` still typed as `interface{}`

`postgres/sqlc/queries/query.sql.go` — A sqlc column override would make this type-safe. Follow-up item.

### ✅ Good changes in this commit

- **Test error messages** now correctly use 1-indexed `completed` counter
- **Batch interval check** simplified from `i != 0 && i%batchInterval == 0` to `completed%batchInterval == 0`
- **Typo fix** "recieved" → "received"
- **`waitForVTXOs` signature** reformatted for readability

---

### ⚠️ PROTOCOL-CRITICAL — Human review still required

This PR refactors the database layer used by `vtxo_repo.go` and `round_repo.go` through the `withReadQuerier`/`withWriteQuerier` abstraction. These are protocol-critical paths (VTXO lifecycle, round management). While the refactor is sound and well-tested, human sign-off is required before merge per protocol.

🤖 Reviewed by Arkana

ghost (approved):
## Arkana Code Review — arkade-os/arkd#1043 (commits 8b893cea..8c05582a)
**Reviewer:** Arkana (aggressive review mode)  
**Scope:** 4 new commits since last review at ce43d369

---

### Changes look solid. Approving with minor nits.

**What changed since last review:**

1. **`GetSweepTxs` → `withReadQuerier`** (`round_repo.go:424-430`): Consistent with all other read paths. ✅
2. **Options pattern for `OpenDb`** (`utils.go:42-70`): Clean functional-options API. WAL + busy_timeout move from runtime PRAGMAs to DSN `_pragma=` parameters — this is actually *more correct* because pragmas are now applied per-connection by the driver, not just at pool-open time. New pool connections get them automatically. ✅
3. **Error preservation in `execTx`** (`utils.go:185,201`): Previous code swallowed the original error on `closeConn` failure (`return closeErr`). Now wraps both: `fmt.Errorf("%w: %w", err, closeErr)`. Good fix. ✅
4. **`errors.Is` for `sql.ErrNoRows`** (`sqlite/asset_repo.go:179,195`): Correct — handles wrapped sentinels. ✅
5. **Postgres type simplification** (`postgres/asset_repo.go:161-163`, `query.sql:451`): `COALESCE(v.asset_amount, 0)::TEXT` coalesces the numeric first then casts — same semantics, but now sqlc generates `string` instead of `interface{}`, eliminating the `stringifyAssetAmount` helper. Clean. ✅
6. **`buildDSN` function** (`utils.go:133-145`): Simple, correct for modernc.org/sqlite v1.33.1 which supports `_pragma=` DSN params.

### Nits (non-blocking):

1. **Inconsistent `ErrNoRows` comparison** — `sqlite/asset_repo.go:67` still uses `err != sql.ErrNoRows` (bare equality). Same pattern exists in `postgres/asset_repo.go:69,174,188`. Not introduced by this PR, but since you're already fixing the sqlite side, consider a follow-up to catch all of them.

2. **No direct unit test for `buildDSN`** — tested indirectly via `OpenDb` calls, but a table-driven test would be cheap insurance against DSN encoding bugs (e.g. special characters in paths, multiple pragma ordering).

### Verdict

Not protocol-critical (no VTXO/signing/forfeit/exit path changes). DB connection hygiene + query optimization only. Previous review concerns are addressed. Ship it.

🤖 Reviewed by Arkana

ghost (commented):
## Arkana Code Review — arkade-os/arkd#1043 (commit ebf5df7c)
**Reviewer:** Arkana (aggressive review mode)  

### New commit: `ebf5df7c` — `chore: golangci-lint fmt`

Formatting-only change (line wrapping in `round_repo.go`, `utils.go`, `vtxo_repo.go`). No logic changes. ✅ No concerns with this commit.

### Prior findings still open

Two items from my previous reviews remain unaddressed:

1. **`vtxo_repo.go:269-271` — `GetRecoverableLiquidity` silently swallows type-assertion failures.**  
   When `amount.(int64)` fails, it returns `(0, nil)` — hiding a data/driver bug as a zero balance. `GetExpiringLiquidity` correctly returns an error in this case. These should be consistent. Suggested fix:
   ```go
   n, ok := amount.(int64)
   if !ok {
       return 0, fmt.Errorf("unexpected type %T for recoverable liquidity amount", amount)
   }
   ```

2. **`intent_fees_repo.go:47` — `err == sql.ErrNoRows` should be `errors.Is(err, sql.ErrNoRows)`.**  
   `withReadQuerier` may wrap errors, breaking the direct equality check. This would cause "no rows" to surface as a hard error instead of returning an empty `IntentFees`.

Neither is a ship-blocker but both are latent bugs that will bite under the right conditions. Would be good to address before merge.

🤖 Reviewed by Arkana

ghost (approved):
## Arkana Code Review — arkade-os/arkd#1043 (commits 21df1f73..a700225c)
**Reviewer:** Arkana (aggressive review mode)  

### New commits since last review

**`21df1f73` — fix: rabbit / arkana comments**

1. `intent_fees_repo.go:47` — `errors.Is(err, sql.ErrNoRows)` replaces bare `==` check. ✅ Correct — `withReadQuerier` may wrap errors, so `errors.Is` is required for reliable sentinel matching.

2. `vtxo_repo.go:270` — `GetRecoverableLiquidity` now returns an error on unexpected type assertion instead of silently returning 0. ✅ Good — silent nil on type mismatch was a latent bug that could mask data corruption.

3. `e2e_test.go` — Fixed indentation in `TestUnrolledVtxoRejoinBatch` (was mixing tabs/spaces inside `t.Run`). Pure formatting, no logic change. ✅

**`3b1e901e` — chore: fmt**
Formatting only. ✅

**`a700225c` — ci: increase TestGetAssetQueryChurn timeouts/deadlines**
- `recvVtxosTimeout`: 10s → 20s
- `vtxosInDBDeadline`: 5s → 10s  
- `batchInDbDeadline`: 5s → 10s

Reasonable for CI flake reduction. ✅

### Verdict
All three commits address prior review feedback cleanly. No new logic, no protocol-critical changes, no cross-repo impact. LGTM.

🤖 Reviewed by Arkana

internal/test/e2e/e2e_test.go
```diff
@@ -5421,6 +5423,341 @@ func TestAsset(t *testing.T) {
...
+				GetVtxos(ctx, indexer.WithOutpoints(outpoints))
+			require.NoError(t, err)
+
+			if len(res.Vtxos) == len(outpoints) {
+				vtxos := res.Vtxos
+				for _, v := range vtxos {
+					dbVtxos[v.Outpoint] = v
+				}
+				break
+			}
```
Dunsin-cyber, line 5863: the polling loop stops as soon as it sees all VTXOs in the db, but at that point the spent-marking write may not have finished yet, which i think is what's causing the flaky "failed to update spent vtxo in db" in CI. would it make sense to make the loop wait for both the count and the spent state before breaking?

internal/infrastructure/db/sqlite/utils.go
```diff
@@ -49,21 +160,36 @@ func extendArray[T any](arr []T, position int) []T {
...
 	for range maxRetries {
-		tx, err := db.BeginTx(ctx, nil)
+		conn, err := db.Conn(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to acquire connection: %w", err)
+		}
+
+		tx, err := conn.BeginTx(ctx, nil)
 		if err != nil {
+			_ = closeConn(conn, isInterruptError(ctx, err))
```
Dunsin-cyber: looks good overall! one thing i noticed since the write pool only has 1 connection, if closeConn fails silently after a BeginTx error that connection is never returned, and every future write blocks forever. what do you think about at least logging the error here instead of dropping it with _?

ghost (commented):
## Arkana Code Review — arkade-os/arkd#1043 (commits e47ab548..8731ce9e)
**Reviewer:** Arkana (aggressive review mode)  
**Scope:** Only the 2 new commits since last review.

---

### 🐛 Bug: `allSpent` / `allPreconf` overwrite instead of accumulate

**`internal/test/e2e/e2e_test.go:5753-5760`** (commit `e47ab548`)

The new polling logic overwrites on each iteration instead of ANDing:

```go
allSpent := true
for _, spent := range spentVtxos {
    allSpent = dbVtxos[spent].Spent  // ← overwrites; only last element matters
}

allPreconf := true
for _, unspent := range unspentVtxos {
    allPreconf = dbVtxos[unspent].Preconfirmed  // ← same bug
}
```

If `spentVtxos` has 3 entries and only the last one has `Spent=true`, this evaluates to `true` even though the first two aren't spent yet. Conversely if the last one is the slowest, it masks the others being ready. Either way, the polling condition is wrong.

**Fix:**
```go
allSpent = allSpent && dbVtxos[spent].Spent
// ...
allPreconf = allPreconf && dbVtxos[unspent].Preconfirmed
```

I flagged this exact anti-pattern in the adjacent loop in my previous review. This new commit introduces the same bug in a different location.

---

### ⚠️ Noisy log: unconditional warning on successful close

**`internal/infrastructure/db/sqlite/utils.go:179-181`** (commit `8731ce9e`)

```go
closeErr := closeConn(conn, isInterruptError(ctx, err))
log.WithError(closeErr).
    Warn("failed to close connection on unsuccessful begin transaction")
```

This logs a warning **even when `closeErr` is nil** — i.e., when the connection was successfully closed/discarded. Every failed `BeginTx` (including normal context cancellations) will produce a misleading warning with `err=<nil>`. Compare with the existing pattern 10 lines below at `~line 192` which correctly guards with `if closeErr != nil`.

**Fix:**
```go
if closeErr := closeConn(conn, isInterruptError(ctx, err)); closeErr != nil {
    log.WithError(closeErr).
        Warn("failed to close/discard connection after unsuccessful BeginTx")
}
```

---

### ✅ Good changes

- Using `indexer.WithSpentOnly()` (line ~5830) instead of manually iterating and checking `v.Spent` is cleaner and delegates filtering to the query layer. 👍
- Reducing cancellation sleep from 5ms→3ms is fine for stress test tuning.

---

**Verdict:** Two small but real bugs — one logic error in the test, one noisy log. Neither is protocol-critical. No request-changes escalation needed, but please fix before merge.

🤖 Reviewed by Arkana

internal/test/e2e/e2e_test.go
chris-ricketts: Nice spot - this was the issue for the postgres CI failure. Fixed in [863b53b](https://github.com/arkade-os/arkd/pull/1043/commits/863b53b65326b54dab99fa151afb9a856b6ae051)

internal/infrastructure/db/sqlite/utils.go
chris-ricketts: Yeah I'm not sure what can be done if `BeginTx` fails and `closeConn` fails. The intention with `closeConn` is is to discard "poisoned" connections from the pool, so a fresh one opens.

I've added the log message in [b7fd652](https://github.com/arkade-os/arkd/pull/1043/commits/b7fd65223c3b7ed2a2f05046ebc7acd4ce288dcb)

PR #1046, opened 2026-04-23
feat: add CEL-based batch_trigger gate for round start

### LINKED ISSUE #1045: feat: CEL formula to control batch trigger conditions
## Proposed Solution

Add a CEL-based formula (similar to our existing fee programs) that determines **whether the server should start a new batch round**.

The formula would have access to context variables such as:

- `intents_count` — number of pending intents queued
- `current_feerate` — current mempool fee rate (sat/vbyte)
- `time_since_last_batch` — seconds elapsed since the last batch completed
- `boarding_inputs_count` — number of pending boarding UTXOs
- `total_boarding_amount` — total sats across pending boarding inputs

**Example formula:**

```
intents_count > 1 && (current_feerate <= 2 || time_since_last_batch >= 3600)
```

This means: *"only start a batch if there's more than one intent, AND either fees are low enough or an hour has passed."*

## Why CEL?

- Operators can tune batch policy without recompiling arkd
- Works the same way as our existing CEL fee programs
- Flexible enough to express economic conditions (feerate gates) and time-based fallbacks

## Acceptance Criteria

- [ ] New `batch_trigger` CEL program in arkd config (alongside existing fee programs)
- [ ] CEL environment exposes at minimum: `intents_count`, `current_feerate`, `time_since_last_batch`
- [ ] Sane default formula so existing deployments are not broken
- [ ] Docs/config example updated

## Related Context

- Existing `ARKD_UTXO_MIN_AMOUNT` env var is a short-term mitigation but doesn't solve the batching economics
- SDK-side graceful handling of `UtxoMinAmount` is tracked separately

Kukks:
## Summary

Closes #1045.

Adds an optional `ARKD_BATCH_TRIGGER` CEL formula that decides whether the server should start a new batch round. The gate is evaluated at the top of `startRound()`. If the formula returns `false`, the round is skipped and rechecked one registration window (`session_duration / 6`) later. When unset (default), the legacy "always start a round" behaviour is preserved.

The new `pkg/ark-lib/batchtrigger` package mirrors the design of `pkg/ark-lib/arkfee` — CEL programs are compiled once at startup and reused per evaluation. Misconfigured programs fail validation in `Config.Validate()` so they never make it past server startup. Runtime eval errors fail open (logged at warn level) so a buggy formula cannot permanently halt rounds.

### CEL environment

| Variable | Type | Description |
| --- | --- | --- |
| `intents_count` | `double` | Number of pending intents queued |
| `current_feerate` | `double` | Current mempool fee rate (sat/kvbyte) |
| `time_since_last_batch` | `double` | Seconds since the last finalized batch (`0` until first batch) |
| `boarding_inputs_count` | `double` | Total pending boarding UTXOs across queued intents |
| `total_boarding_amount` | `double` | Total satoshis across pending boarding UTXOs |
| `total_intent_fees` | `double` | Sum of `(inputs - outputs)` across pending intents |

`now() -> double` is also available.

### Example

```cel
intents_count > 1.0 && (current_feerate <= 2.0 || time_since_last_batch >= 3600.0)
```

## Test plan

- [x] `go test ./pkg/ark-lib/batchtrigger/... -v -count=1` — all subtests pass (compile/parse + eval against valid & invalid fixtures, all variables accessible)
- [x] `go test ./internal/core/application/ -run TestAggregateIntentTriggerData -v` — pure aggregation helper covered (5 cases)
- [x] `go test ./internal/core/application/ -run TestEvalBatchTrigger -v` — gate evaluation incl. issue #1045 examples (9 cases)
- [x] `GOOS=linux go vet ./...` clean
- [x] `GOOS=linux go build ./...` clean
- [x] CI green on PR

## Notes on review feedback

CodeRabbit flagged the following, all addressed in `5f12bb66`:

- `collectTriggerContext` previously called `Intents().Len()` and `Intents().ViewAll()` separately, which could disagree under concurrent registrations. Both now derive from the same `ViewAll` snapshot.
- The CEL eval failure path now logs at `warn` (not `error`) — matches the "fail open" wording in the README.
- `batchtrigger.Trigger` keeps `cel.Program` in a named (unexported) field so the package's public API is just `Eval`/`Source` — callers can no longer bypass `Trigger.Eval`'s nil guard.
- `batchtrigger.Parse` now validates the output type via `Type.IsExactType(cel.BoolType)` instead of pointer equality.
- The `now()` overload doesn't need `LateFunctionBinding` because `BatchTriggerEnv` does not enable the constant folding optimizer (`cel.OptOptimize`); a comment now documents this.

The two duplicate-compilation nitpicks (compile in `Validate()` then again in `NewService`) and the "skip ViewAll when the formula doesn't reference aggregates" optimization were left as future work — the cost is microseconds at startup / per recheck, and the alternatives meaningfully grow the API surface.

## Summary by CodeRabbit

* **New Features**
  * Added `ARKD_BATCH_TRIGGER` to conditionally gate when the server starts new batch rounds via a CEL boolean formula; unset keeps legacy unconditional behavior. When the formula evaluates false, the server delays and retries later.
* **Documentation**
  * Added documentation for the `batchtrigger` CEL program, including available variables, strict typing rules, and `now()` helper; invalid formulas are rejected during startup.
* **Tests**
  * Added test coverage for trigger parsing/validation, correct evaluation outcomes, and integration of the startup gate behavior.


ghost (commented):
## 🔍 Arkana Code Review — arkd#1046

**feat: add CEL-based batch_trigger gate for round start**

Overall: **Well-structured PR** that follows the established `arkfee` CEL pattern. The package decomposition, compile-once-eval-many design, fail-open semantics, and test coverage are solid. No cross-repo breakage — `NewService` is internal and all callers are updated.

That said, this touches **round lifecycle** (when rounds start). Flagging for mandatory human review per protocol.

---

### 🟡 Issues

#### 1. Cold-start footgun: `time_since_last_batch == 0` is ambiguous (Medium)
**`pkg/ark-lib/batchtrigger/types.go:18`, `internal/core/application/service.go:2434-2438`**

After server restart, `lastBatchAt` is 0 (not persisted), so `time_since_last_batch` is 0. This makes 0 mean both "never batched since boot" and "batched 0 seconds ago" — semantically different states.

**Practical consequence:** An operator writes:
```cel
intents_count > 5.0 || time_since_last_batch >= 3600.0
```
After a restart with 3 pending intents, the server waits **up to an hour** before the first batch despite having work to do, because `time_since_last_batch` reads as 0 (recently batched) instead of ∞ (never batched).

**Suggested fix:** Initialize `lastBatchAt` to 0 but treat 0 as "infinitely stale" — i.e., when `last == 0`, set `TimeSinceLastBatch` to `math.MaxInt64` (or a large sentinel like `999999`). Alternatively, persist `lastBatchAt` to the repo/cache so it survives restarts. Document whichever choice you make prominently in the README.

#### 2. Double cache read creates minor inconsistency (Low)
**`internal/core/application/service.go:2420-2447`**

`collectTriggerContext` calls `s.cache.Intents().Len(ctx)` for `IntentsCount`, then separately calls `s.cache.Intents().ViewAll(ctx, nil)` for aggregation. Between the two calls, intents can be pushed or popped. Result: `intents_count` in the CEL context may not match the data backing `boarding_inputs_count`/`total_intent_fees`.

**Suggested fix:** Drop the `Len()` call. Use `int64(len(intents))` from the `ViewAll` result instead. Single read, always consistent.

#### 3. Redundant cache cleanup on retry (Low)
**`internal/core/application/service.go:2299-2333` vs `2528-2543`**

When the gate denies and retries, `startRound` re-enters and hits the `existingRound != nil` branch again, running the full cleanup (DeleteVtxos, Reset forfeit txs, etc.) on already-cleaned data. These ops are idempotent so it's not a bug, but it's wasted work every `sessionDuration/6` seconds while the gate denies.

**Suggested fix:** Move the gate check above the cleanup block, or add a guard (e.g., `if existingRound.Id != lastCleanedRoundId`). Not blocking.

---

### 🟢 Things Done Well

- **Compile-time validation:** Invalid CEL programs are rejected in `Config.Validate()` before the server boots. Operators get fast feedback.
- **Fail-open on eval errors:** A runtime eval failure starts the batch rather than halting it. Correct safety direction for a financial system.
- **Atomic `lastBatchAt`:** No mutex needed, correct concurrent access pattern.
- **Test coverage:** 5 aggregation cases + 9 eval cases + nil/empty/all-variables tests + JSON fixture-driven tests for the package itself. Solid.
- **`DeleteVtxos` doesn't affect intent data:** Confirmed — it operates on a separate vtxo tracking set. Pending intents survive cleanup and remain readable by the trigger.

---

### ❓ Questions for Author

1. Was the cold-start `time_since_last_batch == 0` behavior intentional? The README says "0 if no batch finalized since boot" but this makes time-based escape hatches fail on fresh starts.
2. Any plans to persist `lastBatchAt` to survive restarts? For long `time_since_last_batch` thresholds (hours), losing this on restart is operationally painful.

---

### 🔴 Human Review Required

This PR modifies **round lifecycle** — specifically when `startRound()` proceeds vs. defers. While the gate can only *delay* rounds (never corrupt them), and the fail-open design limits blast radius, this is protocol-adjacent code on a Bitcoin L2. **A human must review and approve before merge.**

cc @Kukks — nice clean PR overall. The cold-start semantics (#1) are the main thing I'd want addressed before shipping.

ghost (commented):
## 🔍 Arkana — Follow-up review (commit ae0858e)

New commit `ae0858e` is a **lint-only fix** — drops redundant embedded struct selectors in `aggregateIntentTriggerData` (`service.go:2399-2400`):

```go
// before (QF1008 lint warning)
inputAmount := it.Intent.TotalInputAmount() + boardingAmount
outputAmount := it.Intent.TotalOutputAmount()

// after
inputAmount := it.TotalInputAmount() + boardingAmount
outputAmount := it.TotalOutputAmount()
```

`TimedIntent` embeds `domain.Intent` (`ports/live_store.go:93`), so promoted method access is semantically identical. ✅ No logic change.

**Previous findings still stand** (from initial review):
1. 🟡 **Cold-start footgun** — `time_since_last_batch == 0` is ambiguous (never-batched vs just-batched)
2. 🟡 **Double cache read** — `Len()` + `ViewAll()` can diverge under concurrent intent churn

No new issues introduced. Waiting on author response to the above before approving.

ghost (commented):
## 🔍 Arkana — Follow-up review (commit 5f12bb6)

Commit `5f12bb6` ("refactor: address CodeRabbit review on batch_trigger") — **all changes are correct**.

### What changed

| Change | Verdict |
|--------|---------|
| **Double-read race fixed** — `Len()` removed, `IntentsCount` derived from `len(intents)` via single `ViewAll()` call | ✅ Exactly addresses finding #2 from initial review |
| **`Errorf` → `Warnf`** on eval failure | ✅ Consistent with fail-open semantics |
| **Unexported `program` field** — `cel.Program` was embedded, now `program cel.Program` | ✅ Proper encapsulation, no accidental interface leak |
| **`!ast.OutputType().IsExactType(cel.BoolType)`** instead of `!=` | ✅ Safer type comparison (pointer equality on `*Type` is fragile across CEL versions) |
| **Drop redundant `float64()` cast** in `now()` | ✅ `int64` → `float64` implicit conversion is fine |
| **`now()` constant-folding comment** | ✅ Good documentation |

### Outstanding

🟡 **Cold-start footgun** (finding #1) — `time_since_last_batch == 0` is still ambiguous between "never batched" and "just batched." This is an operator-facing documentation concern, not a correctness bug. Consider mentioning it in `pkg/ark-lib/batchtrigger/README.md` so operators know to use `time_since_last_batch <= 0.0` or an explicit sentinel to handle the never-batched case.

No new issues. Still flagged for **human review** per protocol (round lifecycle gate).

internal/core/domain/batchtrigger/types.go
```diff
@@ -0,0 +1,40 @@
...
+type Context struct {
+	// IntentsCount is the number of pending intents queued.
+	IntentsCount int64
+	// CurrentFeerate is the current mempool fee rate in sat/kvbyte
+	// (matches ports.WalletService.FeeRate).
+	CurrentFeerate uint64
+	// TimeSinceLastBatch is the number of seconds elapsed since the last batch
+	// was finalized. It is 0 when no batch has been finalized yet since the
+	// server started.
+	TimeSinceLastBatch int64
```
bitcoin-coder-bob, line 19:
```suggestion
	SecondsSinceLastBatch int64
```

naming nit to be specific, easy to grok the quantity of time being talked about here

internal/core/application/service.go
```diff
@@ -2362,6 +2378,96 @@ func (s *service) start() {
...
+// collectTriggerContext gathers a snapshot of the variables exposed to the
+// batch_trigger CEL program. Errors are logged and surfaced as zero values so
+// that a transient failure (e.g. a wallet RPC blip) cannot wedge the round
+// scheduler — the gate then falls through to the worst-case interpretation
+// (no fee/no intent/no boarding) which a sensible formula will reject.
+func (s *service) collectTriggerContext(ctx context.Context) batchtrigger.Context {
+	tc := batchtrigger.Context{}
+
+	if feerate, err := s.wallet.FeeRate(ctx); err != nil {
+		log.WithError(err).Warn("batch_trigger: failed to read fee rate")
```
bitcoin-coder-bob, line 2390: does not setting the feerate here cause downstream some unexpected behavior? by not erring here, then `tc.CurrentFeerate` will retain its zero value. When the CEL program is then used to determine if to kick off a round, it would then be using a likely inaccurate feerate value leading to possibly incorrect evaluation

internal/core/application/service.go
```diff
@@ -2362,6 +2378,96 @@ func (s *service) start() {
...
+			tc.TimeSinceLastBatch = now - last
+		}
+	}
+
+	// Read intents once so IntentsCount and the boarding/fee aggregates all
+	// derive from the same snapshot — using a separate Len() call would race
+	// with concurrent intent registrations.
+	intents, err := s.cache.Intents().ViewAll(ctx, nil)
+	if err != nil {
+		log.WithError(err).Warn("batch_trigger: failed to view pending intents")
```
bitcoin-coder-bob: similar comment to the one earlier in this function. If we hit this log line then return, then we have 4 variables that do not get set and retain their zero values: `tc.IntentsCount`, `tc.BoardingInputsCount`, `tc.TotalBoardingAmount`, `tc.TotalIntentFees`

internal/core/application/service.go
```diff
@@ -3318,6 +3439,8 @@ func (s *service) finalizeRound(roundId string, roundTiming roundTiming) {
 		return
 	}
 
+	s.lastBatchAt.Store(time.Now().Unix())
```
bitcoin-coder-bob: this is only set on successful rounds. If we fail a round, this is not updated. Not sure if that is the intended behavior, but info that we should surface probably in the README.

internal/core/domain/batchtrigger/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## CEL Environment
+
+A program must return `bool`. The following variables are exposed:
+
+| Variable | Type | Description |
+|----------|------|-------------|
+| `intents_count` | `double` | Number of pending intents queued |
+| `current_feerate` | `double` | Current mempool fee rate in sat/kvbyte (as reported by the wallet) |
+| `time_since_last_batch` | `double` | Seconds elapsed since the last finalized batch (`0` if no batch finalized since boot) |
```
bitcoin-coder-bob, line 38: would we want to seed `lastBatchAt` to the server start time, to that the time since last batch starts ticking right from boot up?

bitcoin-coder-bob (commented):
a nice to have in the future would be an admin endpoint to update this value on the fly

internal/core/application/service.go
Kukks: It's intended

Kukks:
> a nice to have in the future would be an admin endpoint to update this value on the fly

part of the #934 work i guess

bitcoin-coder-bob:
> > a nice to have in the future would be an admin endpoint to update this value on the fly
> 
> part of the #934 work i guess

talked with @altafan the plan is to get the PR from issue 934 merged first, then do a separate PR for this addition

altafan (approved):

PR #1051, opened 2026-04-27
fix(indexer): enforce pagination when no page param is supplied

### LINKED ISSUE #1048: Indexer: impose pagination if response is too big
Based on some config env var, the indexer should return paginated responses when the list of items is bigger than the threshold. Clietns must be aware of this, therefore requires coordination with sdks to introduce the feature

Dunsin-cyber:
Fixes #1048               

  ## Description                                                                                                                                                                                                         
  Right now if a client calls the indexer without a page param, the server just dumps back the entire result set. The fix is small: when no page is provided, the server falls back to the first page using the existing
  per-method size limit. That way responses are always bounded and clients always get back page info, so they know whether there's more to fetch.                                                                        
                       
  This will also need a bit of coordination with the SDKs as you mentioned, since older clients that never sent a page param will start seeing paginated responses.                                                      
                       
  Hi @altafan — I went with the per-method `maxPageSize` constants that were already there instead of adding a new `INDEXER_MAX_PAGE_SIZE` env var. The existing limits seemed reasonable as defaults, but happy to add 
  the env var if you wanted it tunable at runtime. Let me know.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pagination behavior to ensure pagination fields are consistently computed and applied across all scenarios, including when no explicit page parameters are provided.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/internal/core/application/indexer.go
+++ b/internal/core/application/indexer.go
@@ -1064,7 +1064,7 @@ func hashOutpoints(outpoints []Outpoint) ([]byte, error) {
 
 func paginate[T any](items []T, params *Page, maxSize int32) ([]T, PageResp) {
 	if params == nil {
-		return items, PageResp{}
+		params = &Page{PageSize: maxSize, PageNum: 1}
 	}
 	if params.PageSize <= 0 {
 		params.PageSize = maxSize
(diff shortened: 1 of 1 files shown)


altafan:
Hi @Dunsin-cyber. Sorry, but these are not the right fixes, prob due to the lack of info i left in the issue. Will try to update it so you can work on this again if you want.

Dunsin-cyber:
> Hi @Dunsin-cyber. Sorry, but these are not the right fixes, prob due to the lack of info i left in the issue. Will try to update it so you can work on this again if you want.

alright. would be waiting for a clearer info...

PR #1067, opened 2026-05-17
feat(identity): add ClaimKey method to interface

### LINKED ISSUE #1066: identity.ListKeys returns empty when keys are allocated via GetKey
### Description

`identity.Identity` has two ways to obtain a key: `NewKey` advances and persists an internal counter, `GetKey(id)` does not. `ListKeys` reads from that counter, so any caller that allocates by id through `GetKey` ends up with `ListKeys` returning an empty list even when keys are in active use.

Proposing a small interface addition (`ClaimKey(ctx, id) error`) so those callers can register the allocation. Trivial no op in the singlekey backend; real work belongs in any HD backend.

Reproduction and downstream impact: arkade-os/go-sdk#176.

Dunsin-cyber:
fixes #1066 

allow callers that allocate keys by id via `GetKey` (for example the go-sdk contract manager) to register the allocation so `ListKeys` reflects them across restarts. The singlekey backend implements `ClaimKey` as a no op since it has only one key.

Refs: arkade-os/go-sdk#180

## Summary by CodeRabbit

* **New Features**
  * Added a new public Identity API method to support additional key-management operations.

* **Tests**
  * Added test coverage validating idempotent behavior and proper error handling when identity state is uninitialized.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/arkd/pull/1067?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/client-lib/identity/identity.go
+++ b/pkg/client-lib/identity/identity.go
@@ -31,6 +31,7 @@ type Identity interface {
 	GetKeyIndex(ctx context.Context, id string) (index uint32, err error)
 	NewKey(ctx context.Context) (key *KeyRef, err error)
 	GetKey(ctx context.Context, id string) (key *KeyRef, err error)
+	ClaimKey(ctx context.Context, id string) error
 	ListKeys(ctx context.Context) (keys []KeyRef, err error)
 	SignTransaction(
 		ctx context.Context, tx string, keys map[string]string,
--- a/pkg/client-lib/identity/identity_test.go
+++ b/pkg/client-lib/identity/identity_test.go
@@ -182,20 +182,50 @@ func TestListKeys(t *testing.T) {
 }
 
 func TestClaimKey(t *testing.T) {
-	t.Run("not initialized", func(t *testing.T) {
-		svc := newTestIdentity(t)
-		err := svc.ClaimKey(t.Context(), "anything")
-		require.Error(t, err)
+	t.Run("valid", func(t *testing.T) {
+		t.Run("initialized is a no op", func(t *testing.T) {
+			svc, _ := newUnlockedTestIdentity(t)
+
+			require.NoError(t, svc.ClaimKey(t.Context(), "anything"))
+			require.NoError(t, svc.ClaimKey(t.Context(), "anything")) // idempotent
+
+			keys, err := svc.ListKeys(t.Context())
+			require.NoError(t, err)
+			require.Len(t, keys, 1)
+		})
+
+		t.Run("locked is also a no op in singlekey", func(t *testing.T) {
+			svc, _ := newUnlockedTestIdentity(t)
+			require.NoError(t, svc.Lock(t.Context()))
+			require.True(t, svc.IsLocked())
+
+			// pubkey only operation, callable while locked
+			require.NoError(t, svc.ClaimKey(t.Context(), "anything"))
+
+			keys, err := svc.ListKeys(t.Context())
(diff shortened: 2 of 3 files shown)


ghost (changes requested):
## Code Review — `feat(identity): add ClaimKey method to interface`

### Summary

The PR adds a `ClaimKey(ctx context.Context, id string) error` method to the `identity.Identity` interface. The singlekey backend implements it as a no-op (expected — one key, nothing to track). Tests cover the singlekey path. The change itself is small and well-scoped.

**However, there are cross-repo breakage concerns that must be resolved before merge.**

---

### 🔴 Critical: Cross-repo compilation breakage

**`identity/identity.go` in `arkade-os/go-sdk`** — The go-sdk's HD identity (`identity/identity.go`) implements `identity.Identity` from arkd. Adding `ClaimKey` to the interface without a corresponding implementation in go-sdk will **break go-sdk at compile time** the moment it bumps its arkd dependency.

The PR body says "Refs: arkade-os/go-sdk#176" but that PR does not exist (returns 404). There is no open PR in go-sdk with "ClaimKey" in the title either. The companion implementation **must** land in go-sdk in lockstep with this change, or at minimum the go-sdk PR must exist and be ready before this merges.

Additionally, the go-sdk HD identity is where `ClaimKey` actually matters — the HD identity needs to persist the claimed key index so `ListKeys` survives restarts. The singlekey no-op is trivial; the HD implementation is the interesting (and potentially dangerous) part. **I want to see that implementation before approving this interface change.**

**`arkade-os/dotnet-sdk`** — Also references `ClaimKey` in `NArk.Scratchpad/Program.cs` and `NArk.Swaps/Boltz/BtcTransactionBuilder.cs`. Verify the dotnet-sdk either already has this method or has a companion PR queued.

### 🟡 Minor: Test coverage gaps

1. **`identity_test.go`** — The test only covers the singlekey backend (no-op path). The real complexity is in the HD identity's `ClaimKey` which must persist state. That belongs in go-sdk's test suite, but since we're adding an interface method here, consider adding a doc comment on the interface method specifying the contract:
   - Must be idempotent
   - Must persist the key so `ListKeys` includes it after restart
   - Must return an error if the identity is not initialized

2. **Missing test: locked state** — `identity_test.go:183-198`: The test covers "not initialized" and "initialized (no-op)" but doesn't test calling `ClaimKey` on a locked identity. The singlekey impl doesn't check `IsLocked()` — is that intentional? Every other mutating method (`SignTransaction`, `SignMessage`, `Dump`) checks both `ErrNotInitialized` and `ErrIsLocked`. If `ClaimKey` is meant to work while locked, document why; otherwise add the guard.

### 🟢 What looks good

- Interface placement is correct (between `GetKey` and `ListKeys`)
- `singlekey/identity.go:157-161`: The nil-data guard matches the pattern used by every other method
- Tests are idempotent and verify `ListKeys` count doesn't inflate
- No protocol-critical code touched (no VTXO, signing, forfeit, or exit path changes)

### Action items

1. **Blocking**: Where is the go-sdk companion PR? Link it or create it before this merges.
2. **Blocking**: Verify dotnet-sdk compatibility.  
3. **Non-blocking**: Add a doc comment on `ClaimKey` in the interface specifying the behavioral contract.
4. **Non-blocking**: Consider adding a locked-state guard in singlekey's `ClaimKey` for consistency, or document why it's unnecessary.

---
*🤖 Reviewed by Arkana*

PR #1074, opened 2026-05-20
Add tx filters to indexer.GetSubscription

### LINKED ISSUE #1072: Add tx filters to indexer.GetSubscription
As follow-up to #951, we want to add filters so that clients can also be notified about "txs of interest" alongside scripts (already supported) when they call GetSubscription (and update with UpdateSubscription).

Let's use CEL formulas to define these new filters. This way, we can easily add new ones later without worrying about breaking changes.

These are the filters we want to have in place for txs:
```
/* Filter txs by op_ret - check if ARK op return exists */
has(tx.extension)

/*
* Filter txs by packet type - check if a specific packet is present
* eg. hasPacket(tx.extension, 0x00)  -> check if asset packet is present
*/
hasPacket(tx.extension, <packet type>) 

/*
* Filter by packet type and data - check if a specific packet type is present and includes specific piece of data
* eg. tx.extension[0x00].contains("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa0000") -> check if asset packet contains given asset id
*/
tx.extension[<packet type>].contains("<data>")
```

The work done for fees can be used as reference: https://github.com/arkade-os/arkd/tree/master/pkg/ark-lib/arkfee/celenv

Note: .contains() should be available only if tx.extension is defined as `map<uint32, string>`

bitcoin-coder-bob:
closes https://github.com/arkade-os/arkd/issues/1072

forked from `bob/stream-single-connection` which is up for review: https://github.com/arkade-os/arkd/pull/951

---

Adds CEL-based filtering of indexed transactions to `IndexerService.GetSubscription` / `UpdateSubscription`, complementing the existing script topic filter. A tx event is now dispatched to a listener when **any** of the listener's script topics match **or** when **any** of its CEL tx expressions evaluates to true.

The proto shape was reworked from the original oneof-based draft after altafan's review (https://github.com/arkade-os/arkd/pull/1074#pullrequestreview-4364136387). `SubscriptionFilter` is now a flat message carrying both `expressions` and a `ScriptFilter` side by side; `UpdateSubscriptionResponse` is empty. This is **intentionally breaking** against master (deletes `ModifyScripts`, `OverwriteScripts`, `ScriptsFilter`, `ScriptsFilterResult` and reshapes `SubscriptionFilter` / `UpdateSubscriptionResponse`) per altafan's request. The `verify-proto` CI job will fail on the breaking-change check; expected.

Supported expression shapes (per the issue):
- `has(tx.extension)` true when the tx carries an ARK OP_RETURN extension
- `hasPacket(tx.extension, <packetType>)` true when a specific packet type is present
- `tx.extension[<packetType>].contains(\"<hex>\")` substring match against the packet's hex-encoded body

All filter inputs are validated upfront (CEL compile, script hex parse) before any state mutation. An InvalidArgument response is therefore guaranteed not to have changed the subscription.

Changes

Area: `api-spec/protobuf/ark/v1/indexer.proto`
What changed: Flattened `SubscriptionFilter { repeated string expressions = 1; ScriptFilter scripts = 2; }`. New `ScriptFilter { repeated string add = 1; repeated string remove = 2; }`. Empty `UpdateSubscriptionResponse {}`. Deleted `ModifyScripts`, `OverwriteScripts`, `ScriptsFilter`, `ScriptsFilterResult`, and the prior `SubscriptionFilter.filter` / `UpdateSubscriptionResponse.result` oneofs. Full per-call semantics (expression overwrite, scripts add/remove with empty-clears-all sentinel, atomicity, GetSubscription init-flow deviations) are documented on `UpdateSubscriptionRequest`.
────────────────────────────────────────
Area: `internal/interface/grpc/handlers/txfilter/` (new package)
What changed: `tx` CEL env declaring `tx.extension` as a dynamic map (runtime: `map[int64]string` of packet type → hex-encoded packet bytes), `hasPacket` custom function, `Parse()` (rejects non-bool expressions and bounds eval cost via `cel.CostLimit`), `Eval()` building the activation from `extension.NewExtensionFromTx`. `tx.extension` is only populated when the tx actually carries an extension, so `has(tx.extension)` is a meaningful guard. Activation is built once per (event, listener) pair and reused across the listener's programs.
────────────────────────────────────────
Area: `internal/interface/grpc/handlers/broker.go`
What changed: Listener now holds `txFilters map[string]txfilter.Filter` keyed by the original expression string. Single broker method `installTxFilters(id, filters)` atomically replaces the set; it accepts pre-compiled filters so the handler can validate scripts + expressions upfront before any mutation. `MaxTxFiltersPerListener = 64` is enforced inside `installTxFilters` so any future caller is bounded. `matchesTx` takes a lazy `getTx` thunk and listeners with no filters skip the decode entirely.
────────────────────────────────────────
Area: `internal/interface/grpc/handlers/indexer.go`
What changed: Single `applyFilter(subscriptionID, filter, initial)` handles both `GetSubscription`'s initial-filter dispatch and `UpdateSubscription`. Flow is: compile expressions → parse `scripts.add` → parse `scripts.remove` → install expressions → addTopics → removeTopics (or removeAllTopics when both add and remove are empty). When `initial` is true the scripts.remove / clear-all branches are skipped (nothing to remove on a fresh subscription). `listenToTxEvents` lazily decodes `event.Tx` once per event (only if at least one listener has tx filters) and dispatches on `scriptMatch || txMatch`.
────────────────────────────────────────
Area: `internal/interface/grpc/handlers/indexer_test.go`
What changed: 32 `t.Run` subtests across `TestGetSubscription`, `TestUpdateSubscription`, `TestTxFilter` covering: empty filter no-op vs expression-wipe, scripts add-only / remove-only / combined / clear-all, expression dedup, atomic rollback when one input in a multi-input request is invalid, cross-field atomicity (invalid script does not mutate expressions), OR semantics (script-only / tx-only / both / neither), lazy-parse contract, hex-fallback for sweep tx PSBT decode, GetSubscription init-flow ignoring scripts.remove, cap enforcement at the gRPC entry point, and old-flow disconnect preserving tx-only listeners. Test helper `seedTxFilters` wraps `compileTxFilters` + `installTxFilters` for fixture setup.
────────────────────────────────────────
Area: `internal/interface/grpc/handlers/txfilter/*_test.go`
What changed: `TestFilter` and `TestTx` cover `has(tx.extension)`, `hasPacket`, `.contains()`, parser-rejects-invalid-CEL, parser-rejects-non-bool, and the issue's exact `hasPacket(tx.extension, <type>) / tx.extension[<type>].contains(\"<hex>\")` patterns against fixture txs. Includes a `README.md` describing the supported expression vocabulary.

Acceptance criteria (https://github.com/arkade-os/arkd/issues/1072#issuecomment-4498937060)

7/8 met as written, 1 with a deviation worth flagging:

1. **Shape**: filters are carried as `repeated string expressions` directly on `SubscriptionFilter`, alongside `ScriptFilter scripts`. This is altafan's chosen flat shape; the original issue criterion predates that review.
2. **Map key type**: `tx.extension` keys are CEL `int` (runtime `int64`), not `uint32`. Done so users can write `tx.extension[0x00]` directly (matching the issue's example literals); with uint keys they would need `0x00u`.

All other criteria are met and covered by tests.

Notes

- **Breaking proto change against master**: see top of description. Intentional per altafan's review. The `verify-proto` CI job will fail on the breaking-change check.
- **Lazy tx decode**: an event whose listeners have no tx filters incurs no PSBT/hex parsing or `MsgTx.Deserialize` cost.
- **CEL eval cost bound**: per-program `cel.CostLimit(MaxEvalCost = 1_000_000)` set at `txfilter.Parse` time; a runaway expression aborts mid-evaluation with a cost error which `matchesTx` treats as a non-match.
- **Atomicity**: validation of every input (CEL expressions, scripts to add, scripts to remove) happens before any state mutation. If any single input fails validation no mutations are applied, including valid entries in the same request.

Review follow-ups

Addressed feedback from prior automated reviews and altafan's review on this PR:

- **Flat SubscriptionFilter**: full proto reshape from the original oneof-based draft, per altafan's review comment.
- **Empty UpdateSubscriptionResponse**: dropped the `ScriptsFilterResult` / `TxFilterResult` echo per the same review; a successful ack is sufficient since the client knows what it sent.
- **Cross-field atomicity**: validation of scripts and expressions happens before any mutation; regression tests cover the case where one field is invalid and the other valid.
- **DoS guards**: `MaxTxFiltersPerListener = 64` enforced inside `broker.installTxFilters`; `cel.CostLimit(MaxEvalCost = 1_000_000)` bounds per-evaluation CPU.
- **Activation reuse**: `txfilter` exposes `NewTx(tx)` once and reuses the resulting activation across the listener's programs in `matchesTx`.
- **Sentinel error**: `ErrSubscriptionNotFound` in broker; handler call sites use `errors.Is` for NotFound classification.
- **Old-flow disconnect bug**: the old-flow defer checks both scripts and tx filters before deciding to remove the listener vs schedule a timeout. Regression test added.
- **\`event.Tx\` format fix**: `parseTxOnce` now tries base64-PSBT first (commitment/ark/checkpoint txs), with hex raw-tx as fallback (sweep txs). Hex-fallback regression test added.
- **Proto comment overhaul**: `UpdateSubscriptionRequest` now documents the full set of supported field combinations, atomicity guarantees, error cases, and the GetSubscription init-flow deviations.

Pushed back (with reasoning in PR thread):
- Duplicate query params in generated OpenAPI: pre-existing codegen artifact, upstream gateway issue.
- Misleading path `POST /v1/indexer/script/updateScripts`: routes belong to #951's surface; renaming should happen there.
- Rate limit on `UpdateSubscription`: out of scope; cap + idempotent operations bound state growth, rate limiting belongs in middleware.

## Summary by CodeRabbit

* **New Features**
  * Transaction filtering using CEL expressions for subscriptions
  * UpdateSubscription RPC accepting expressions and script add/remove in a single flat filter
  * Per-subscription tx-filter cap and retention when tx filters remain on disconnect

* **Tests**
  * Expanded tests covering subscription lifecycle, tx-filter semantics, cross-field atomicity, parsing fallbacks, and cap enforcement


ghost (changes requested):
## 🔍 Aggressive Code Review — arkade-os/arkd#1074

Reviewer: Arkana (automated, protocol-aware)

---

### Summary

This PR adds CEL-based transaction filters to `GetSubscription`, introduces a new single-connection subscription flow (empty `subscription_id` → server creates inline), adds `UpdateSubscription` RPC, and adds a `SubscriptionStartedEvent` to the stream response. The proto changes, handler logic, CEL environment, and test coverage are generally well-structured. However, there are security and correctness issues that must be addressed before merge.

---

### 🔴 CRITICAL — CEL DoS: No limits on expression count or evaluation cost

**Files:** `broker.go:102-177`, `indexer.go:532-706`

The `addTxFilters` / `overwriteTxFilters` broker methods accept unbounded `[]string` expression lists. A malicious client can:

1. **Memory exhaustion**: Call `UpdateSubscription` repeatedly with `Modify.AddExpressions`, growing `listener.txFilters` without bound. Each compiled `cel.Program` is ~1-10KB.
2. **CPU exhaustion**: `matchesTx()` (`broker.go:148-166`) evaluates ALL programs for EVERY transaction event for EVERY listener. With N listeners × M expressions × T transactions/sec, this is `O(N·M·T)` CEL evaluations.
3. **No evaluation timeout**: `celenv.Eval` (`tx_env.go:48-58`) has no context/timeout. A pathological expression (e.g., deeply nested string operations on large packet hex data) blocks the single `listenToTxEvents` goroutine, stalling ALL subscriptions.

**Required fixes:**
- Add a per-subscription expression limit (e.g., `maxTxFiltersPerListener = 64`).
- Add a per-call limit on expressions in a single `UpdateSubscription` request.
- Consider adding `cel.EvalOption(cel.OptTrackCost)` + `cel.CostLimit()` to cap evaluation cost.

---

### 🔴 HIGH — `activation()` called once per program, not once per tx

**File:** `broker.go:148-166`, `tx_env.go:47-58`

`matchesTx` calls `celenv.Eval(prg, tx)` in a loop, and `Eval` calls `activation(tx)` every time. `activation()` parses the OP_RETURN extension and serializes every packet to hex — this is non-trivial work. For a listener with M expressions, this runs M times per event instead of once.

```go
// broker.go:162 — this loop calls Eval → activation(tx) for each program
for _, prg := range programs {
    ok, err := celenv.Eval(prg, tx)
```

**Fix:** Compute the activation map once outside the loop and pass it to a new `EvalWithActivation(prg, act)` function. This changes `O(M·E)` to `O(M + E)` where E = extension parse cost.

---

### 🟡 MEDIUM — Silent error swallowing in `matchesTx`

**File:** `broker.go:162-165`

```go
ok, err := celenv.Eval(prg, tx)
if err == nil && ok {
    return true
}
```

CEL evaluation errors are silently ignored. If an expression panics, returns a non-bool, or hits an unexpected type, this silently drops the match. At minimum, log at debug level so operators can detect misbehaving expressions.

---

### 🟡 MEDIUM — OpenAPI spec has duplicate parameter names

**File:** `indexer.openapi.json:25-46`, `indexer.openapi.json:136-157`

`filter.scripts.modify.removeScripts` appears twice in both endpoint definitions. Same for `filter.txs.modify.removeExpressions`. This is likely a protobuf→OpenAPI codegen quirk (the `add_*` and `remove_*` fields map to the same query param name due to the parent `modify` wrapper), but it will confuse REST clients and may cause undefined behavior in OpenAPI code generators.

**Fix:** Verify the codegen output. If this is a codegen bug, file upstream. If hand-maintained, deduplicate.

---

### 🟡 MEDIUM — `ScriptsFilterResult.Added/Removed` returns raw input, not canonical form

**File:** `indexer.go:541-549`

```go
return &arkv1.UpdateSubscriptionResponse{
    Result: &arkv1.UpdateSubscriptionResponse_Scripts{
        Scripts: &arkv1.ScriptsFilterResult{
            Added:   modify.GetAddScripts(),    // raw input
            Removed: modify.GetRemoveScripts(),  // raw input
            All:     h.scriptSubsHandler.getTopics(subscriptionID), // formatted
        },
    },
}
```

`Added` and `Removed` return the raw user input, but `All` returns `formatTopic()`-processed scripts (from `getTopics`). A client comparing `Added` entries against `All` entries may find mismatches if the input format differs from the stored format. Return the parsed/canonical form for all three fields.

---

### 🟡 MEDIUM — Fragile error string matching

**File:** `indexer.go:464-467`, `indexer.go:570-573`

```go
if strings.Contains(err.Error(), "listener not found") {
    return status.Error(codes.NotFound, err.Error())
}
```

And:

```go
if strings.Contains(err.Error(), "not found") {
    return nil, status.Error(codes.NotFound, err.Error())
}
```

Error classification via substring matching is brittle. If `broker` error messages change, these will silently produce wrong gRPC status codes. Define sentinel errors in the broker package and use `errors.Is()`.

---

### 🟢 LOW — `parseTxOnce` is safe but could use `sync.OnceValue`

**File:** `indexer.go:800-817`

The `parseTxOnce` closure with `parsedTxAttempted` bool works correctly because `listenToTxEvents` processes events sequentially. However, `sync.OnceValue` from Go 1.21+ would make the intent clearer and be future-proof if this ever becomes concurrent.

---

### 🟢 LOW — `UpdateSubscription` permission is `write`, `SubscribeForScripts` is `write`

**File:** `permissions.go:258-262`

Both are `write` on `EntityIndexer`, which is consistent. ✅

---

### 🟢 Cross-repo Impact

Two test repos (`layerzero-usdt0-arkade-demo`, `introspector-review`) use the old two-step subscription flow (`SubscribeForScripts` → `GetSubscription(id)`). The old flow is preserved — the behavior change only applies when `subscription_id` is empty. The new `SubscriptionStartedEvent` oneof variant in `GetSubscriptionResponse` is additive and won't break existing gRPC clients (they'll ignore unrecognized oneof values). REST/SSE clients that don't handle the new variant should be checked.

**No SDK repos found with direct indexer subscription consumers**, but the proto changes will propagate on next SDK codegen. No breaking change for existing consumers.

---

### 🟢 Protocol Safety

This PR does NOT touch VTXO handling, transaction signing, forfeit paths, round lifecycle, connector trees, or exit paths. The CEL filters are read-only event filtering in the indexer subscription layer — they cannot modify transaction processing. **No protocol-critical flag needed.**

---

### ✅ Test Coverage

Excellent test coverage: 1428 lines of new tests covering:
- New flow (empty subscription_id): started event, script events, cleanup, heartbeat, invalid input
- Old flow (existing subscription_id): preserved behavior, timeout, cleanup
- `UpdateSubscription`: overwrite, modify, validation, NotFound, idempotency
- `TxFilter`: CEL compilation, evaluation, OR semantics, error resilience, atomicity
- `celenv` package: compile rejection, eval correctness, issue examples

Missing test: no test for max-expression-count limits (because the limit doesn't exist yet — see critical finding above).

---

### Verdict: **Request Changes**

The CEL DoS vector (unbounded expressions, no eval cost limit, no timeout) is the blocking issue. The `activation()` per-program overhead is a significant perf bug. The other items are medium/low severity.

🤖 Reviewed by [Arkana](https://github.com/ArkLabsHQ/arkana)

ghost (commented):
## 🔍 Incremental Review — new commit `47c2f5c8` ("go mod tidy")

Only change since my last review: `go.mod` removes `// indirect` from `github.com/google/cel-go v0.26.1`, correctly marking it as a direct dependency now that `pkg/ark-lib/indexer/celenv/` imports it. ✅ Trivial and correct.

**All findings from my previous review still apply.** The critical items remain unaddressed:

1. **🔴 CEL DoS** — no per-subscription expression limit, no eval cost cap
2. **🔴 Redundant `activation()` call** — parsed once per program instead of once per tx
3. **🟡 Silent error swallowing** in `matchesTx`
4. **🟡 Fragile error string matching** — use sentinel errors
5. **🟡 OpenAPI duplicate parameters**
6. **🟡 Raw vs canonical form mismatch** in filter results

Keeping my request-changes status until the critical items are resolved.

🤖 Reviewed by [Arkana](https://github.com/ArkLabsHQ/arkana)

bitcoin-coder-bob:

api-spec/openapi/swagger/ark/v1/indexer.openapi.json
```diff
@@ -424,6 +424,119 @@
...
+            "in": "query",
+            "style": "simple",
+            "schema": {
+              "minItems": 1,
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          },
```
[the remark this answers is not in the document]
bitcoin-coder-bob: same codegen artifact already present for ScriptsFilter.modify.removeScripts from PR #951. It's a meshapi-gateway oneof+repeated quirk, not introduced by this PR; should be filed upstream.

bitcoin-coder-bob:
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> # Get PR details and diff
> gh pr view 1074 --repo arkade-os/arkd --json title,body,files,additions,deletions,baseRefName,headRefName
> 
> echo "---FILES CHANGED---"
> gh pr diff 1074 --repo arkade-os/arkd
> ```
> 
> Length of output: 167764
> 
> `@bitcoin-coder-bob` — solid implementation overall. The CEL env design is clean, the lazy-decode thunk is a nice optimization, and the atomic rollback (compile-before-mutate) pattern is the right call. Test coverage is thorough. A few things worth addressing before merge:
> 
> ### 🐛 Bug: Duplicate query parameters in generated OpenAPI spec
> In `api-spec/openapi/swagger/ark/v1/indexer.openapi.json`, both GET subscription endpoints contain **two identical `$in: query` entries** for `filter.scripts.modify.removeScripts` and `filter.txs.modify.removeExpressions`. OpenAPI 3 treats duplicate parameter names at the same location as an invalid spec, which will break validators, SDK generators, and Swagger UI. This is a codegen artifact — worth investigating the proto/gateway annotation that produces it and filing upstream if needed.
> 
> ### ⚠️ Semantic issue: old-flow disconnect cleanup ignores tx filters
> In `indexer.go` the old-flow cleanup path only considers script topics when deciding whether to remove the listener immediately or schedule a timeout:
> 
> ```go
> // Old flow defer:
> defer func() {
>     topics := h.scriptSubsHandler.getTopics(subscriptionId)
>     if len(topics) > 0 {
>         h.scriptSubsHandler.startTimeout(subscriptionId, ...)
>     } else {
>         h.scriptSubsHandler.removeListener(subscriptionId)  // tx filters ignored
>     }
> }()
> ```
> 
> A subscription created via the old flow that is later updated to hold only tx filters (no scripts) will be destroyed on disconnect rather than placed in the timeout window. Suggested fix:
> 
> ```go
> topics := h.scriptSubsHandler.getTopics(subscriptionId)
> txFilters := h.scriptSubsHandler.getTxFilters(subscriptionId)
> if len(topics) > 0 || len(txFilters) > 0 {
>     h.scriptSubsHandler.startTimeout(...)
> } else {
>     h.scriptSubsHandler.removeListener(...)
> }
> ```
> 
> ### ⚠️ Misleading HTTP path
> `POST /v1/indexer/script/updateScripts` now routes both script and tx-filter updates. The path will confuse API consumers. Something like `/v1/indexer/subscription/update` would be more accurate and future-proof (especially as more filter types are added).
> 
> ### 💡 `Removed` echoes requested expressions, not actual
> In `applyTxFilter` (and similarly for scripts):
> 
> ```go
> Txs: &arkv1.TxFilterResult{
>     Added:   addExprs,
>     Removed: removeExprs,  // echoed, not diffed
>     All:     h.scriptSubsHandler.getTxFilters(subscriptionID),
> },
> ```
> 
> Clients removing expressions that don't exist will see them in `Removed` even though nothing changed. This is consistent with the scripts side and the idempotency is tested, but the proto comment on `TxFilterResult.removed` should document this behavior so it doesn't surprise API consumers.
> 
> ### 🔍 Minor: no rate-limiting on `UpdateSubscription`
> The PR description flags `extension.NewExtensionFromTx` eval cost but the more directly attackable surface is repeated `UpdateSubscription` calls in a tight loop, each triggering CEL compilation. Compilation is more expensive than eval for complex grammars. Worth adding a note (or a TODO in the code) even if full rate-limiting is out of scope for this PR.
> 
> ### 📝 Minor: doc mismatch in `variables.go`
> ```go
> var tx = cel.VariableWithDoc(
>     TxVariableName,
>     cel.MapType(cel.StringType, cel.DynType),   // outer tx map, string keys — correct
>     "tx.extension is a map<int, string> ...",   // describes extension, not tx itself
> )
> ```
> 
> The declaration is correct (`tx` has `string` keys because its only field is `"extension"`). The doc comment is also fine for users. But it's easy to misread `cel.MapType(cel.StringType, cel.DynType)` and wonder why integer indexing works. A short inline comment clarifying that `tx` is `{extension: map<int,string>}` wrapped as `map<string,dyn>` for CEL would prevent future confusion.
> 
> ### ✅ Positives
> * **Atomic rollback**: `compileTxFilters` validates all expressions before any listener mutation — no partial state on invalid input.
> * **Lazy decode thunk**: `parseTxOnce` parsed at most once per event regardless of how many listeners have tx filters; listeners without filters pay zero decode cost.
> * **`ErrSubscriptionNotFound` sentinel**: error wrapping with `%w` enables clean `errors.Is` checks in callers; broker_test.go is correctly updated to `ErrorIs`.
> * **`MaxTxFiltersPerListener` cap with TOCTOU-safe projection** under listener lock is solid.
> * **`MaxEvalCost` + `cel.CostLimit`**: runtime protection against runaway expressions.
> * **Test depth**: the OR-semantics and lazy-parse contract subtests are particularly valuable.
> 
> ✅ Actions performed

per the misleading http path comment: I agree the name is wrong now, but routes are defined in PR #951's branch, cleaner to rename the whole script-subscription path family together over there rather than diverge piecemeal here. Flagging here that I should fix that in #951.

As for rate limiting UpdateSucscription, feels out of scope. The MaxTxFiltersPerListener cap bounds total per-listener work; duplicate adds are idempotent so a tight loop of identical adds doesn't grow state. Rate limiting belongs in middleware/gateway not per-RPC.

bitcoin-coder-bob:
@coderabbitai @arkanaai review

pkg/ark-lib/indexer/celenv/tx_env.go
```diff
@@ -0,0 +1,102 @@
...
+func init() {
+	var err error
+	TxEnv, err = cel.NewEnv(
+		tx,
+		hasPacketFunction,
+	)
+	if err != nil {
+		panic(err)
+	}
+}
```
altafan: can we avoid to use the init pattern? And let's definitely remove the panic!!

bitcoin-coder-bob: done in [6c0a80a](https://github.com/arkade-os/arkd/pull/1074/commits/6c0a80a71ea945f90650f5055894b480f0ae9edd)

altafan (changes requested):
@bitcoin-coder-bob tested manually the tx filters and they work just es expected, sweet!

After reviewing with @Kukks, we agreed on simplifying the protos:
```
message GetSubscriptionRequest {
  string subscription_id = 1;
  SubscriptionFilter filter = 2;
}

message UpdateSubscriptionRequest {
  string subscription_id = 1;
  SubscriptionFilter filter = 2;
}

message UpdateSubscriptionResponse {}

message SubscriptionFilter {
  repeated string expressions = 1;
  ScriptFilter scripts = 2;
}

message ScriptFilter {
  repeated string add = 1;
  repeated string remove = 2;
}
```

The subscription filter can be either a list of expressions (cel formulas) that the indexer always combines with OR (i removed your comments from the snippet for sake of demonstration, but let's describe this ofc), or it can be (for now) some list of scripts to be added or removed (will be migrated to a cel formula later and we'll get rid of this field in a future bump of version).

UpdateSubscription rpc ALWAYS overwrites the expressions, if any, and takes care of adding or removing scripts just like the exitisng subscribe/unsubscribe rpcs do

altafan (approved):

PR #1079, opened 2026-05-21
fix(sweeper): move the scheduled task map to LiveStore port

Dunsin-cyber:
moves the scheduled-task id set out of a per-process map on the sweeper struct and into a LiveStore-backed ScheduledTasksStore, matching the rest of the live-store conventions.

Closes the TODO at [internal/core/application/sweeper.go:43](https://github.com/arkade-os/arkd/blob/2c9612a0009ba6cea82001c59eff34d6d9f66103/internal/core/application/sweeper.go#L41-L44)

please review @altafan @sekulicd 

## Summary by CodeRabbit

* **Refactor**
  * Move scheduled-task tracking to a claim-based persistent store and unify scheduling so all tasks follow the same claim lifecycle.
* **Bug Fixes**
  * Improve handling and logging when unscheduling or releasing tasks; failures are now reported.
* **Tests**
  * Add tests for claim semantics, idempotency, presence checks, and clearing behavior.
* **Chores**
  * Configure scheduled-task TTL and wire it into the live store setup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/arkd/pull/1079?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -947,7 +947,9 @@ func (c *Config) liveStoreService() error {
 			return fmt.Errorf("invalid REDIS_URL: %w", err)
 		}
 		rdb := redis.NewClient(redisOpts)
-		liveStoreSvc = redislivestore.NewLiveStore(rdb, c.txBuilder, c.RedisTxNumOfRetries, scheduledTaskTTL)
+		liveStoreSvc = redislivestore.NewLiveStore(
+			rdb, c.txBuilder, c.RedisTxNumOfRetries, scheduledTaskTTL,
+		)
 	default:
 		err = fmt.Errorf("unknown liveStore type")
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -928,6 +928,14 @@ func (c *Config) liveStoreService() error {
 		return fmt.Errorf("tx builder not set")
 	}
 
+	// TTL must outlive the longest sweep schedule, with a safety buffer.
+	longestSweep := c.VtxoTreeExpiry.Seconds()
+	if d := c.CheckpointExitDelay.Seconds(); d > longestSweep {
+		longestSweep = d
+	}
+
+	scheduledTaskTTL := time.Duration(longestSweep)*time.Second + 24*time.Hour
+
 	var liveStoreSvc ports.LiveStore
 	var err error
 	switch c.LiveStoreType {
@@ -939,7 +947,7 @@ func (c *Config) liveStoreService() error {
 			return fmt.Errorf("invalid REDIS_URL: %w", err)
 		}
 		rdb := redis.NewClient(redisOpts)
-		liveStoreSvc = redislivestore.NewLiveStore(rdb, c.txBuilder, c.RedisTxNumOfRetries)
+		liveStoreSvc = redislivestore.NewLiveStore(rdb, c.txBuilder, c.RedisTxNumOfRetries, scheduledTaskTTL)
 	default:
 		err = fmt.Errorf("unknown liveStore type")
 	}
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -197,7 +197,7 @@ func NewService(
 		cache:                     cache,
 		scanner:                   scanner,
 		sweeper: newSweeper(
-			wallet, repoManager, builder, scheduler, noteUriPrefix,
+			wallet, repoManager, builder, scheduler, noteUriPrefix, cache,
 		),
 		boardingExitDelay:             boardingExitDelay,
 		operatorPrvkey:                operatorSigningKey,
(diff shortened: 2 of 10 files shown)


ghost (changes requested):
## Code Review — arkade-os/arkd#1079

Solid direction: moving the scheduled-task map into `LiveStore` is the right call for multi-instance arkd. The interface is clean, tests are thorough (including the concurrency race test — nice), and the Redis `SETNX` approach is correct for atomic claiming. However, there is one **critical** bug and a few issues that need fixing before merge.

---

### 🔴 CRITICAL — Stale claims cause permanently missed sweeps after a crash

**Files:** `sweeper.go`, `redis/scheduled_tasks.go`

The old code reset the task map on every restart (`sweeper.go:59`, master):
```go
func (s *sweeper) start(ctx context.Context) error {
    s.scheduledTasks = make(map[string]struct{})  // ← wiped on restart
```

The PR removes this line. Now claims live in Redis with **TTL 0** (forever). If an arkd instance:
1. Claims a task via `AddIfAbsent` → key is set in Redis
2. Crashes before the scheduler fires the callback
3. Restarts and calls `start()` → re-discovers the sweepable batch → calls `scheduleTask` → `AddIfAbsent` returns `false` → **task is silently skipped**

No instance will ever schedule that sweep again. The operator's funds remain locked in the VTXO tree past expiry. **This is a funds-at-risk bug.**

**Fix options (pick one):**
- **(A)** Add a `Clear(ctx)` method to `ScheduledTasksStore` and call it at the top of `sweeper.start()`, same as the old `make(map[string]struct{})`. This preserves the original restart-recovery semantics.
- **(B)** Set a TTL on the Redis key equal to `sweepAt - now + margin`. This naturally expires stale claims, but adds complexity.
- **(C)** Use a Redis key prefix that includes a boot/instance ID, so each restart gets a fresh namespace. (Defeats the purpose of cross-instance dedup, though.)

I'd recommend **(A)** — it's the simplest and matches the existing semantics exactly. In a multi-instance setup, only one instance should be running `sweeper.start()` anyway (the leader), so clearing the shared set on startup is safe.

---

### 🟡 HIGH — Redis key leak on long-running instances

**File:** `redis/scheduled_tasks.go:24`

```go
return s.rdb.SetNX(ctx, scheduledTaskKey(id), "1", 0).Result()
//                                                   ^ TTL = 0 = forever
```

Even if the crash scenario above is fixed, any key whose `Remove` call fails (network blip, Redis timeout) will persist forever. Over months of operation this is unbounded growth. Consider setting a generous TTL (e.g., 30 days — well beyond any realistic sweep window) as a safety net.

---

### 🟡 MEDIUM — `removeTask` error swallowed in goroutine

**File:** `service.go:3379-3384`

The error is now logged but execution continues in the goroutine regardless. If `Remove` fails, the task remains claimed in the shared store, which means a future restart could skip it (same crash scenario, different trigger). The logging is good, but consider whether the goroutine should retry or whether this is acceptable.

---

### 🟢 Minor nits

1. **`redis/round.go:17-20`**: The `scheduledTaskKeyPrefix` constant is declared in `round.go` but only used by `scheduled_tasks.go`. Move it to `scheduled_tasks.go` for locality.

2. **Blank line:** `redis/scheduled_tasks.go:11` — extra blank line between import block and type declaration. Cosmetic only.

3. **PR title:** Has leading whitespace: `"  fix(sweeper):..."` — minor but worth cleaning up.

---

### Summary

The core approach is sound and the test coverage (including the 100-goroutine race test) is well done. The critical issue is the missing cleanup on restart — without it, a crash permanently orphans sweep tasks. Add a `Clear` method, call it on startup, and this is good to go.

⚠️ **Protocol-critical**: This PR touches sweep scheduling, which controls when the operator reclaims expired VTXO tree outputs. Requesting human review from @altafan or @sekulicd per protocol-critical policy.

ghost (approved):
## Follow-up Review — arkade-os/arkd#1079 (new commits)

All three issues from my previous review are resolved:

- ✅ **CRITICAL (stale claims after crash):** `Clear(ctx)` added to `ScheduledTasksStore` interface and called at top of `sweeper.start()` — matches the old `make(map[string]struct{})` semantics exactly. Good.
- ✅ **HIGH (Redis key leak):** `scheduledTaskTTL = 30 * 24 * time.Hour` applied to `SetNX`. Safety net in place.
- ✅ **MEDIUM (removeTask error swallowed):** `removeTask` now returns `error`, callers in `service.go:3379` log warnings appropriately.

The checkpoint sweep path now goes through `scheduleTask` unconditionally (removed the early `execute()` shortcut), which is cleaner — all paths go through the same claim→execute→release flow.

---

### One remaining minor issue

**`sweeper.go:406` — orphaned claim if `ScheduleTaskOnce` fails**

```go
return s.scheduler.ScheduleTaskOnce(task.at, func() { ... })
```

The gocron implementation (`gocron/service.go:43`) can return an error from `s.scheduler.Every(...).Do(task)`. If it does, `AddIfAbsent` already claimed the slot but `releaseClaim` is never called — the claim stays until `Clear()` on next restart (or 30-day TTL). Not critical thanks to those safety nets, but a one-liner fix:

```go
err := s.scheduler.ScheduleTaskOnce(task.at, func() { ... })
if err != nil {
    releaseClaim()
}
return err
```

---

### Previous minor nit (still open)

`scheduledTaskKeyPrefix` is declared in `redis/round.go` but only used by `redis/scheduled_tasks.go`. Worth moving for locality.

---

Overall: the critical and high-severity issues are all fixed. The code is clean, the test coverage is solid, and the claim semantics are correct. Approving.

⚠️ **Protocol-critical reminder:** This PR touches sweep scheduling (operator fund recovery from expired VTXO trees). Human approval still required from @altafan or @sekulicd before merge.

PR #1080, opened 2026-05-23
fix: tolerate wallet connection loss

### LINKED ISSUE #1031: Panic in scheduleSweepBatchOutput when nbxplorer loses DB connection
## Summary

`arkd` panics with a nil pointer dereference in `scheduleSweepBatchOutput` when `nbxplorer` temporarily loses access to its database (e.g. during DB maintenance/migration).

## Panic trace

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x9352de]

goroutine 3973311 [running]:
github.com/arkade-os/arkd/internal/core/application.(*service).scheduleSweepBatchOutput(0x182fad9fa788, 0x182fb016d600)
        /app/internal/core/application/service.go:3480 +0x17e
created by github.com/arkade-os/arkd/internal/core/application.(*service).registerEvent
```

## Root cause

`scheduleSweepBatchOutput` is a background goroutine that interacts with nbxplorer. When nbxplorer loses its DB connection (e.g. during a maintenance window or rolling restart), it returns an error or nil response that `scheduleSweepBatchOutput` does not guard against, resulting in a nil pointer dereference and a full `arkd` crash.

## Impact

- `arkd` process exits and must restart (VTXO restore watch required, ~14 min downtime observed on 2026-04-22)
- Any in-progress rounds are aborted

## Expected behaviour

`scheduleSweepBatchOutput` (and background goroutines in general) should recover from transient nbxplorer errors gracefully — log the error, skip or retry the sweep, and continue running rather than crashing the entire process.

## Fix direction

- Add nil checks / error guards before dereferencing return values from nbxplorer calls in `service.go:3480`
- Consider a `recover()` in the goroutine spawned by `registerEvent` as a safety net
- Optionally: retry logic with backoff for transient nbxplorer unavailability

## Context

Discovered during DB maintenance on 2026-04-22. nbxplorer's DB migration caused a brief connection blip; arkd's auto-unlocker triggered a wallet lock/unlock round-trip which hit this code path.

Dunsin-cyber:
## What's wrong

The code assumed calls to the wallet (and nbxplorer behind it) would never fail. So a brief connection drop could break a running flow or, worse, save wrong data to disk.

The worst case is in `scheduleSweepBatchOutput`. When it couldn't reach the wallet to ask "what block are we on?", it just made up the answer (block height 0). That scheduled the sweep at around block 40 instead of the real `currentHeight + 40` — so the sweep fires way too early and the safety window users rely on to pull their money out basically disappears.
  
Restarting didn't help, because the wrong schedule was already saved. The  same path could also hit a nil value and crash the process (#1031).

## What this changes

Instead of pretending a failed call worked, we retry — and if the wallet still can't be reached, we stop and try again later instead of saving bad data.

  - [x] Retry wallet calls automatically when the connection blips
  - [x] `waitForConfirmation` keeps trying instead of giving up on the first error
  - [x] `scheduleSweepBatchOutput` stops instead of making up a block height (fixes #1031); the round is picked back up on the next startup
  - [x] `Withdraw` does NOT auto-retry, so we never send money twice

## Summary by CodeRabbit

* **Bug Fixes**
  * Confirmation polling now respects shutdown and treats transient confirmation errors as retryable, avoiding inaccurate scheduling.
  * Sweep scheduling stops when the wallet is unavailable instead of guessing times.
  * Withdrawal calls are protected from automatic retries to prevent potential double-spends.

* **Tests**
  * Added a unit test ensuring withdrawal requests do not get retried by the client retry logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/arkd/pull/1080?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -3547,7 +3547,9 @@ func (s *service) scheduleSweepBatchOutput(round domain.Round) {
 		return
 	}
 
-	blockTimestamp, err := waitForConfirmation(context.Background(), round.CommitmentTxid, s.wallet)
+	// Use s.ctx so this poll stops on shutdown. On error we bail instead of
+	// guessing the height; the round is re-scheduled on next boot.
+	blockTimestamp, err := waitForConfirmation(s.ctx, round.CommitmentTxid, s.wallet)
 	if err != nil {
 		log.WithError(err).Errorf(
 			"wallet unavailable; cannot schedule sweep for %s — will be picked up on next startup",
--- a/internal/core/application/service.go
+++ b/internal/core/application/service.go
@@ -3549,11 +3549,11 @@ func (s *service) scheduleSweepBatchOutput(round domain.Round) {
 
 	blockTimestamp, err := waitForConfirmation(context.Background(), round.CommitmentTxid, s.wallet)
 	if err != nil {
-		log.WithError(err).Warnf(
-			"failed to wait for confirmation of commitment tx %s, schedule task time may be inaccurate",
+		log.WithError(err).Errorf(
+			"wallet unavailable; cannot schedule sweep for %s — will be picked up on next startup",
 			round.CommitmentTxid,
 		)
-		blockTimestamp = &ports.BlockTimestamp{Time: time.Now().Unix()}
+		return
 	}
 
 	var expirationTimestamp int64
--- a/internal/core/application/sweeper.go
+++ b/internal/core/application/sweeper.go
@@ -470,15 +470,16 @@ func (s *sweeper) createBatchSweepTask(commitmentTxid, vtxoTreeRootTxid string) 
 				return
 			}
 
-			// schedule AFTER the root input is confirmed
+			// Wait for the root input to confirm (s.ctx so it stops on shutdown).
+			// On error, bail instead of guessing the height; re-scheduled on next boot.
 			rootInput := vtxoTree.Root.UnsignedTx.TxIn[0].PreviousOutPoint.Hash.String()
-			blockTimestamp, err := waitForConfirmation(context.Background(), rootInput, s.wallet)
+			blockTimestamp, err := waitForConfirmation(s.ctx, rootInput, s.wallet)
 			if err != nil {
-				log.WithError(err).Warnf(
-					"failed to wait for confirmation of batch input tx %s, schedule task time "+
-						"may be inaccurate", rootInput,
+				log.WithError(err).Errorf(
+					"wallet unavailable; cannot schedule sweep for batch input tx %s — "+
+						"will be picked up on next startup", rootInput,
 				)
-				blockTimestamp = &ports.BlockTimestamp{Time: time.Now().Unix()}
+				return
 			}
 
 			var expirationTimestamp int64
--- a/internal/core/application/utils.go
+++ b/internal/core/application/utils.go
@@ -542,6 +542,10 @@ func waitForConfirmation(
 		case <-ticker.C:
 			confirmed, blockTimestamp, err := wallet.IsTransactionConfirmed(ctx, txid)
 			if err != nil {
+				// On shutdown, stop instead of retrying.
+				if ctx.Err() != nil {
+					return nil, ctx.Err()
+				}
 				log.WithError(err).Warnf(
 					"transient error checking confirmation of %s; will retry on next tick", txid,
 				)
--- a/internal/core/application/utils.go
+++ b/internal/core/application/utils.go
@@ -541,7 +541,13 @@ func waitForConfirmation(
 			return nil, ctx.Err()
 		case <-ticker.C:
 			confirmed, blockTimestamp, err := wallet.IsTransactionConfirmed(ctx, txid)
-			if confirmed && err == nil {
+			if err != nil {
+				log.WithError(err).Warnf(
+					"transient error checking confirmation of %s; will retry on next tick", txid,
+				)
+				continue
+			}
+			if confirmed {
 				log.Debugf(
 					"tx %s confirmed at block height %d, block time %d",
 					txid,
@@ -549,9 +555,6 @@ func waitForConfirmation(
 					blockTimestamp.Time,
 				)
 				return blockTimestamp, nil
-			}
-			if err != nil {
-				return nil, err
 			}
 		}
 	}
--- a/internal/infrastructure/wallet/wallet_client.go
+++ b/internal/infrastructure/wallet/wallet_client.go
@@ -5,11 +5,13 @@ import (
 	"encoding/hex"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/arkade-os/arkd/internal/core/domain"
 	arklib "github.com/arkade-os/arkd/pkg/ark-lib"
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/wire"
+	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
 	log "github.com/sirupsen/logrus"
 
 	arkwalletv1 "github.com/arkade-os/arkd/api-spec/protobuf/gen/arkwallet/v1"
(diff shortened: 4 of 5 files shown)


ghost (changes requested):
## Arkana Code Review — arkade-os/arkd#1080

Overall direction is **correct and important** — the old `time.Now()` fallback on block-height schedulers was a real bug that could shorten the safety window. Retrying transient wallet errors instead of bailing is the right call. But there are issues that need fixing before merge.

---

### 🔴 MUST FIX

**1. `waitForConfirmation` now retries forever with `context.Background()` — no escape hatch**
`internal/core/application/service.go:3550` — `scheduleSweepBatchOutput` passes `context.Background()`. With the old code, a wallet error would break the loop and return. With this PR, a wallet error logs a warning and `continue`s, meaning the goroutine polls indefinitely if the wallet stays down or the txid is somehow invalid.

This is a goroutine leak. If the wallet goes down for hours, you accumulate one stuck goroutine per round that completes during the outage. Since `scheduleSweepBatchOutput` is called via `go s.scheduleSweepBatchOutput(round)` at `service.go:349`, these are fire-and-forget goroutines with no cancellation.

**Fix:** Either:
- Add a max retry count or timeout context in `scheduleSweepBatchOutput` (e.g., `context.WithTimeout(ctx, 1*time.Hour)`), then let the `return` path handle it as before — bail and rely on next-startup recovery.
- Or add a retry ceiling inside `waitForConfirmation` itself (e.g., N consecutive errors → return error).

**2. Same bug exists in `sweeper.go:475-482` — not fixed by this PR**
`internal/core/application/sweeper.go:475-482` has the **exact same pattern** this PR fixes in `service.go:3552-3556`:
```go
blockTimestamp, err := waitForConfirmation(context.Background(), rootInput, s.wallet)
if err != nil {
    blockTimestamp = &ports.BlockTimestamp{Time: time.Now().Unix()}
}
```
This is the same dangerous fallback on block-height schedulers. If you're fixing the pattern in `scheduleSweepBatchOutput`, you should fix it here too, otherwise the same bug class remains.

Similarly, `fraud.go:46-61` has a variant of this pattern that falls back to `GetCurrentBlockTime` — which is better but still problematic if the wallet is truly unreachable.

**3. `grpc_retry` dependency — does it build?**
`wallet_client.go` imports `github.com/grpc-ecosystem/go-grpc-middleware/retry`, but the PR doesn't modify `go.mod` or `go.sum`. While `go-grpc-middleware v1.4.0` is in `go.mod`, the `/retry` sub-package may or may not be included in the module's dependency graph already. If CI passes, this is fine — but please confirm. If it doesn't build, `go.sum` needs updating.

---

### 🟡 SHOULD FIX

**4. `Withdraw` retry opt-out: `grpc_retry.WithMax(0)` — verify semantics**
`wallet_client.go:458` — The comment says "opt out of the interceptor" by passing `grpc_retry.WithMax(0)`. Double-check that `WithMax(0)` means "zero retries" and not "unlimited retries" in `go-grpc-middleware/retry`. The [library source](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/retry/options.go) uses `maxRetry uint` — `0` should mean no retries, but this is safety-critical (double-spend vector). Please add a unit test or at minimum a code comment citing the library behavior.

**5. `waitForConfirmation` silently swallows non-transient errors**
`utils.go:544-548` — After this PR, ALL errors from `IsTransactionConfirmed` are treated as transient and retried. But some errors are permanent (e.g., `codes.InvalidArgument` if the txid is malformed, `codes.NotFound` if the tx doesn't exist). These will now retry forever instead of surfacing immediately.

Consider distinguishing transient vs. permanent errors, or at minimum log at `Error` level after N consecutive failures.

**6. No tests for the new retry/bail-out behavior**
The PR description checkbox "Add a safety net so a crash in a background task can't take down the server" is unchecked. There are no new tests for:
- `waitForConfirmation` retrying on transient error then succeeding
- `scheduleSweepBatchOutput` returning early when `waitForConfirmation` fails
- `Withdraw` not retrying

These are protocol-critical code paths — they deserve test coverage.

---

### ✅ LOOKS GOOD

- The `return` instead of fallback in `scheduleSweepBatchOutput` is the correct fix — relying on next-startup recovery is far safer than scheduling at the wrong height.
- Disabling retry on `Withdraw` is the right call — idempotency isn't guaranteed.
- Using `grpc_retry` at the transport level rather than hand-rolling retry loops is clean.
- Log level escalation from `Warn` to `Error` on the bail-out path is appropriate.

---

### ⚠️ PROTOCOL-CRITICAL FLAG

This PR touches **sweep scheduling** and **round lifecycle** — both protocol-critical paths. The sweep schedule determines when users lose the ability to unilaterally exit. A bug here directly impacts fund safety. **Requesting human review even though the direction is correct.**

cc @louisinger — you authored the original fallback in `5d4ecade`. The same pattern in `sweeper.go:475-482` should be addressed.

ghost (changes requested):
## Arkana Code Review — arkade-os/arkd#1080

Overall direction is **correct and important** — the old `time.Now()` fallback on block-height schedulers was a real bug that could shorten the safety window. Retrying transient wallet errors instead of bailing is the right call. But there are issues that need fixing before merge.

---

### 🔴 MUST FIX

**1. `waitForConfirmation` now retries forever with `context.Background()` — no escape hatch**
`internal/core/application/service.go:3550` — `scheduleSweepBatchOutput` passes `context.Background()`. With the old code, a wallet error would break the loop and return. With this PR, a wallet error logs a warning and `continue`s, meaning the goroutine polls indefinitely if the wallet stays down or the txid is somehow invalid.

This is a goroutine leak. If the wallet goes down for hours, you accumulate one stuck goroutine per round that completes during the outage. Since `scheduleSweepBatchOutput` is called via `go s.scheduleSweepBatchOutput(round)` at `service.go:349`, these are fire-and-forget goroutines with no cancellation.

**Fix:** Either:
- Add a max retry count or timeout context in `scheduleSweepBatchOutput` (e.g., `context.WithTimeout(ctx, 1*time.Hour)`), then let the `return` path handle it as before — bail and rely on next-startup recovery.
- Or add a retry ceiling inside `waitForConfirmation` itself (e.g., N consecutive errors → return error).

**2. Same bug exists in `sweeper.go:475-482` — not fixed by this PR**
`internal/core/application/sweeper.go:475-482` has the **exact same pattern** this PR fixes in `service.go:3552-3556`:
```go
blockTimestamp, err := waitForConfirmation(context.Background(), rootInput, s.wallet)
if err != nil {
    blockTimestamp = &ports.BlockTimestamp{Time: time.Now().Unix()}
}
```
This is the same dangerous fallback on block-height schedulers. If you're fixing the pattern in `scheduleSweepBatchOutput`, you should fix it here too, otherwise the same bug class remains.

Similarly, `fraud.go:46-61` has a variant of this pattern that falls back to `GetCurrentBlockTime` — which is better but still problematic if the wallet is truly unreachable.

**3. `grpc_retry` dependency — does it build?**
`wallet_client.go` imports `github.com/grpc-ecosystem/go-grpc-middleware/retry`, but the PR doesn't modify `go.mod` or `go.sum`. While `go-grpc-middleware v1.4.0` is in `go.mod`, the `/retry` sub-package may or may not be included in the module's dependency graph already. If CI passes, this is fine — but please confirm. If it doesn't build, `go.sum` needs updating.

---

### 🟡 SHOULD FIX

**4. `Withdraw` retry opt-out: `grpc_retry.WithMax(0)` — verify semantics**
`wallet_client.go:458` — The comment says "opt out of the interceptor" by passing `grpc_retry.WithMax(0)`. Double-check that `WithMax(0)` means "zero retries" and not "unlimited retries" in `go-grpc-middleware/retry`. The [library source](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/retry/options.go) uses `maxRetry uint` — `0` should mean no retries, but this is safety-critical (double-spend vector). Please add a unit test or at minimum a code comment citing the library behavior.

**5. `waitForConfirmation` silently swallows non-transient errors**
`utils.go:544-548` — After this PR, ALL errors from `IsTransactionConfirmed` are treated as transient and retried. But some errors are permanent (e.g., `codes.InvalidArgument` if the txid is malformed, `codes.NotFound` if the tx doesn't exist). These will now retry forever instead of surfacing immediately.

Consider distinguishing transient vs. permanent errors, or at minimum log at `Error` level after N consecutive failures.

**6. No tests for the new retry/bail-out behavior**
The PR description checkbox "Add a safety net so a crash in a background task can't take down the server" is unchecked. There are no new tests for:
- `waitForConfirmation` retrying on transient error then succeeding
- `scheduleSweepBatchOutput` returning early when `waitForConfirmation` fails
- `Withdraw` not retrying

These are protocol-critical code paths — they deserve test coverage.

---

### ✅ LOOKS GOOD

- The `return` instead of fallback in `scheduleSweepBatchOutput` is the correct fix — relying on next-startup recovery is far safer than scheduling at the wrong height.
- Disabling retry on `Withdraw` is the right call — idempotency isn't guaranteed.
- Using `grpc_retry` at the transport level rather than hand-rolling retry loops is clean.
- Log level escalation from `Warn` to `Error` on the bail-out path is appropriate.

---

### ⚠️ PROTOCOL-CRITICAL FLAG

This PR touches **sweep scheduling** and **round lifecycle** — both protocol-critical paths. The sweep schedule determines when users lose the ability to unilaterally exit. A bug here directly impacts fund safety. **Requesting human review even though the direction is correct.**

cc @louisinger — you authored the original fallback in `5d4ecade`. The same pattern in `sweeper.go:475-482` should be addressed.

Dunsin-cyber:
hi @altafan , please review

PR #1082, opened 2026-05-26
Enforce canonical encoding for Arkade payloads

msinkec:
Enforces that the Arkade `OP_RETURN` payload has a single canonical byte encoding, closing encoding-level malleability. The guiding invariant: **any payload arkd accepts must re-serialize to exactly the bytes it parsed.**

### Changes

- **Canonical varints** — new shared `internal/varint.ReadCanonical` rejects non-minimal LEB128, e.g. `0x80 0x00`; used by the asset and extension decoders.
- **Asset group** — rejects undefined presence bits and a metadata flag set over an empty list.
- **Extension framing** — rejects trailing bytes after the payload push and a non-minimal `OP_RETURN` push opcode.

All checks are O(1) on the decode path, with no re-serialization.

### Scope

Encoding canonicality only. List and packet ordering are unchanged: those already round-trip, and asset-group order is semantically meaningful as issuance index.

### Testing

- Targeted rejection tests per source, plus canonical-acceptance tests.
- Round-trip fuzz oracles for both the packet and the extension: `parse(x) ok ⇒ serialize(parse(x)) == x`; approximately 11M execs each, with no counterexample.

### Compatibility

arkd already emits canonical bytes, so no legitimate flow is affected. This only tightens acceptance of malformed or third-party input. It assumes there are no non-canonical historical on-chain payloads to re-index.

## Release Notes

* **Bug Fixes**
  * Stricter validation for asset groups, packets, and extensions: now rejects non-canonical encodings including invalid presence flags, empty metadata with flag set, and non-minimal varint encodings.
  * Enhanced serialization integrity checks ensure parsed data re-serializes to exactly the original input.

* **Tests**
  * Added comprehensive tests validating canonical encoding across asset groups, packets, and extensions.
  * Added fuzz testing for serialization round-trip integrity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/arkd/pull/1082?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/ark-lib/asset/asset_group.go
+++ b/pkg/ark-lib/asset/asset_group.go
@@ -236,6 +236,14 @@ func newAssetGroupFromReader(r *bytes.Reader) (*AssetGroup, error) {
 		return nil, err
 	}
 
+	// Canonical encoding: only the three defined presence bits may be set.
+	const definedPresenceMask = maskAssetId | maskControlAsset | maskMetadata
+	if presence&^definedPresenceMask != 0 {
+		return nil, fmt.Errorf(
+			"non-canonical asset group: undefined presence bits set in 0x%02x", presence,
+		)
+	}
+
 	var assetId *AssetId
 	var controlAsset *AssetRef
 	var metadata []Metadata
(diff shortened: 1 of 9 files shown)


msinkec:
> ### 🔴 P0 — Correctness Gap
 pkg/ark-lib has no remaining direct parser use of binary.ReadUvarint; the only real call is inside internal/varint.ReadCanonical, which is the intended wrapper. Both extension and asset varint decode paths are covered by it.

msinkec:
Other issues raised by Arkana can be ignored IMO.

PR #1083, opened 2026-05-26
Startup retry loop for `nbxplorer` connection

s373nZ:
Avoids an `arkd-wallet` crash and restart by ensuring `nbxplorer` can serve RPCs before continuing with a connection. Retries 30 times at 5 second intervals for a total of 2.5 minutes, leaving the orchestrator to restart afterward.

## Summary by CodeRabbit

* **New Features**
  * Introduced a bounded connection retry mechanism for the Bitcoin service: multiple reconnection attempts with wait intervals, per-attempt warnings on failure, and clearer error reporting when all retries are exhausted.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/arkd/pull/1083?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/arkd-wallet/core/infrastructure/nbxplorer/service.go
+++ b/pkg/arkd-wallet/core/infrastructure/nbxplorer/service.go
@@ -45,7 +45,7 @@ type nbxplorer struct {
 }
 
 const (
-	nbxplorerMaxRetries   = 30
+	nbxplorerMaxRetries    = 30
 	nbxplorerRetryInterval = 5 * time.Second
 )
 
--- a/pkg/arkd-wallet/core/infrastructure/nbxplorer/service.go
+++ b/pkg/arkd-wallet/core/infrastructure/nbxplorer/service.go
@@ -44,6 +44,11 @@ type nbxplorer struct {
 	groupID string
 }
 
+const (
+	nbxplorerMaxRetries   = 30
+	nbxplorerRetryInterval = 5 * time.Second
+)
+
 func New(rawURL string) (ports.Nbxplorer, error) {
 	rawURL = strings.TrimSuffix(rawURL, "/")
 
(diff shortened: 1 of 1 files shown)


ghost (approved):
## ✅ Approve — Startup retry loop for nbxplorer connection

**Scope**: Infrastructure startup code only. Not protocol-critical. No VTXO/signing/forfeit/round/exit paths touched.

**What it does**: Wraps the `GetBitcoinStatus` call in `New()` with a retry loop — 30 attempts × 5s = 2.5 min max wait — so `arkd-wallet` doesn't crash-loop when nbxplorer takes a moment to come up.

### Assessment

The change is correct, minimal, and solves a real operational problem. The retry logic is straightforward with no subtle bugs.

### Nits (non-blocking)

1. **No context propagation in retry loop** — `service.go:77`: The loop uses `context.Background()` and `time.Sleep()`. If the process receives a shutdown signal during startup, it'll block up to 2.5 minutes before giving up. Ideally `New()` would accept a `context.Context` (or at least the retry sleep would use a `time.After`/`select` pattern). This is a pre-existing design limitation since `New()` never took a context, so not blocking this PR on it — but worth a follow-up.

2. **No test coverage** — There are zero `*_test.go` files under `pkg/arkd-wallet/core/infrastructure/nbxplorer/`. The retry logic is simple enough to review by inspection, but a unit test with a mock HTTP server that fails N times then succeeds would be nice. Again, non-blocking given the simplicity.

3. **Fixed interval vs backoff** — 5s fixed interval is fine for "waiting for a dependent service to start." No need for exponential backoff here.

### Verified

- [x] Constants are reasonable (30 retries × 5s = 2.5 min)
- [x] Loop terminates correctly on both success and max retries
- [x] Error message on final failure includes attempt count
- [x] Warning log on each retry includes attempt number and error
- [x] `status` is properly used after the loop (line `svc.minRelayTxFee = status.MinRelayTxFee`)
- [x] Single caller in `config.go:126` — no API change, no downstream breakage
- [x] No cross-repo impact (internal infrastructure, unexported behavior change)

LGTM. Ship it.

pkg/arkd-wallet/core/infrastructure/nbxplorer/service.go
```diff
@@ -44,6 +44,11 @@ type nbxplorer struct {
 	groupID string
 }
 
+const (
+	nbxplorerMaxRetries   = 30
+	nbxplorerRetryInterval = 5 * time.Second
+)
```
Dunsin-cyber, line 50: LGTM, one little thing is `gofmt` alignment on line 48 might fail CI lint

Dunsin-cyber (commented):
tACK 138424a

ghost (approved):
**Re-review after `6c964d41`** — cosmetic whitespace fix only (aligned `=` in constant block). No logic change. Previous approval stands.

PR #1089, opened 2026-06-03
Use `minRelayTxFee` as fallback for fee estimation

s373nZ:
When `arkd-wallet` starts, it calls `nbxplorer` to estimate the current fee rate. `nbxplorer` delegates this to `bitcoind`'s `estimatesmartfee` RPC. If `bitcoind` doesn't have enough mempool data to estimate it returns nothing. `nbxplorer` passes that back as a zero fee rate, and `arkd-wallet` rejects it with `invalid fee rate received: 0.000000` and fails.

The fix passes `fallbackFeeRate` as a query parameter to the `nbxplorer` fee estimation endpoint. This is a supported but previously unused parameter in their API. When `bitcoind` can't estimate, `nbxplorer` returns the fallback value instead of zero. We use `minRelayTxFee` as the fallback.

## Summary by CodeRabbit

* **New Features**
  * Added fallback fee rate estimation to ensure more reliable transaction fee calculation, providing improved handling when primary fee sources are unavailable.

### DIFF
--- a/pkg/arkd-wallet/core/infrastructure/nbxplorer/service.go
+++ b/pkg/arkd-wallet/core/infrastructure/nbxplorer/service.go
@@ -397,7 +397,8 @@ func (n *nbxplorer) GetNewUnusedAddress(ctx context.Context, derivationScheme st
 // EstimateFeeRate retrieves fee rate from /v1/cryptos/{cryptoCode}/fees/{blockCount} endpoint.
 func (n *nbxplorer) EstimateFeeRate(ctx context.Context) (chainfee.SatPerKVByte, error) {
 	blockCount := 1
-	data, err := n.makeRequest(ctx, "GET", fmt.Sprintf("/v1/cryptos/%s/fees/%d", btcCryptoCode, blockCount), nil)
+	fallback := float64(n.minRelayTxFee) / 1000
+	data, err := n.makeRequest(ctx, "GET", fmt.Sprintf("/v1/cryptos/%s/fees/%d?fallbackFeeRate=%f", btcCryptoCode, blockCount, fallback), nil)
 	if err != nil {
 		return 0, fmt.Errorf("failed to estimate fee rate: %w", err)
 	}


ghost (approved):
## ✅ Approved

Clean, minimal fix for a real startup failure. The logic is correct and well-scoped.

### Analysis

**Unit conversion** (`service.go:400`):
```go
fallback := float64(n.minRelayTxFee) / 1000
```
`n.minRelayTxFee` is `SatPerKVByte` (sat/kVB). Dividing by 1000 yields sat/vB, which is the unit nbxplorer's fee endpoint returns (confirmed by the inverse on line 414: `resp.FeeRate * 1000` → `SatPerKVByte`). Consistent with the API contract. ✓

**10% margin consistency**: `n.minRelayTxFee` already includes the 10% margin (line 128). This is the same value used as the floor on line 415 (`max(satPerKvB, n.minRelayTxFee)`), so the fallback and floor are consistent. If bitcoind can't estimate and nbxplorer returns the fallback, we get `fallback * 1000` back as `satPerKvB`, which equals `n.minRelayTxFee`, and `max(n.minRelayTxFee, n.minRelayTxFee)` = `n.minRelayTxFee`. No surprises. ✓

**Safety net preserved**: The `resp.FeeRate <= 0` check (line 410) still catches any edge case where nbxplorer returns garbage even with a fallback. ✓

**Not protocol-critical**: Fee estimation infrastructure only — doesn't touch VTXOs, signing, forfeit paths, or round lifecycle. Worst case of a bad estimate is a slow-to-confirm tx, not fund loss.

### Nit (non-blocking)

`%f` on line 401 produces trailing zeros (e.g., `1.100000`). `%g` would be cleaner in the URL, but functionally equivalent.

### Missing

No unit tests, but the entire `nbxplorer` package has zero test files — that's a pre-existing gap, not this PR's problem.

🤖 Reviewed by Arkana

s373nZ:
Not sure 1 sat / vByte or 0.1 sat / vByte is a practical default. Would this be better a configuration variable? Maybe one changeable during runtime?

PR #1092, opened 2026-06-08
vtxo chain pagination

bitcoin-coder-bob:
> ## Merge plan: merge this before #908
>
> This PR is the **cursor-based pagination** split out of the "Scale the DAG" PR
> (**#908**) so it can land on `master` first. The two are orthogonal: this is the
> `GetVtxoChain` API feature, while #908 is the marker-DAG performance work.
>
> 1. **Merge this PR (#1092) into `master` first.** It adds the `page_token` /
>    `next_page_token` proto fields, the HMAC-signed (outpoint-bound) cursor, and a
>    full-build plus offset-slice implementation. Fully backward compatible.
> 2. **#908 then rebases onto `master`,** drops the now-duplicated pagination
>    pieces, and swaps the cursor internals from the offset-over-full-rebuild used
>    here (O(N^2) across all pages, accepted as interim) to a frontier-resume walk
>    (O(N)). The external contract (`page_token` / `next_page_token`, opaque
>    tokens) is identical, so the swap is transparent to clients.

---

## Summary

Adds **cursor-based pagination** to the indexer's `GetVtxoChain` RPC via new `page_token` / `next_page_token` fields. Extracted from the "Scale the DAG" PR (#908) so it can land on `master` independently and ahead of the marker-DAG work.

## What changed

- **Proto**: `GetVtxoChainRequest.page_token` (5) and `GetVtxoChainResponse.next_page_token` (4), plus regenerated stubs and swagger.
- **`getVtxoChain`**: builds the full chain (unchanged), then slices the requested page by an absolute offset carried in an opaque cursor. A new `paginateByOffset` helper sits alongside the existing `PageNum` `paginate`.
- **Cursor**: `encodeChainCursor` / `decodeChainCursor` produce an HMAC-signed, base64url token. The HMAC key is derived from the auth private key (stable across auth tokens), and the cursor is **bound to the outpoint** so it cannot be forged or replayed against a different chain.
- **Old and new page systems decoupled**: when a `page_token` is present the legacy `page` struct is ignored entirely (page size is fixed at `maxPageSizeVtxoChain`), so the old offset pagination can be dropped as a unit once API v2 lands.
- **Handler**: a request combining an intent with a `page_token` is rejected with `InvalidArgument`. Cursor continuation is done via the `auth_token` returned by the first call, not by re-submitting the intent proof.

## Auth-token TTL: session keepalive was added, then removed

An earlier revision of this PR carried a **pagination session keepalive**: `tokenCache.touch` / `isActive` plus a continuation fallback in `validateChainAuth` that extended a cache session on every successful page and accepted timestamp-expired tokens as long as the session stayed live. The goal was to let a long pagination outlast the 5-minute auth-token TTL.

**It has been removed.** Nobody asked for it, and it quietly undid the property the TTL was introduced for: a continuously-paginating client could keep one token alive forever, with no absolute cap. With it gone, the auth token's embedded-timestamp TTL is a hard ceiling again.

> Note for #908: the keepalive is **removed** here, not fixed, so #908 should drop it on rebase rather than carrying it forward.

### Behavioral consequence

A chain that takes longer than the auth-token TTL (default 5 minutes) to fully paginate now fails mid-way with `auth token expired`. It fails cleanly, with no panic or hang.

**Recommended client behavior (Option 1):** on `auth token expired`, re-submit the intent to mint a fresh token, then continue with the **last `page_token`** already received. The cursor is bound to the outpoint and signed with a stable server key, so a fresh token plus the saved `page_token` resumes at the exact same offset with no re-fetch. The only cost is one extra intent validation per TTL window, and the TTL stays a true hard ceiling.

**Future option, if client friction proves real (Option 2):** rotate the token. Return a freshly-signed token in each paginated response (cheap, since no intent re-validation is needed), but embed an immutable origin timestamp and refuse to rotate past `origin + maxSessionLifetime`. That gives a sliding window for legitimate pagination together with a hard cap on total session life, which is the bound the removed keepalive lacked.

**Caveat (why this likely will not bite in practice):** the TTL only becomes a constraint because pagination is O(N^2) here (the full chain is rebuilt on every page). #908's frontier-resume walk makes it O(N) and roughly 13-20x faster per its benchmarks, so a multi-minute walk drops to seconds and the token should rarely expire mid-pagination. Once #908 lands, this is unlikely to be a real concern. The TTL is also operator-configurable via `authTokenExpirySec` as a stopgap.

## API behavior

- `page_token` empty and `page` nil: full chain, no `next_page_token` (unchanged default, which the redemption path relies on).
- `page` set, or a `page_token` provided: one page is returned, and `next_page_token` is set when more pages remain. Pass it back as `page_token` to continue.
- `page_token` present: it takes precedence over and fully ignores the `page` struct.
- `intent` plus `page_token` together: `InvalidArgument`.
- The token is opaque, so when #908 swaps the cursor encoding (offset to frontier outpoints), clients are unaffected.

## Backward compatibility

Existing offset (`PageNum`) callers get identical slices to before, plus a bonus `next_page_token` they can ignore. A `page_token`, when present, takes precedence over the whole `page` struct.

## Security / hardening

- Cursors are HMAC-signed and outpoint-bound. They are left unsigned only in public-exposure deployments (no auth key), where all chain data is already publicly accessible. Withheld and private modes always have a key.
- A nil auth key on the auth path returns a clean error instead of panicking.

## Known caveat

Building the full chain on every page is O(N) per call, so paginating an N-deep chain is O(N^2). This is expected for the interim step on `master`. #908 replaces the internals with a frontier-resume walk that makes pagination O(N), and the external API does not change.

## Testing

- `TestChainCursorRoundTrip`: encode/decode round-trip, outpoint-binding rejection, tampered-signature rejection, invalid base64, and the no-HMAC-key path.
- `TestPaginateByOffset`: first, middle, last, and past-end pages, plus zero and oversized page sizes.
- `TestGetVtxoChainPagination`: page through the chain via `next_page_token` and assert the reassembled pages equal the full chain.
- `TestGetVtxoChainPageTokenIgnoresPageStruct`: a `page_token` with a tiny `page.PageSize` still returns the full max-size page, proving the page struct is ignored on the cursor path.
- `TestGetVtxoChainExpiredTokenRejectedWithPageToken`: an expired token is rejected even on a continuation, locking in that the keepalive is gone.
- `TestValidateChainAuthNilKey`: a server with no auth key returns an error instead of panicking.
- `TestGetVtxoChainRejectsPageTokenWithIntent` (handler): intent plus `page_token` is rejected with `InvalidArgument`.

`make lint` clean, `go build ./...` clean, affected test packages green.


ghost (changes requested):
## Code Review — vtxo chain pagination (`arkd#1092`)

Overall direction is solid: HMAC-signed cursors bound to outpoints, backward-compatible proto additions, good test coverage on cursor round-trips and auth keepalive. Several issues need fixing before merge.

---

### 🔴 P0 — Session keepalive attack via `touch()` before cursor validation

**`internal/core/application/indexer.go:408`** — `validateChainAuth` calls `i.tokenCache.touch(hash)` unconditionally at the end. This runs **before** `getVtxoChain` validates the cursor via `decodeChainCursor`.

A client with an **expired** auth token but a still-active cache entry can send `page_token="garbage"` on every request. The flow:
1. `validateChainAuth(token, outpoint, isPaginating=true)` — expired timestamp is tolerated, signature is valid, cache entry found → **`touch()` extends the session**
2. `getVtxoChain` → `decodeChainCursor("garbage", …)` → returns error
3. Call fails, but the session was already extended

Repeating this keeps the session alive **indefinitely** without ever fetching data. The token cache never expires.

**Fix:** Move `touch()` out of `validateChainAuth` and into `getVtxoChain` **after** successful cursor decode and page fetch. Or: validate the cursor before calling `validateChainAuth`.

---

### 🔴 P0 — Full chain rebuilt on every page request

**`internal/core/application/indexer.go:528-529`** — `getVtxoChain` calls `buildVtxoChain` (which does multiple DB queries, tree traversal, visited-set tracking) on **every** page request, then throws away everything except the requested slice.

For a VTXO with a chain of length N, fetching all pages with `page_size=1` does **N × buildVtxoChain** calls = **O(N²)** DB work total. `buildVtxoChain` itself is O(N) with multiple `GetVtxos` + `GetRoundVtxoTree` queries per iteration.

This defeats the purpose of pagination (reducing server load for large chains). At minimum, document this as a known limitation. Ideally, cache the built chain keyed by `(outpoint, cache_hash)` with a short TTL, or push pagination into the DB layer.

---

### 🟡 P1 — Intent path silently drops `page_token`

**`internal/interface/grpc/handlers/indexer.go:331-336`** — The intent code path calls `GetVtxoChainByIntent(ctx, *intent, page)` which does not accept a `pageToken` parameter. However, `page_token` is on the shared `GetVtxoChainRequest` proto message.

A client paginating via intent will send `page_token` on subsequent requests but the server ignores it, silently falling back to offset-based paging. Either:
- Wire `page_token` through to `GetVtxoChainByIntent` → `getVtxoChain`, or
- Document that cursor pagination is not supported for intent-based calls and validate: if `intent != nil && page_token != ""`, return `InvalidArgument`

---

### 🟡 P1 — No max page size enforcement (regression from `paginate`)

**`internal/core/application/indexer.go:533-535`**:
```go
pageSize := int(maxPageSizeVtxoChain)
if page != nil && page.PageSize > 0 {
    pageSize = int(page.PageSize)
}
```

The old `paginate` function used `maxPageSizeVtxoChain` as a default, but neither old nor new code caps the client-supplied size. Since this PR introduces a new pagination path, please add:
```go
if pageSize > int(maxPageSizeVtxoChain) {
    pageSize = int(maxPageSizeVtxoChain)
}
```

A client requesting `page_size=1000000` currently rebuilds the full chain and returns it all in one shot, negating pagination entirely.

---

### 🟡 P1 — Unsigned cursors when `privkey == nil`

**`internal/core/application/indexer.go:484-488`** — When `cursorHMACKey` is nil (privkey not configured), `encodeChainCursor` produces plain-JSON cursors with no HMAC. A client can forge `{"o":"any_outpoint","n":9999}` to jump to any offset.

Since `privkey == nil` means `exposurePublic` (no auth needed), this isn't a direct auth bypass. But it violates the "opaque cursor" contract and lets clients probe chain lengths. Consider using a random HMAC key as fallback when privkey is nil.

---

### 🟡 P1 — Cross-repo SDK impact

The proto change is **additive** (new `page_token`/`next_page_token` fields) so existing SDKs won't break. However:

- **.NET SDK** (`NArk.Core/Transport/GrpcClient/GrpcClientTransport.Exit.cs`) paginates with `page.Next != page.Total` loop — this still works but doesn't benefit from cursors
- **TS SDK** (`ts-sdk/src/providers/indexer.ts`) uses `page.index`/`page.size` query params — same
- **Rust SDK** uses generated proto stubs — fields are available but unused

No coordinated SDK PRs are needed for merge, but the SDKs should be updated to use cursor-based pagination to get the consistency benefits. Consider filing issues on the SDK repos.

---

### 🟢 Minor / Nits

1. **`internal/core/application/indexer.go:525`** — `paginateByOffset` with `pageSize <= 0` sets `pageSize = len(items)` which returns everything. Combined with the nil-page-and-empty-token early return at line 530, this is reachable only via `page != nil && page.PageSize == 0`. Document or clamp to `maxPageSizeVtxoChain`.

2. **`internal/core/application/indexer.go:540`** — `resp.Current` computed as `offset/max(pageSize, 1) + 1` gives wrong `Current` when offset is not page-aligned (e.g., if page size changed between requests). Since cursor-based callers shouldn't rely on page numbers, consider omitting `Current` or documenting it as approximate.

3. **`internal/core/application/token_cache.go:64`** — `touch()` iterates all outpoints in the map to update expiry. Since all outpoints share the same expiry (set by `add`), this is correct but could be simplified with a single `expiresAt` field on the entry. Low priority.

---

### ⚠️ Protocol-Critical Flag

This PR touches **VTXO chain traversal** and **auth token validation** — both protocol-critical paths. The auth keepalive change (`validateChainAuth` accepting expired tokens during pagination) is a deliberate security policy change that needs **human sign-off** even after the above issues are fixed.

**Requesting changes.** Fix P0s (session keepalive attack, or at least move `touch` after cursor validation) and P1s before re-review.

bitcoin-coder-bob:
@coderabbitai @arkanaai review

ghost:
## Review — vtxo chain pagination

Good work overall. The design is solid and this is a clean, self-contained split. Notes below, roughly priority order.

---

### 🟡 Minor: `verifyAuthTokenSignature` has no nil-guard on `authPrvkey`

`extractTokenHash` guards against a nil `authPrvkey` at the top of the function, but `verifyAuthTokenSignature` goes straight to `i.authPrvkey.PubKey()` without checking. In practice this path is only reached for `exposureWithheld`/`exposurePrivate` modes where a private key is required, so it shouldn't panic in production. But it's inconsistent with the existing pattern and a future code path change could bite you. Suggest adding the same guard at the top:

```go
if i.authPrvkey == nil {
    return "", fmt.Errorf("auth key not configured")
}
```

---

### 🟡 Minor: unsigned cursors when `privkey` is nil

When `NewIndexerService` is called with a nil private key, `cursorHMACKey` stays nil and cursors are plain unprotected JSON. Any client can craft an arbitrary offset. The outpoint binding in `decodeChainCursor` still validates the outpoint matches, but a client can skip to an arbitrary position in the chain.

This is benign for `exposurePublic` since all chain data is already public, but worth adding a comment near the `cursorKey = nil` path to make the security posture explicit, e.g.:

> // Without a private key (public-exposure deployments), cursors are unsigned.
> // Clients can craft arbitrary offsets, which is acceptable because all
> // chain data is already publicly accessible.

---

### ✅ Session keepalive design is correct

The touch-after-success ordering (cursor decoded and chain built before `tokenCache.touch`) correctly prevents a client from extending a session via spam of invalid page tokens. `TestGetVtxoChainInvalidPageTokenDoesNotExtendSession` locks this in nicely.

---

### ✅ Backward compatibility preserved

`page == nil && pageToken == ""` fast-paths to the full chain, matching the old `paginate` behaviour when `params == nil` (which also returned the full slice unchanged). The redemption path's dependency on this is safe.

---

### ✅ `page_token` + intent rejection is in the right place

Good call putting the `InvalidArgument` rejection in the gRPC handler before the request reaches the application layer. The error message clearly tells clients to switch to `auth_token` mode for continuation.

---

### ✅ O(N²) caveat

Well-documented in both the PR description and the inline code comment, with a concrete migration path to #908. Acceptable for an interim `master` landing.

---

### ✅ Tests

Coverage is thorough: round-trip, outpoint-binding rejection, tampered signature, expired-vs-active session, invalid token not extending session, and full paginate-and-reassemble integration. Happy with this.

---

**Summary:** Two minor defensive-coding items (nil guard on `authPrvkey` + unsigned-cursor comment) worth addressing before merge, but nothing blocking. The architecture, auth flow, and backward compatibility story are all clean.

> ⚠️ Auth session logic touches VTXO chain traversal — flagging for human reviewer sign-off per protocol-critical review policy, even though this doesn't touch signing, forfeit, or exit paths directly.

ghost (changes requested):
## Arkana Code Review — arkade-os/arkd#1092

**Reviewer:** Arkana (aggressive review, protocol-critical scope)
**Verdict:** Request changes — one real issue, several notes, **human sign-off required** (auth token handling + VTXO chain access).

---

### 🔴 Issue: Indefinite session extension via `touch` (security tradeoff, needs explicit decision)

**`internal/core/application/indexer.go:373`** + **`internal/core/application/token_cache.go:73-82`**

The `touch` method extends the cache entry by a full `invalidationDuration` on every successful page fetch. A client that proved ownership once and received a 5-minute auth token can keep the session alive **indefinitely** by fetching at least one page per TTL window. The signed timestamp in the auth token becomes effectively meaningless after the first page.

This is documented as intentional, but it changes the security model: auth tokens were previously bounded by their signed TTL. Now they're bounded by activity. For a deep chain (thousands of nodes), a session could remain alive for hours.

**Recommendation:** Either:
1. Add a hard ceiling: `touch` refuses to extend beyond `N * invalidationDuration` from the original token timestamp (extract the timestamp from the token, store it in the cache entry). This bounds the total session lifetime.
2. Or explicitly document this is accepted and that `RevokeTokens` is the administrative escape hatch.

This needs a human decision — I won't approve without one.

---

### 🟡 Notes (non-blocking but worth addressing)

**1. TOCTOU between `isActive` and `getOutpoints` — `internal/core/application/indexer.go:1050-1058`**

```go
if !i.tokenCache.isActive(hash) {
    return "", fmt.Errorf("auth token expired")
}
// ← sweep goroutine can delete the entry here
outpoints, _, ok := i.tokenCache.getOutpoints(hash)
if !ok {
    return "", fmt.Errorf("auth token not found")
}
```

The `isActive` check uses `RLock`, releases it, then `getOutpoints` re-acquires `RLock`. Between these calls, the sweep goroutine could delete the entry. Result: a valid paginating client gets "auth token not found" instead of success. This is **benign** (fails safe, not open), but the error message is confusing. Consider combining `isActive` + `getOutpoints` into a single lock-holding method, or at minimum, mapping the "not found" error to "expired" when `isPaginating=true`.

**2. Code duplication: `verifyAuthTokenSignature` vs `validateAuthToken` — `internal/core/application/indexer.go:999-1030`**

These two functions share ~90% of their code (base64 decode, length check, signature parse, signature verify). Only difference: `validateAuthToken` also checks the timestamp. Extract the common logic into a helper that returns the parsed message bytes, then have both callers check what they need. This reduces the surface area for future divergence bugs.

**3. Unsigned cursors in public exposure mode — `internal/core/application/indexer.go:1259-1263`**

When `cursorHMACKey` is nil (no auth private key, i.e. public exposure), `encodeChainCursor` emits an unsigned token and `decodeChainCursor` skips HMAC verification. A client can forge a cursor with any offset. This is safe because: (a) public mode has no auth, (b) the client can already control offset via `page.PageNum`, (c) outpoint binding is still JSON-checked. But it should be documented with a comment in `encodeChainCursor` explaining why this is acceptable.

**4. `maxPageSizeVtxoChain` is a default, not a cap — `internal/core/application/indexer.go:537-540`**

```go
pageSize := int(maxPageSizeVtxoChain)
if page != nil && page.PageSize > 0 {
    pageSize = int(page.PageSize)
}
```

A client can set `page.PageSize = 10000000` and get the entire chain in one response. This is consistent with the existing `paginate` function (pre-existing behavior), but since this PR refactors the pagination path, it's a good time to add `pageSize = min(pageSize, int(maxPageSizeVtxoChain))` as a cap. Not a blocker.

---

### ✅ What looks good

- **HMAC cursor design** is solid: SHA-256 HMAC, key derived from auth private key with domain separator (`"cursor-hmac"`), outpoint-bound, base64url-encoded. Prevents forging and cross-chain replay.
- **Session touch only after successful page fetch** (`indexer.go:370-374`): prevents abuse via invalid `page_token` spam. Good defensive design, and tested (`TestGetVtxoChainInvalidPageTokenDoesNotExtendSession`).
- **Intent + page_token rejection at gRPC layer** (`handlers/indexer.go:335-341`): clear error message guiding the client to use auth_token for continuation.
- **Proto field numbers** (5 and 4) don't collide with existing fields. Additive, backward compatible.
- **Test coverage** is thorough: cursor round-trip, outpoint binding, tampered signature, pagination reassembly, auth keepalive semantics, invalid token non-extension. All critical paths covered.
- **O(N²) tradeoff** is explicitly documented and the path to O(N) via #908 is clear.

### Cross-repo impact

The new `page_token`/`next_page_token` fields are **additive and backward compatible**. Existing SDK pagination loops (notably .NET SDK's `RestClientTransport.Exit.cs:19-42` and `GrpcClientTransport.Exit.cs:14-40`) continue to work with index-based pagination. SDKs should eventually be updated to prefer cursor-based pagination for correctness across chain mutations, but that's a separate effort and not blocked by this PR.

Proto copies in `rust-sdk` and `dotnet-sdk` will need regeneration when they sync, but the new fields are purely additive — no breaking change.

---

**⚠️ PROTOCOL-CRITICAL FLAG:** This PR modifies auth token validation logic and VTXO chain access patterns. Requires human review and sign-off before merge, per Arkana protocol rules.

🤖 Reviewed by Arkana

bitcoin-coder-bob:

ghost:
## Arkana Review — `vtxo chain pagination` (#1092)

Clean extraction from #908 — the scope is well-contained and the design holds up. Here are my findings:

---

### ✅ What looks good

**Cursor security model is sound.**
`encodeChainCursor` binds the cursor to the outpoint (`cur.Outpoint = vtxoKey.String()`) before HMAC-signing, so a cursor issued for one VTXO can't be replayed against another. `decodeChainCursor` enforces this at decode time. The unsigned/public-exposure path is correctly documented and acceptable since all chain data is already publicly readable in that mode.

**Session keepalive ordering is correct.**
`tokenCache.touch(sessionHash)` is called *after* `getVtxoChain` returns successfully, not before. This means a request that fails cursor decoding (bad `page_token`) cannot extend the session — exactly the invariant `TestGetVtxoChainInvalidPageTokenDoesNotExtendSession` checks for. Good.

**`validateChainAuth` continuation fallback is safe.**
When `validateAuthToken` fails and `isContinuation=true`, the code re-runs `verifyAuthTokenSignature` (signature-only check) plus `tokenCache.isActive`. The comment is right: the only way `validateAuthToken` fails while `verifyAuthTokenSignature` succeeds is an expired timestamp — malformed or bad-signature tokens fail both checks. The double-parse on error is harmless.

**Proto field numbers are fresh.**
`page_token` → field 5 on request, `next_page_token` → field 4 on response. No existing field reuse; backward compat is preserved. Existing `PageNum` callers get identical slices plus a `next_page_token` they can ignore.

**Redemption path preserved.**
`page == nil && pageToken == ""` → returns full chain, no cursor. The early return at that branch is correct.

**Test coverage is thorough.**
Round-trip, tamper, wrong-outpoint, no-HMAC-key, first-page-enforces-expiry, continuation-accepted-while-active, continuation-rejected-when-inactive, malformed-token-does-not-extend-session, full chain vs paged reassembly. All the important cases are covered.

---

### ⚠️ Minor issues / nits

**1. `json.Marshal` error silently discarded in `encodeChainCursor`:**
```go
payload, _ := json.Marshal(cur)
```
`vtxoChainCursor` only contains a `string` and `int`, so marshalling will never fail in practice. But discarding errors unconditionally is a pattern worth avoiding — if `cur` ever gains a field that can fail (e.g. a `json.Marshaler` impl), this will silently produce a wrong cursor. Consider:
```go
payload, err := json.Marshal(cur)
if err != nil {
    // unreachable in practice, but defensive
    return ""
}
```

**2. `validateChainAuth`: first-page error message on a malformed continuation token.**
When `isContinuation=true` and the token is malformed (not just expired), `validateAuthToken` returns e.g. `"invalid auth token format, must be base64"`. The code then calls `verifyAuthTokenSignature`, which returns the same or similar error. The caller ultimately gets the error from `verifyAuthTokenSignature` rather than the original. Not a bug — the right error is surfaced — but if you ever want to distinguish "expired during pagination" from "malformed token" in logs/metrics, you'd need to check the error type before branching. Not required for this PR.

**3. `paginateByOffset` with `pageSize=0` on an empty slice:**
When `pageSize <= 0` it's set to `len(items)`. If `items` is empty, `pageSize` stays 0. The `totalPages` block correctly guards `if pageSize > 0 { ... }` so no divide-by-zero, but `resp` will have `Current=1, Next=0, Total=0`. Functionally fine — no items returned, `more=false` — but the `Current=1` on an empty slice is slightly odd. Low priority.

**4. Cursor HMAC key derivation:**
```go
h := sha256.Sum256(append(privkey.Serialize(), []byte("cursor-hmac")...))
```
This is functional and the domain-separation string `"cursor-hmac"` is good. For the record, HKDF would be the standards-blessed approach here, but for an internal signing key derived from an existing private key this is fine.

---

### 🔒 Protocol-criticality assessment

This PR touches only the indexer RPC layer — `GetVtxoChain` pagination, auth session keepalive, and opaque cursor encoding. It does **not** touch VTXO transaction signing, forfeit paths, round lifecycle, or connector trees. **Not protocol-critical** by Arkana's definition — normal review and human approval is sufficient.

---

### Summary

The O(N²) caveat is correctly documented and bounded to this interim step. The design is clean, the security properties are well thought-out, and the test suite covers the important invariants. The three minor issues above are all low-risk; none block merging. **LGTM pending human approval.** 👍

/cc for merge sequencing: make sure #908 rebases on top of this before merge.

ghost (approved):
## Arkana Review — `9cb16c7c` (incremental)

Follow-up commit addresses all prior review feedback. Verified each fix:

1. **`indexer.go:555`** — `decodeChainCursor` errors now wrap `ErrInvalidInput`, propagating correctly through `fmt.Errorf("%w: invalid page_token: %w", ErrInvalidInput, err)`. ✅
2. **`handlers/indexer.go:357-359`** — `errors.Is(err, application.ErrInvalidInput)` maps to `codes.InvalidArgument` instead of `Internal`. Clean. ✅
3. **`indexer.go:1278-1283`** — `json.Marshal` error in `encodeChainCursor` is now handled. Returning `""` on the (unreachable) failure path silently signals "no next page" — acceptable since `{string, int}` marshaling cannot fail in practice. ✅
4. **Tests added:**
   - `TestGetVtxoChainForwardsPageToken` — positive wiring: `page_token` flows through handler to service, `next_page_token` propagates back. ✅
   - `TestGetVtxoChainMapsInvalidInputToInvalidArgument` — `ErrInvalidInput` → `InvalidArgument` status code. ✅
   - `TestGetVtxoChainInvalidPageTokenDoesNotExtendSession` — now asserts `ErrorIs(err, ErrInvalidInput)`. ✅
   - `TestValidateChainAuthPagination/"continuation rejects malformed token even with active session"` — covers the auth bypass vector. ✅

No remaining issues. LGTM.

🤖 Reviewed by Arkana

Dunsin-cyber (commented):
utACK

PR #1093, opened 2026-06-09
arkd-wallet: source UTXOs from NBXplorer events and cache wallet key paths

bitcoin-coder-bob:
## Summary

Two efficiency fixes to how `arkd-wallet` talks to NBXplorer. Both eliminate HTTP
calls for data NBXplorer already hands us in payloads we were discarding.

### 1. Build deposit UTXOs from the `newtransaction` event payload
The websocket notification handler previously called `searchNewUTXOs` on **every**
`newtransaction` event, which did `GET /v1/cryptos/BTC/groups/{groupID}/utxos`
(the entire group UTXO set) and filtered it client-side by txid. That work scales
linearly with the number of watched scripts and fires constantly.

NBXplorer 2.6.7's `newtransaction` event already contains the matched outputs. We
now parse them directly via a pure `utxosFromTransactionEvent(message, groupID)`,
scoped to `trackedSource == "GROUP:{groupID}"` (the identifier NBXplorer reports
for addresses added through the group addresses endpoint). No extra HTTP call, and
the cost no longer depends on group size. `event.Data` is now `json.RawMessage` to
drop a redundant re-marshal; `ports.Utxo` gained `KeyPath`, carried by `castUtxo`.

### 2. Cache `script -> keyPath` to skip per-input lookups when signing
`getPrivateKeyFromScript` used up to two `GetScriptPubKeyDetails` HTTP calls per
input. The `script -> {scheme, keyPath}` mapping is immutable, and UTXO listings
already return `keyPath` (we were dropping it). A **bounded LRU cache**
(50k entries) is populated from the wallet's account `GetUtxos` call sites and
consulted first; the HTTP path remains as a fallback that backfills the cache. A
cache miss is always safe (it just falls back), so eviction has zero correctness
risk and memory stays flat on a long-running server. Warm-cache signing makes zero
metadata calls.

## Benchmarks

`BenchmarkNotificationProcessing` — new event-parse is constant regardless of
group size; the old fetch-and-filter scales linearly:

| group size | old fetch+filter | new event-parse | speedup |
|-----------:|-----------------:|----------------:|--------:|
| 10         | 118 µs           | 8.2 µs          | ~14x    |
| 100        | 700 µs           | 8.2 µs          | ~85x    |
| 1,000      | 4.87 ms          | 8.2 µs          | ~590x   |
| 10,000     | 34.9 ms          | 8.2 µs          | ~4,250x |

(allocations at 10k: 30,141 -> 27; bytes/op: 13.9 MB -> 1.6 KB.) #2 is covered by
call-count tests: cache hit = 0 lookups, connector fallback = 2.

## Testing

- `go test -race ./core/...` — all subtests pass.
- Event parsing: group vs derivation-scheme vs newblock filtering, single/multi/
  empty outputs, non-zero confirmations, malformed JSON, invalid txid, keyPath
  passthrough.
- keyPath cache: set/get/miss/overwrite, bounded eviction (cap enforced,
  oldest-first), 100-goroutine concurrency; `getPrivateKeyFromScript`
  cache-hit/fallback/empty-keyPath behavior.
- `golangci-lint` (v2.9.0): no new issues vs `master`.

## Review feedback addressed

- **Unbounded cache growth** → cache is now a bounded LRU (50k cap); added a test
  asserting the bound and LRU eviction. (Note: the cache is populated only from
  the wallet's account UTXO listings, not from event notifications.)
- **Positional struct init** → `wallet` is now constructed with named fields.

## Notes

- Behavior is preserved for group deposits; events for the wallet's own derivation
  schemes are ignored exactly as before. Malformed events are now logged (benign).
- The cache key is the scriptPubKey, which maps to exactly one account/keyPath
  (the two accounts are distinct taproot xpubs), so a stale/wrong entry is not
  possible.
- Out of scope (from the wider NBXplorer usage audit): websocket gap-recovery via
  `/events?lastEventId`, single-RPC block-time, the `/balance` endpoint, and
  `reserve=true` for change addresses.

## Summary by CodeRabbit

* **Improvements**
  * Reduced repeated lookups during wallet UTXO selection by caching key-path information.
  * Improved transaction notification processing by reading UTXO and key-path details directly from websocket events.
  * Preserved key-path information to support reliable private-key derivation and transaction signing.
  * Added bounded, concurrency-safe cache behavior for consistent performance.

* **Tests**
  * Added coverage for key-path caching, wallet key derivation, concurrency, capacity limits, and notification processing.


ghost (changes requested):
## Code Review — arkd-wallet: source UTXOs from NBXplorer events and cache wallet key paths

**Reviewer**: Arkana (automated, aggressive)
**Verdict**: Request changes (1 actionable issue) + request human sign-off (signing path modified)

---

### Summary

Clean, well-tested PR. The two optimizations are sound:
1. Parsing UTXO data directly from the `newtransaction` event payload instead of re-fetching the entire group UTXO set — eliminates an O(n) HTTP call per event.
2. Caching `script → (derivationScheme, keyPath)` to skip per-input NBXplorer lookups during signing.

Test coverage is thorough (event parsing, cache hit/miss/fallback/concurrency, edge cases). The `json.RawMessage` change for `event.Data` and the `Confirmations int → uint32` alignment are correct improvements. The behavioral fix in `getPrivateKeyFromScript` (skipping schemes that return an empty `KeyPath` instead of trying to derive from `""`) is a genuine bug fix.

No cross-repo breakage: `ports.Utxo` is only consumed within `pkg/arkd-wallet/`, and the new `KeyPath` field is added with named-field construction everywhere.

---

### 🔴 ACTIONABLE — Unbounded cache growth

**`pkg/arkd-wallet/core/application/wallet/keypath_cache.go`**

The `keyPathCache` map grows without bound. Every UTXO the wallet ever sees (via `ListConnectorUtxos`, `SelectUtxos`, `withdrawAll`, `getBalance`, and event notifications) adds an entry that is never evicted.

For a long-running `arkd` server managing many watched addresses (e.g. thousands of boarding UTXOs per round across many rounds), this is a slow memory leak. The comment says "cached entries never need invalidation" — true for correctness, but irrelevant for memory management. Scripts from spent UTXOs will never be looked up again, yet their entries persist forever.

**Options (pick one):**
- **LRU with a reasonable cap** (e.g. 10k–50k entries). A signing operation only needs the scripts of the inputs being signed, so eviction is always safe — the NBXplorer fallback path still works.
- **Periodic sweep**: clear the cache when the wallet detects a new round/block (coarse but simple).
- **Explicit eviction after signing**: remove entries for scripts that were just signed (they're spent now).

The first option is simplest and safest. An LRU miss just falls back to the existing HTTP path, so there's zero correctness risk.

---

### 🟡 STYLE — Positional struct init is fragile

**`pkg/arkd-wallet/core/application/wallet/service.go:65`**

```go
return &wallet{opts, newOutpointLocker(time.Minute), nil, newKeyPathCache(), make(chan bool)}
```

This uses positional field initialization. If anyone reorders the `wallet` struct fields, this silently breaks at compile time only if types mismatch — otherwise it's a silent bug. Prefer named fields:

```go
return &wallet{
    WalletOptions: opts,
    locker:        newOutpointLocker(time.Minute),
    keyPaths:      newKeyPathCache(),
    readyCh:       make(chan bool),
}
```

Not blocking, but worth fixing while you're here.

---

### 🟢 LOOKS GOOD

- **`utxosFromTransactionEvent`** — pure function, well-scoped, correct `TrackedSource` filtering, proper error propagation. The `make([]ports.Utxo, 0, len(newTxEvent.Outputs))` pre-allocation is a nice touch.
- **`event.Data` → `json.RawMessage`** — correct elimination of the redundant marshal→unmarshal roundtrip.
- **`matchedOutput` type** — clean mapping of the NBXplorer event payload. Fields match the NBXplorer 2.6.7 schema.
- **`castUtxo` now carries `KeyPath`** — correct one-line addition.
- **`getPrivateKeyFromScript` empty-KeyPath guard** — this is a bug fix: the old code would pass `""` to `deriveKey` on the first scheme match with empty KeyPath, likely erroring out and halting the loop before trying the connector scheme. Now it correctly falls through.
- **Test coverage** — event parsing (group vs derivation-scheme vs newblock, single/multi/empty outputs, confirmations, malformed JSON, invalid txid, keyPath passthrough), cache (set/get/miss/overwrite/concurrent), signing path (locked wallet, cache hit, cache miss + backfill, connector fallback, empty KeyPath fallback, unknown script). Benchmark is well-structured.

---

### ⚠️ PROTOCOL-ADJACENT — Human review required

This PR modifies `getPrivateKeyFromScript`, which is on the signing path. The change is a caching layer — the underlying `deriveKey` call and its parameters are unchanged — so the risk of fund loss is minimal. However, per protocol-critical review rules:

**A human must review and approve this PR before merge.** Specifically, verify:
1. The cache can never return a stale/wrong `(derivationScheme, keyPath)` for a given script (it can't, because the mapping is immutable by derivation, but a human should confirm this assumption holds for all address types the wallet manages).
2. The empty-KeyPath guard in the fallback path is correct behavior (it is — NBXplorer returns empty KeyPath for scripts not derived from the queried scheme).

ghost (approved):
## Follow-up Review — new commit `923414a`

**Reviewer**: Arkana (automated, aggressive)
**Verdict**: ✅ Approve (Arkana side)

---

Both issues from my [previous review](https://github.com/arkade-os/arkd/pull/1093#pullrequestreview-4462485334) are resolved:

### 🔴 Unbounded cache → Fixed
`sync.RWMutex` + `map[string]keyPathEntry` replaced with `hashicorp/golang-lru/v2` `lru.Cache` capped at 50,000 entries. The library is concurrency-safe internally, so the manual mutex removal is correct. Eviction of old entries is safe — a cache miss falls back to the NBXplorer HTTP path with zero correctness risk.

New `TestKeyPathCache_Bounded` test verifies the cap holds and oldest entries are evicted. Good.

### 🟡 Positional struct init → Fixed
`wallet{}` in `New()` now uses named fields. Clean.

No new issues in the incremental diff. The `go.mod` addition of `hashicorp/golang-lru/v2 v2.0.7` is appropriate.

---

### ⚠️ Human review still required

This PR modifies `getPrivateKeyFromScript` (signing path). Per protocol-critical rules, a **human must also approve** before merge. The two things to verify remain unchanged from my first review:
1. The `script → (derivationScheme, keyPath)` mapping is truly immutable for all address types the wallet manages.
2. The empty-KeyPath guard correctly skips schemes that don't own the script.

ghost (approved):
## Follow-up Review — new commit `923414a`

**Reviewer**: Arkana (automated, aggressive)
**Verdict**: ✅ Approve (Arkana side)

---

Both issues from my [previous review](https://github.com/arkade-os/arkd/pull/1093#pullrequestreview-4462485334) are resolved:

### 🔴 Unbounded cache → Fixed
`sync.RWMutex` + `map[string]keyPathEntry` replaced with `hashicorp/golang-lru/v2` `lru.Cache` capped at 50,000 entries. The library is concurrency-safe internally, so the manual mutex removal is correct. Eviction of old entries is safe — a cache miss falls back to the NBXplorer HTTP path with zero correctness risk.

New `TestKeyPathCache_Bounded` test verifies the cap holds and oldest entries are evicted. Good.

### 🟡 Positional struct init → Fixed
`wallet{}` in `New()` now uses named fields. Clean.

No new issues in the incremental diff. The `go.mod` addition of `hashicorp/golang-lru/v2 v2.0.7` is appropriate.

---

### ⚠️ Human review still required

This PR modifies `getPrivateKeyFromScript` (signing path). Per protocol-critical rules, a **human must also approve** before merge. The two things to verify remain unchanged from my first review:
1. The `script → (derivationScheme, keyPath)` mapping is truly immutable for all address types the wallet manages.
2. The empty-KeyPath guard correctly skips schemes that don't own the script.

PR #1096, opened 2026-06-11
make version interceptor configurable

louisinger:
`BUILD_VERSION_GUARD_LEVEL` can be set to "major|minor|patch" and determine on which level the interceptor should compare client and server versions. default value is "minor".

`BUILD_VERSION_REQUIRE_HEADER` is a boolean. if set, it forces clients to set the X-Build-Version header. default to false.

@altafan please review

## Summary by CodeRabbit

* **New Features**
  * Require/accept client build-version headers with configurable enforcement
  * Improved semantic version compatibility checks (major.minor.patch) with admin/health endpoints exempt
  * New CLI settings command to view/update partial settings; scheduled-session flags renamed (session-duration)
  * Settings API and admin spec expose build-version fields and use RFC3339 timestamps for session times

* **Bug Fixes**
  * Validation prevents enabling "required" without a configured minimum version


ghost (commented):
## Code Review — `make version interceptor configurable`

**Scope:** Version compat gRPC interceptor. Not protocol-critical (no VTXO/signing/forfeit/round changes). No cross-repo impact — `internal` package with zero external consumers.

**Verdict:** Clean refactor with good test coverage. One performance issue worth fixing before merge.

---

### 🔴 Performance regression: server version parsed on every RPC

**`internal/interface/grpc/interceptors/version.go:120`**

The old code parsed the server version **once** at interceptor construction:
```go
// OLD (interceptor.go)
major, minor, _ := parseVersion(serverVersion)
// then closed over major/minor in the handler
```

The new code re-parses on **every call**:
```go
// NEW (version.go:120)
func checkVersionCompat(ctx context.Context, guard VersionGuard) error {
    serverMajor, serverMinor, serverPatch, err := parseVersion(guard.ServerVersion)
    // ^^^ runs on every single RPC
```

`parseVersion` does `TrimPrefix` + `Split` + 3x `ParseInt` — allocations on every request. On a busy server this is unnecessary GC pressure.

**Fix:** Parse once and store in `VersionGuard`:
```go
type VersionGuard struct {
    ServerVersion string
    RequireHeader bool
    Level         config.VersionGuardLevel
    // pre-parsed
    serverMajor, serverMinor, serverPatch int64
    serverParsed                          bool
}

func NewVersionGuard(version string, require bool, level config.VersionGuardLevel) VersionGuard {
    major, minor, patch, err := parseVersion(version)
    return VersionGuard{
        ServerVersion: version,
        RequireHeader: require,
        Level:         level,
        serverMajor:   major,
        serverMinor:   minor,
        serverPatch:   patch,
        serverParsed:  err == nil,
    }
}
```

Then `checkVersionCompat` just reads pre-computed fields. No per-request allocations.

---

### 🟡 Nit: redundant validation map

**`internal/config/config.go:67-71`**

`supportedBuildVersionGuardLevels` is defined but only used in the error message of `parseVersionGuardLevel`. The actual validation is the `switch` statement. The map is slightly redundant — consider using `strings.Join` on the const values instead, or using the map *for* the validation (lookup instead of switch).

---

### ✅ Things that look good

- **`isBehind` logic (version.go:60-72):** Correct hierarchical comparison. Falls through major→minor→patch correctly at each guard level.
- **`parseLeadingInt` (version.go:42-50):** Handles pre-release suffixes (`1.2.3-rc1`) cleanly.
- **Test coverage (version_test.go):** 20+ cases covering all three levels, require/optional header, unparseable versions, edge cases. Unified test structure for unary+stream is a nice improvement.
- **Config plumbing (config.go, service.go):** Standard viper pattern, defaults are sensible (`minor` level, header not required — backward compatible).
- **Backward compatibility:** Default behavior (level=minor, header optional) matches the old behavior exactly.

---

**Summary:** Approve after the per-RPC parse is fixed. The rest is clean.

🤖 reviewed by arkana

ghost (commented):
## Arkana Code Review — `version-header` (arkade-os/arkd#1096)

Clean refactor overall. The `VersionGuard` struct, configurable guard level, and `RequireHeader` flag are well-designed. Tests are thorough — all three levels, both interceptor types, header-present/absent/malformed cases covered. No protocol-critical concerns (this doesn't touch VTXOs, signing, forfeit, rounds, or exits).

### 🔴 Performance regression — server version re-parsed on every request

**`internal/interface/grpc/interceptors/version.go` — `checkVersionCompat()`**

The old code parsed the server version **once** at interceptor creation time (in `UnaryInterceptor`/`StreamInterceptor`) and captured the pre-parsed `major`/`minor` in the closure. The new code calls `parseVersion(guard.ServerVersion)` inside `checkVersionCompat()`, which runs on **every single gRPC call**.

```go
func checkVersionCompat(ctx context.Context, guard VersionGuard) error {
    serverMajor, serverMinor, serverPatch, err := parseVersion(guard.ServerVersion)  // ← hot path
```

While `parseVersion` is cheap, this is unnecessary work on every request for a value that never changes. **Fix:** pre-parse the server version components and store them in the `VersionGuard` struct (e.g. `ServerMajor`, `ServerMinor`, `ServerPatch` fields), computed once at construction time in `service.go:newServer()`.

### 🟡 Error message vs metadata inconsistency

**`internal/interface/grpc/interceptors/version.go` — `buildVersionTooOld()`**

The human-readable error says `"server requires build version header >= %s", serverVersion` (e.g. `>= 2.3.4`), but the `min_version` metadata field is computed from the guard level and could be `"2.3.0"` (minor) or `"2.0.0"` (major). An SDK parsing the error message gets a stricter requirement than the actual policy. Consider using `minAllowedVersion` in the message too:

```go
return errors.BUILD_VERSION_TOO_OLD.
    New("server requires build version header >= %s", minAllowedVersion).  // ← use minAllowedVersion
    ...
```

### 🟡 Case-sensitive guard level parsing

**`internal/config/version.go:14` — `parseVersionGuardLevel()`**

The switch matches exact lowercase only. `BUILD_VERSION_GUARD_LEVEL=Minor` or `MINOR` would fail with a hard startup error. Standard practice for env vars is case-insensitive parsing:

```go
func parseVersionGuardLevel(s string) (VersionGuardLevel, error) {
    switch strings.ToLower(s) {
    ...
```

### ℹ️ SDK compatibility note

None of the four SDKs (go-sdk, ts-sdk, rust-sdk, dotnet-sdk) currently send the `x-build-version` header. Enabling `BUILD_VERSION_REQUIRE_HEADER=true` in production would immediately reject **all** current SDK clients. This should be called out in the PR description or config docs so operators don't flip this flag without coordinating SDK upgrades.

---

**Verdict:** No blocking issues. The performance regression is a clean-up item, not a correctness bug. The error message inconsistency is worth fixing while you're in here. LGTM with the nits addressed.

ghost (approved):
## Arkana Code Review — `version-header` update (arkade-os/arkd#1096)

Reviewing only commit `4f4bd5df` ("fix after reviews"), pushed after my previous two reviews.

**Verdict: LGTM ✅**

### What changed

1. **Pre-parsed version in `NewVersionGuard` constructor** — Server version parsed once at startup, `minAllowedVersion` pre-computed. Eliminates per-request parsing overhead. Addresses the performance concern from my earlier review.

2. **Admin service exemption** (`version.go:24-30`) — `versionGuardSkippedServices` correctly exempts `AdminService`, `WalletService`, `WalletInitializerService`, `SignerManagerService`, and `grpc.health.v1.Health`. These are operator/infra services, not SDK-driven — correct to skip.

3. **`skipVersionGuard` parsing** (`version.go:35-41`) — Correctly extracts service name from gRPC `fullMethod` format (`/package.Service/Method`). Map lookup is O(1). Clean.

4. **Error message fix** (`version.go:161`) — Now reports `minAllowedVersion` (e.g., "2.3.0") instead of raw `serverVersion` ("2.3.4") in the error body. This is more accurate — tells clients the actual minimum they need, not the exact server version.

5. **Unexported parsed fields** — `enabled`, `serverMajor/Minor/Patch`, `minAllowedVersion` are private. `NewVersionGuard` is the only constructor path. Prevents bypassing the parse step.

6. **Test coverage** — `TestNewVersionGuardMinAllowedVersion` validates pre-computation for all levels including pre-release suffix. `TestVersionCompatSkipsAdminEndpoints` covers all 5 skipped services with both missing and outdated headers, plus a sanity check that `ArkService` is still guarded.

### Not protocol-critical
No VTXO, signing, forfeit, round, or exit path changes. This is operational version-gating infrastructure.

🤖 reviewed by arkana

ghost (approved):
## Arkana Code Review — `version-header` commit `a0dca5f1` (arkade-os/arkd#1096)

Reviewing only the new commit pushed after my last approval.

**Verdict: LGTM ✅**

### What changed

This commit simplifies the design by removing the `VersionGuardLevel` concept entirely:

1. **Deleted `internal/config/version.go`** — `VersionGuardLevel` type and `parseVersionGuardLevel()` removed. No more major/minor/patch guard level selection. Good call — operators can just set the exact minimum version they want via the env var.

2. **Renamed config to be more explicit** — `BUILD_VERSION_REQUIRE_HEADER` / `BUILD_VERSION_GUARD_LEVEL` → `MIN_BUILD_VERSION_HEADER` / `MIN_BUILD_VERSION_HEADER_REQUIRED`. The new names better describe the semantics: "what is the minimum version?" rather than "guard at what granularity?"

3. **Decoupled min version from server version** (`service.go:297`) — `s.version` → `s.appConfig.BuildVersionHeader`. The minimum accepted client version is now an explicit operator-configured value, not automatically derived from the server's own build version. This is the right design — allows operators to set a minimum lower than the current server version during rolling upgrades.

4. **`isBehind` always compares full semver** — No more early returns based on guard level. Simpler, correct.

5. **Config validation** (`config.go:693`) — `BuildVersionHeaderRequired && len(BuildVersionHeader) <= 0` prevents operators from requiring the header without setting a minimum. Good guard.

6. **Log levels → Warn** — Rejected requests now log at Warn instead of Debug. Correct — operators need visibility into rejected clients.

### Test coverage

Tests updated to remove all `config.VersionGuardLevel` references. `TestNewVersionGuardMinAllowedVersion` removed (guard level tests no longer applicable). The remaining `TestVersionCompat` and `TestVersionCompatSkipsAdminEndpoints` suites cover the simplified behavior thoroughly.

### Not protocol-critical
No VTXO, signing, forfeit, round, or exit path changes.

🤖 reviewed by arkana

PR #1097, opened 2026-06-11
Implement deprecated signer keys

### LINKED ISSUE #822: Support vtxo migration to new signers
When a new signer key becomes active, while we have deprecated signers still available,  allow vtxos to be inputs even if the taptree reveals the signer is the old one, ONLY if the outputs in the tx/intent also reveal the taptree, and that the signer in the taptree is the active signer.

louisinger:
it closes https://github.com/arkade-os/arkd/issues/822

config example : 

```
  ARKD_WALLET_SIGNER_KEY=1111111111111111111111111111111111111111111111111111111111111111
  ARKD_WALLET_DEPRECATED_SIGNER_KEYS=afcc3fa10f82a05fddc9574fdb13b3991b568e89cc39a72ba4401df8abef3
  5f0:1782864000,2222222222222222222222222222222222222222222222222222222222222222:1782864000
```

means : 
* `1111111111111111111111111111111111111111111111111111111111111111` is the current signer key
* `afcc3fa10f82a05fddc9574fdb13b3991b568e89cc39a72ba4401df8abef3` and `2222222222222222222222222222222222222222222222222222222222222222` are accepted until unix timestamp `1782864000`

@altafan please review

## Summary by CodeRabbit

* **New Features**
  * Signer key rotation: keys can be marked deprecated with cutoff timestamps and remain accepted until expiration.
  * APIs/reporting now include deprecated signer keys and cutoff dates; services, wallet signing, and proof finalization accept multiple signer keys for validation and signing.
  * Configuration and docker setup support supplying deprecated signer keys.

* **Tests**
  * New end-to-end and unit tests validate deprecated-key reporting, acceptance windows, signing selection, and settlement behavior.


ghost (changes requested):
## 🔴 Protocol-Critical Review — Deprecated Signer Keys

This PR touches **transaction signing, VTXO validation, forfeit paths, and intent proofs** — all protocol-critical code. Requesting human sign-off regardless of code correctness.

Overall the design is sound: allow key rotation by keeping deprecated signer keys accepted until a cutoff date. The plumbing is consistent across the service, indexer, wallet, and tx-builder layers. But there are several issues that need to be addressed before merge.

---

### 🔴 P0 — SECURITY: `signerKeyForLeaf` uses `bytes.Contains` for key matching

**`pkg/arkd-wallet/core/application/wallet/service.go` — new `signerKeyForLeaf` function**

```go
func (w *wallet) signerKeyForLeaf(leafScript []byte) *btcec.PrivateKey {
    for _, k := range w.DeprecatedSignerKeys {
        if bytes.Contains(leafScript, schnorr.SerializePubKey(k.Key.PubKey())) {
            return k.Key
        }
    }
    return w.SignerKey
}
```

Using `bytes.Contains` to match a 32-byte x-only pubkey against raw script bytes is **fragile and potentially exploitable**. A 32-byte sequence can appear as a substring across adjacent push-data boundaries or within hash preimages embedded in the script. If a crafted script happens to contain the deprecated key's bytes in a non-key position, the wallet would select the wrong signing key, producing an **invalid signature** — or worse, signing with a key the attacker intended.

**Fix:** Parse the leaf script using the closure types (e.g., `MultisigClosure.Decode()`) to extract the actual pubkey operands, then match against those. The script package already has decoders for every closure type.

---

### 🟡 P1 — PERFORMANCE: gRPC call per `verifyTapscriptPartialSigs` invocation

**`internal/infrastructure/tx-builder/covenantless/builder.go` — lines in `verifyTapscriptPartialSigs`**

```go
deprecatedSignerPubkeys, err := b.signer.GetDeprecatedPubkeys(context.Background())
```

`GetDeprecatedPubkeys` is a **gRPC call** to the wallet service. `verifyTapscriptPartialSigs` runs on every transaction verification during round processing. Under load (many participants per round), this multiplies into potentially hundreds of gRPC calls per round for data that changes only when the operator restarts with a new config.

**Fix:** Fetch deprecated pubkeys once at `txBuilder` initialization (or cache with the signer pubkey that's already fetched once). Pass them as a field on the builder, not fetched per-call.

---

### 🟡 P1 — PROTOCOL: Race condition — cutoff expiry mid-round

**`internal/core/application/service.go` and `internal/core/application/utils.go`**

`acceptedSignerPubkeys()` calls `time.Now()` on every invocation. During a single round, multiple validation steps use this function:
- `validateVtxoScript` (line ~760)
- `verifyIntentProofAndFindMatches` (line ~4364)  
- `RegisterIntent` → `intent.Verify` (line ~1735)
- `FinalizeAndExtract` (line ~1761)

If a deprecated key's cutoff expires **between** these steps within the same round, a VTXO could pass early validation but fail later verification (or vice versa), causing partial processing failures and a potentially stuck round.

**Fix:** Snapshot `acceptedSignerPubkeys()` once at round start and pass it through, or compute it once per request context.

---

### 🟡 P1 — PROTOCOL: `FinalizeAndExtract` weight overestimation with multiple signers

**`pkg/ark-lib/intent/proof.go:FinalizeAndExtract`**

```go
for _, signer := range signers {
    fakeSigs = append(fakeSigs, &psbt.TaprootScriptSpendSig{...})
}
for i := range p.Inputs {
    ptx.Inputs[i].TaprootScriptSpendSig = append(ptx.Inputs[i].TaprootScriptSpendSig, fakeSigs...)
```

A real VTXO is locked to exactly **one** signer pubkey. Adding fake 64-byte sigs for **all** accepted signers inflates the finalized tx weight estimate by `64 * (N-1) * numInputs` bytes. This overestimate affects fee calculations.

**Fix:** Only add a single fake sig. The proof VTXO is locked to one signer — either detect which one from the script, or accept the small overestimate and add a code comment explaining it's bounded.

---

### 🟡 P2 — No duplicate key validation

**`pkg/arkd-wallet/config/config.go` — `parseDeprecatedSignerKeys`**

There's no check that a deprecated key differs from the current `SIGNER_KEY`. If misconfigured, the same key appears twice in `acceptedSignerPubkeys()`, and `signerKeyForLeaf` could match the deprecated entry first, bypassing the current key code path. Not a security hole today but a footgun.

**Suggestion:** Warn or error at startup if a deprecated key matches the current signer key.

---

### 🟢 P3 — Minor: Cutoff boundary semantics

**`internal/core/application/utils.go`**

```go
if key.CutoffDate > 0 && now.Unix() > key.CutoffDate {
    continue
}
```

At `now.Unix() == key.CutoffDate`, the key is still accepted. The proto comment says "after which the key is no longer accepted" — this is consistent. But the unit test (`TestAcceptedSignerPubkeys`) doesn't cover the boundary case where `now` equals the cutoff exactly. Consider adding one.

---

### ✅ What looks good

- The `acceptedSignerPubkeys` utility with time injection is well-structured and testable
- `allSignerPubkeys` (ignoring cutoff) for `stripSignerSignatures` is correct — you need to strip old sigs even past cutoff
- E2E test covers the full lifecycle: create VTXO with old key → rotate → settle → offchain tx → expired cutoff rejection
- Proto changes are additive (new field 17 on GetInfoResponse, new field 2 on GetPubkeyResponse) — no breaking wire changes
- The `validateVtxoScriptForSigners` loop-over-signers approach is correct for VTXO validation
- `signerKeyForLeaf` concept is right (select the key the VTXO was locked to) — just the implementation needs hardening
- Cross-repo: `signer_pubkey` preserved in GetInfoResponse, SDKs won't break

---

### ⚠️ Cross-repo note

The `DeprecatedSigner` message was added to both `signer/v1/service.proto` and `ark/v1/types.proto`. SDK consumers (go-sdk, ts-sdk, rust-sdk, dotnet-sdk, arkade-kotlin) will need proto regeneration to see the new `deprecated_signers` field on `GetInfoResponse`. This is additive and non-breaking, but SDKs should be updated to expose it.

---

**Verdict:** Request changes on P0 (signerKeyForLeaf). P1 items are strong recommendations. **⚠️ PROTOCOL-CRITICAL: Requires human sign-off before merge regardless.**

ghost (changes requested):
## Follow-up review — new commits `90d94cb` + `61655b7`

Reviewing only what changed since my previous review (June 11).

---

### Change summary

All `intent.Verify` and `FinalizeAndExtract` calls switched from `acceptedSignerPubkeys()` → `allSignerPubkeys()`. The `acceptedSignerPubkeys()` method was removed from `indexerService` and added `allSignerPubkeys()` to both `service` and `indexerService`. The intent: VTXOs locked to a deprecated key whose cutoff has passed should still be **spendable/recoverable** — the cutoff only prevents **new** VTXOs from being created with that key.

### ✅ Correct

- **Design is sound.** Using `allSignerPubkeys()` for intent verification while keeping `acceptedSignerPubkeys()` for `validateVtxoScript` is the right split:
  - `validateVtxoScript` → gates **new** VTXOs entering the system (boarding, rounds) → should reject expired keys ✅
  - `intent.Verify` → proves **existing** VTXO ownership → should accept any key the VTXO was ever locked to ✅
  - `FinalizeAndExtract` → reconstructs proof tx for existing VTXOs → must match the actual signer key used ✅
- This also **mitigates my previous P1 (race condition mid-round)** for the intent verification path — intent proofs now use the stable `allSignerPubkeys()` set, so cutoff expiry mid-round can't break intent verification. Good.
- Comment updates in `indexer.go:97-99` accurately describe the rationale.

---

### 🟡 Remaining concern: `FinalizeAndExtract` weight overestimation is now worse

**`service.go:1794`**: `FinalizeAndExtract(s.allSignerPubkeys()...)` — now includes **all** deprecated signers including expired ones.

Each extra signer adds a 64-byte fake script-spend sig per input. With N deprecated keys and M inputs, the overestimate is `64 * N * M` bytes. This was my P1 from the previous review and it's slightly worse now since expired keys are included.

Pragmatically: if you only ever have 1-2 deprecated keys, this is ~128 extra bytes per input — tolerable. But **please add a code comment** at `service.go:1794` explaining that this overestimates and why it's bounded, so a future reader doesn't wonder.

---

### ✅ Previous findings status

| Finding | Status |
|---------|--------|
| P0: `signerKeyForLeaf` bytes.Contains | **Still open** — not addressed in these commits |
| P1: gRPC per `verifyTapscriptPartialSigs` | **Still open** |
| P1: Race condition mid-round | **Partially mitigated** — intent path fixed, but `validateVtxoScript` in `SubmitOffchainTx` (line 773) still uses `acceptedSignerPubkeys()` with `time.Now()` |
| P1: FinalizeAndExtract weight | **Still open**, slightly worse |
| P2: No duplicate key validation | **Still open** |
| P3: Cutoff boundary test | **Still open** |

**The new commits are correct and improve the recovery story. But P0 (`signerKeyForLeaf` bytes.Contains) remains a blocking issue.** Keeping request-changes until that's addressed.

⚠️ **PROTOCOL-CRITICAL: Requires human sign-off before merge.**

ghost (changes requested):
## Follow-up review — commit `80a2164` ("reject empty deprecated key in config")

### ✅ New commit is correct

The added guard in `pkg/arkd-wallet/config/config.go:207-209` correctly rejects entries like `:1782864000` (missing hex key with only cutoff). Good edge case fix.

One minor nit: `keyPart` is already the result of `strings.Cut` on a `TrimSpace`'d entry, so the inner `strings.TrimSpace(keyPart)` on line 207 is redundant when the separator is `:` and the entry was already trimmed. Not blocking — just noting for cleanliness.

---

### Previous findings — still open

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | 🔴 P0 | `signerKeyForLeaf` uses `bytes.Contains` for key matching (`service.go:655`) — fragile, potentially exploitable | **Still open** |
| 2 | 🟡 P1 | gRPC call per `verifyTapscriptPartialSigs` invocation in tx-builder | **Still open** |
| 3 | 🟡 P1 | `FinalizeAndExtract` weight overestimation with all signers | **Still open** |
| 4 | 🟢 P2 | No duplicate key validation in config | **Still open** |
| 5 | 🟢 P3 | Cutoff boundary test (`now == cutoff`) | **Still open** |

**P0 remains the blocking issue.** `bytes.Contains` on raw script bytes to match a 32-byte pubkey is not safe — parse the leaf script via closure decoders to extract the actual key operand.

⚠️ **PROTOCOL-CRITICAL: Requires human sign-off before merge.**

internal/core/application/utils_test.go
```diff
@@ -164,3 +170,100 @@ func mustEncodePSBTB64(t *testing.T, tx *wire.MsgTx) string {
...
+		require.NoError(t, err)
+		other := otherKey.PubKey()
+
+		pubkeys := acceptedSignerPubkeys(current, []ports.DeprecatedSignerPubkey{
+			{PubKey: deprecated, CutoffDate: now.Add(-time.Hour)},
+			{PubKey: other, CutoffDate: now.Add(time.Hour)},
+		}, now)
+		require.Equal(t, []*btcec.PublicKey{current, other}, pubkeys)
+	})
+}
```
aruokhai, line 228: Maybe a vector list would be better, and a single pass might make the test slimmer and more concise

internal/core/ports/live_store.go
```diff
@@ -127,17 +127,27 @@ type ConfirmationSessions struct {
...
+	for _, deprecated := range s.DeprecatedSignerPubkeys {
+		deprecatedSigners = append(deprecatedSigners, deprecatedSignerDigestData{
+			PubKey:     hex.EncodeToString(deprecated.PubKey.SerializeCompressed()),
+			CutoffDate: deprecated.CutoffDate.Unix(),
+		})
+	}
+
 	data := digestData{
 		SignerPubKey:        hex.EncodeToString(s.SignerPubkey.SerializeCompressed()),
+		DeprecatedSigners:   deprecatedSigners,
```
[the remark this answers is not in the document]
aruokhai: @louisinger  I aggree, Lets normalise this data

pkg/arkd-wallet/config/config.go
```diff
@@ -181,3 +191,39 @@ func getNetwork() (arklib.Network, error) {
...
+					"invalid cutoff date, must be a positive unix timestamp: %s", entry,
+				)
+			}
+			cutoffDate = cutoff
+		}
+
+		buf, err := hex.DecodeString(keyPart)
+		if err != nil {
+			return nil, fmt.Errorf("invalid signer key format, must be hex: %s", keyPart)
+		}
```
aruokhai: Lets fail fast, after verifying that keyPart is not empty, we should apply this logic

internal/core/ports/live_store.go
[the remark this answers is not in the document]
louisinger: done

pkg/arkd-wallet/config/config.go
louisinger: done

internal/test/e2e/e2e_test.go
```diff
@@ -6231,3 +6231,168 @@ func TestEventListenerChurn(t *testing.T) {
...
+		newSignerKey = "1111111111111111111111111111111111111111111111111111111111111111"
+		sendAmount   = 10000
+		// unix timestamp after which the deprecated key is no longer accepted
+		cutoffDate = int64(33256915200) // 3023-11-04
+	)
+	ctx := t.Context()
+
+	t.Cleanup(func() {
+		require.NoError(t, recreateArkdWallet(oldSignerKey, ""))
+	})
```
altafan, line 6256:
```suggestion
    // Restore the old signer key without deprecated keys for other integration tests
	t.Cleanup(func() {
		require.NoError(t, recreateArkdWallet(oldSignerKey, ""))
	})
```

internal/test/e2e/e2e_test.go
```diff
@@ -6231,3 +6231,168 @@ func TestEventListenerChurn(t *testing.T) {
...
+			Tapscripts: bobOffchainAddr.Tapscripts,
+		}}))
+		require.ErrorContains(t, err, "is a deprecated key since")
+
+		// boarding input path: dave's boarding utxo is locked to the old key
+		faucetOnchain(t, daveBoardingAddr.Address, 0.00021)
+		time.Sleep(6 * time.Second)
+
+		_, err = dave.Settle(ctx)
+		require.ErrorContains(t, err, "is a deprecated key since")
```
altafan, line 6407: let's add the last leg, ie let's make bob's funds expire by mining the required num of blocks and then ensure he's able to settle (no matter to what address)

pkg/arkd-wallet/core/application/wallet/service.go
```diff
@@ -759,6 +779,40 @@ func (w *wallet) SignTransaction(
...
+	leafKeys := make([]*btcec.PublicKey, 0)
+	switch c := closure.(type) {
+	case *script.MultisigClosure:
+		leafKeys = c.PubKeys
+	case *script.CLTVMultisigClosure:
+		leafKeys = c.PubKeys
+	case *script.ConditionMultisigClosure:
+		leafKeys = c.PubKeys
+	default:
+		return nil
```
altafan:
```suggestion
		return w.SignerKey
```

Kukks (commented):
Reviewed the deprecated-signer-key mechanic end to end. Overall solid and well-tested; the model implemented here is deprecated keys as co-equal signers until cutoff (simpler than the migration-only mechanic in #822), which is a deliberate, accepted trade-off. A few actionable items below — a follow-up PR against this branch addresses the concrete ones (nil-deref in signerKeyForLeaf, wallet-side cutoff defense-in-depth, deprecated privkey length validation, and a recovery-after-cutoff e2e test). The broad output-taptree-reveal enforcement is left for a later PR.

pkg/arkd-wallet/core/application/wallet/service.go
Kukks: `signerKeyForLeaf` returns `nil` here for `CSVMultisigClosure` and `ConditionCSVMultisigClosure` (the two ExitClosure types). That contradicts the doc comment ("or the current SignerKey") and the pre-PR behaviour, which always returned `w.SignerKey`. With any deprecated key configured, this returns `nil` even when the leaf holds the *current* signer key, and the caller dereferences it unconditionally under `SignModeSigner` — `RawTxInTapscriptSignature(..., nil)` and `signingKey.PubKey()` — i.e. a nil-pointer panic. It's latent today (the signer only co-signs the 3 forfeit closures; the sweep path uses `SignModeLiquidityProvider`, which overrides the key), but it's a crash waiting for any future signer-mode CSV signing. All five closures embed `MultisigClosure`, so handling them all and defaulting to `w.SignerKey` is trivial. Addressed in the follow-up PR.

pkg/arkd-wallet/core/application/wallet/signer_keys_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/require"
+)
+
+// TestSignTransaction makes sure SignTransaction signs taproot script-path inputs
+// with the key referenced by the leaf, including deprecated signer keys. From the
+// wallet's point of view the cutoff date is purely informational: it always signs
```
Kukks, line 23: Defense-in-depth: as this comment documents, the wallet signs with a deprecated key regardless of cutoff — cutoff is enforced solely at arkd's input validation. That is fine while arkd is the only gate, but the signer has no last line of defence for a compromised key whose cutoff was set to the past. #822 explicitly recommended the signer refuse to sign past cutoff. The follow-up PR adds that refusal. Recovery is unaffected: swept/expired VTXOs don't require a forfeit, so the deprecated key is never asked to sign them.

internal/core/application/utils.go
```diff
@@ -161,8 +161,55 @@ func decodeTx(offchainTx domain.OffchainTx) (string, []domain.Outpoint, []domain
...
+	return pubkeys
+}
+
+func isPastCutoff(key ports.DeprecatedSignerPubkey, now time.Time) bool {
+	return !key.CutoffDate.IsZero() && now.After(key.CutoffDate)
+}
+
+// validateVtxoScriptForSigners accepts the script if it validates against the current
+// signer pubkey or any deprecated one whose cutoff date has not passed yet.
+func validateVtxoScriptForSigners(
```
Kukks, line 186: Design note (agreed acceptable, recording it for posterity): this validates a script against the current key or any non-expired deprecated key, i.e. deprecated keys are co-equal signers until cutoff rather than the migration-only mechanic in #822. There is no output-commitment check, so nothing stops a new VTXO being locked to a deprecated key; enforcement is lazy — cutoff is checked at the next *live* spend (`RequiresForfeit() == true`), while swept/expired VTXOs bypass it and stay recoverable after cutoff. Output-taptree reveal + current-signer enforcement (all closures must commit to the same, current signer) is deferred to a follow-up. Worth a doc note that clients must derive new addresses against `GetInfo.SignerPubkey`.

pkg/arkd-wallet/config/config.go
```diff
@@ -181,3 +204,41 @@ func getNetwork() (arklib.Network, error) {
...
+		keyPart, cutoffPart, hasCutoff := strings.Cut(entry, ":")
+		if strings.TrimSpace(keyPart) == "" {
+			return nil, fmt.Errorf("invalid signer key entry, missing hex key: %s", entry)
+		}
+
+		buf, err := hex.DecodeString(keyPart)
+		if err != nil {
+			return nil, fmt.Errorf("invalid signer key format, must be hex: %s", keyPart)
+		}
+		key, _ := btcec.PrivKeyFromBytes(buf)
```
Kukks, line 231: `btcec.PrivKeyFromBytes` reduces any byte length mod N without error, so a truncated or mistyped hex key here silently yields the *wrong* private key — every coin locked to the intended deprecated key then becomes unspendable, with no startup error. A `len(buf) != 32` check turns a silent fund-stranding misconfig into a startup failure. (The current-signer parse in `internal/config` has the same gap.) Addressed in the follow-up PR.

internal/test/e2e/e2e_test.go
```diff
@@ -6231,3 +6231,168 @@ func TestEventListenerChurn(t *testing.T) {
...
+		}})
+		require.NoError(t, err)
+		require.NotEmpty(t, res.Txid)
+
+		wg.Wait()
+		require.NoError(t, incomingErr)
+		require.NotEmpty(t, incomingFunds)
+	})
+
+	t.Run("expired_cutoff", func(t *testing.T) {
```
Kukks, line 6362: `expired_cutoff` covers the *live* rejection path but not the other half of the design: a swept/expired deprecated-key VTXO must still be recoverable after cutoff (the `RequiresForfeit() == false` case bypasses `validateVtxoInput`). That path is currently untested, so a future change to `RequiresForfeit()` or the validation ordering could silently strand recoverable funds. The follow-up PR adds an e2e: rotate with a past cutoff, expire + sweep the old-key VTXO, then `Settle(WithRecoverableVtxos())` succeeds.

altafan (approved):

PR #1098, opened 2026-06-11
Require arkd-wallet to be initialized and unlocked out of band

bitcoin-coder-bob:
## Why

First step towards letting a single `arkd` front more than one LP, i.e. multiple `arkd-wallet`s. Before arkd can treat wallets as primary/fallback, it has to stop owning their lifecycle: arkd should not be initializing or unlocking wallets, since handling many wallets with different passwords from arkd would be unworkable.

This PR makes arkd expect each `arkd-wallet` to be initialized and unlocked out of band, and repurposes the existing unlock phase to unlock only the macaroon (admin auth) service.

- **#1098 (this PR)** — base `master`
- #1099 — stacked on this branch (`bob/arkd-require-unlocked-wallet`), merges after this
- #1101 — sweep fallback, stacked on #1099
- #1102 — wallet addresses in settings (draft), stacked on #1101

This branch was rebased onto `master` once #939 merged.

## What changes

**Startup contract**
- arkd no longer calls `Create`/`Restore`/`Unlock`/`Lock` on the wallet.
- At startup `ensureWalletReady()` hard-fails unless the primary `arkd-wallet` reports `initialized && unlocked`. Sync state is not enforced (it is transient and already gated by the readiness layer); a zero balance is warn-only.

**Unlock phase is now macaroon-only**
- The macaroon-store password is decoupled from the wallet password. `autoUnlock` is split into `ensureWalletReady` + `autoUnlockMacaroons`; the admin `Unlock` RPC and `arkd wallet unlock` now unlock only the macaroon (admin auth) service.
- `onUnlock` is synchronous and returns an error, so a wrong macaroon password fails fast instead of silently succeeding.
- `onReady` no longer locks the wallet. App services start exactly once via a small join: when the wallet is ready AND the macaroon service is unlocked (or `NO_MACAROONS`), whichever happens last.

**Retired wallet-init surface**
- `GenSeed`/`Create`/`Restore` return `Unimplemented`; `GetStatus` and the macaroon-only `Unlock` stay.
- `arkd wallet create` removed; `arkd wallet unlock` repurposed; now-unused CLI flags removed.

**No shutdown lock**
- arkd no longer locks the wallet on shutdown, so a (possibly shared) pre-unlocked wallet stays usable across restarts.
- The wallet's `Lock` RPC now returns `Unimplemented` (`errWalletLockManagedExternally`): locking is managed out of band on the arkd-wallet directly.

**`arkd-wallet` management CLI**
- The `arkd-wallet` binary gains a management CLI (`cmd/arkd-wallet/main.go`): `start` (default), `create`, `restore`, `unlock`, and `status` subcommands plus `--url`/`--password`/`--mnemonic`/`--addr-gap-limit` flags, driving the wallet's HTTP gateway. This is how an operator initializes/unlocks a wallet out of band now that arkd doesn't.

**Consumers and docs**
- The e2e harness initializes and unlocks the `arkd-wallet` directly via its gateway before arkd starts, and tolerates arkd's startup wait.
- README "Setup arkd" flow and the `ARKD_UNLOCKER_*` descriptions are updated; a docker-compose note is added.

## Operator impact

Each `arkd-wallet` must be initialized and unlocked out of band (via its own API) before arkd starts; otherwise arkd refuses to start. See the updated README "Setup arkd" section.

## Testing

- `go build ./...`, `go vet`, `make lint` (0 issues), the e2e package compiles, and unit tests for the touched pure-Go packages pass.
- A multi-agent adversarial review over the diff found no production-code defects; the harness and doc findings it raised are addressed.

## Summary by CodeRabbit

* **New Features**
  * Added `arkd-wallet` CLI commands for wallet lifecycle management (start, create, restore, unlock, status) via its gateway.
* **Changes**
  * `arkd` no longer initializes or unlocks wallets; wallets must be initialized/unlocked out-of-band before startup.
  * With macaroon (admin auth) enabled, the macaroon/admin-auth store must be unlocked for authenticated admin requests to work.
  * Updated CLI behavior/messages around macaroon unlocking and disabled wallet create/restore and related flags in `arkd`.
* **Documentation**
  * Updated README and regtest compose notes to match the new startup/auth sequence.
* **Tests**
  * Adjusted end-to-end test setup to use the external wallet gateway and new readiness flow.


ghost (changes requested):
## Arkana Code Review — arkade-os/arkd#1098

**Reviewer:** Arkana (aggressive review)
**Verdict:** Request changes — cross-repo breakage + one concurrency concern

---

### Architecture assessment

The lifecycle decoupling is sound. Letting `arkd-wallet` own its own init/unlock is the right prerequisite for multi-wallet. The two-phase join (`walletReady` AND macaroon-unlocked) is correctly synchronized via `atomic.Bool` + idempotent `CompareAndSwap` in `startAppServices`. The `GetReadyUpdate` gRPC stream replays current state on subscribe (`wallet_handler.go:398-403`), so late-joining arkd instances won't miss the ready event.

The `onUnlock` → synchronous error return is a solid upgrade over the old fire-and-forget goroutine — wrong macaroon passwords now fail fast at the RPC boundary.

---

### 🔴 CRITICAL: Cross-repo breakage

`GenSeed`, `Create`, `Restore` now return `Unimplemented`. **At least 7 downstream repos** call these through arkd's admin API and will break:

| Repo | File | What breaks |
|------|------|-------------|
| **kms-unlocker** | `internal/infrastructure/arkd-client/client.go:184-214` | Calls `GenSeed()` and `Create()` directly via gRPC. **Hard break.** |
| **arkade-kotlin** | `buildSrc/src/main/java/SetupTestTask.kt:47` | `arkd wallet create --password secret` CLI removed |
| **layerzero-usdt0-arkade-demo** | `scripts/bootstrap-server-wallet.sh:75-87` | Curls `/v1/admin/wallet/seed` + `/v1/admin/wallet/create` |
| **emulator** | `test/tx_test.go:53-61` | `setupServerWalletAndCLI` calls seed/create/unlock via arkd admin |
| **introspector-review** | `test/tx_test.go:53-61` | Same pattern as emulator |
| **arkd-pentester** | `scripts/wait-for-arkd.sh:35-80` | Calls seed/create via arkd admin API |
| **arkade-regtest** | `docker/compose.ark.yml:58-59` | Uses `ARKD_UNLOCKER_TYPE=env` + `ARKD_UNLOCKER_PASSWORD` — semantics silently change from wallet→macaroon-only; the wallet init/unlock step is now missing entirely |

**Recommendation:** Either (a) land companion PRs for each affected repo simultaneously, or (b) add a migration shim — e.g. have `GenSeed`/`Create`/`Restore` proxy to the `arkd-wallet` gateway with a deprecation log, so downstream repos have a grace period. At minimum, document the breaking change in a migration guide and file tracking issues for each consumer.

The **kms-unlocker** breakage is the most urgent — it's infrastructure that operators depend on.

---

### 🟡 IMPORTANT: `Lock` RPC still exposed but arkd no longer locks on shutdown

`internal/core/application/service.go:392-393` — the `Lock` call is correctly removed from shutdown. But `walletHandler.Lock` (walletservice.go:149-157) is still registered and callable. An operator or automation calling `Lock` via the admin API would lock the shared wallet, breaking other consumers.

**Recommendation:** Either (a) stub `Lock` to return `Unimplemented` with a message like `errWalletManagedExternally` (consistent with the other stubs), or (b) add a prominent warning log when `Lock` is called, noting that it will affect all arkd instances sharing this wallet.

---

### 🟢 GOOD

1. **`tryStartAppServices` concurrency** (`service.go:~590-600`): The two-condition join is safe. Both `onReady` and `onUnlock` can call it from different goroutines; `startAppServices` is idempotent via `CompareAndSwap`. No race.

2. **`ensureWalletReady` balance check** (`service.go:~635-638`): Zero-balance warn-only is the right call — the wallet might be intentionally empty for a fresh setup that's about to be funded.

3. **`waitUntilReady` timeout** (`utils_test.go:~672-690`): Previously could loop forever. The 2-minute timeout + `defer ticker.Stop()` fix both the hang risk and the resource leak. Same for `setupArkdWallet`.

4. **`autoUnlockMacaroons` early return on `NoMacaroons`** (`service.go:~645-647`): Clean short-circuit that the old `autoUnlock` lacked.

5. **e2e `restartArkd`** correctly drops the unlock step since the wallet stays unlocked across container restarts.

---

### 🟡 MINOR

1. **`walletservice.go:~67-68` — `listenWhenReady` guard**: Changed from `onInit != nil && onUnlock != nil && onReady != nil` to `onReady != nil`. Correct since `onInit` is removed, but `onUnlock` can now be nil if passed as nil. If `onUnlock` is nil and someone calls the `Unlock` RPC, the nil check at line ~78 in the diff handles it. Fine, but a comment would clarify intent.

2. **`service.go` — `walletReady` never resets to false**: If the wallet goes offline and comes back, `walletReady` stays true. This is fine because `startAppServices` only fires once via CAS, and the readiness interceptor separately tracks live wallet state. But worth a comment explaining why one-shot semantics are correct here.

3. **`e2e_test.go:3592` — comment says "restart arkd to restart the sweeper"**: The old comment was "lock/unlock the wallet to restart the sweeper." The new comment is accurate but doesn't explain *how* the sweeper gets restarted — is it on `startAppServices`? A reader unfamiliar with the sweeper lifecycle would benefit from slightly more context.

---

### ⚠️ PROTOCOL-CRITICAL FLAG

This PR changes the startup and shutdown lifecycle for a Bitcoin L2 server. While it doesn't directly touch VTXO handling, signing, or round lifecycle code, the wallet lock removal on shutdown and the new hard-fail startup contract affect the operational security posture of the system. **Requesting human review before merge.**

Specifically: if an operator's `arkd-wallet` crashes and restarts locked while arkd is running, arkd will continue to report `walletReady=true` (the atomic never resets). The readiness interceptor gates requests separately, but the disconnect between these two signals should be documented or unified.

---

**TL;DR:** Good architectural direction. Fix the cross-repo breakage (especially kms-unlocker), decide on the `Lock` RPC exposure, and this is ready for human sign-off.

ghost (changes requested):
## Arkana Follow-up Review — New Commit `f68b468f`

**Scope:** `arkd-wallet: add create/unlock/status subcommands`

This partially addresses the cross-repo breakage from my first review by giving operators a first-class CLI to init/unlock the wallet. Good direction. Three issues in the new code:

---

### 🔴 JSON Injection via `fmt.Sprintf` — wallet password/mnemonic not escaped

`cmd/arkd-wallet/main.go:148-151`, `156`, `163`, `170`:

```go
body := fmt.Sprintf(
    `{"seed": "%s", "password": "%s", "gap_limit": %d}`,
    mnemonic, password, ctx.Uint64("addr-gap-limit"),
)
```

If a password contains `"` or `\` (both valid in passwords), the resulting JSON is malformed — the request will fail silently or produce an unexpected payload. A password like `pass"word` yields `{"password": "pass"word"}` which is broken JSON. Worse, a crafted mnemonic like `word1 word2", "extra_field": "val` could inject fields into the request body.

**Fix:** Use `json.Marshal` to construct payloads:
```go
payload := map[string]any{
    "seed":      mnemonic,
    "password":  password,
    "gap_limit": ctx.Uint64("addr-gap-limit"),
}
data, _ := json.Marshal(payload)
```

This affects all four `fmt.Sprintf` JSON constructions in `createAction` and `unlockAction`.

---

### 🟡 Password visible in process list (`/proc/*/cmdline`)

`cmd/arkd-wallet/main.go:63` — `--password` as a CLI flag means the wallet password is visible to any process on the host via `ps aux` or `/proc/<pid>/cmdline`. For a Bitcoin L2 wallet, this is a meaningful exposure in shared hosting or container environments where process namespaces aren't isolated.

**Recommendation:** Support `--password-file` (reads from a file) or read from stdin when `--password` is `-`. This can be a follow-up, but worth noting as a security gap.

---

### 🟢 Minor: Unreachable code after `log.Exit(0)`

`cmd/arkd-wallet/main.go:128-130`:
```go
log.Exit(0)

return nil
```

`log.Exit` calls `os.Exit`, so `return nil` is dead code. Not harmful, but `go vet` may flag it in the future. Consider restructuring to call `svc.Stop()` explicitly and return, or accept the dead code with a comment.

---

### Previous review items — status

My first-review findings (cross-repo breakage, `Lock` RPC exposure) still stand. This commit partially addresses the operator UX gap but doesn't resolve the downstream consumer breakage (kms-unlocker, arkade-regtest, etc.). The JSON injection bug above is the new blocker from this commit.

ghost (changes requested):
## Arkana Review — Rebase Check (force-push `2026-06-12T06:52:43Z`)

Diff is **identical** to pre-rebase. No new code changes — just a rebase onto latest main.

**My two previous reviews still apply in full:**

### Blockers (unchanged)
1. **🔴 Cross-repo breakage** — `GenSeed`/`Create`/`Restore` returning `Unimplemented` breaks at least 7 downstream repos (kms-unlocker, arkade-regtest, emulator, etc.). Needs companion PRs or a migration shim.
2. **🔴 JSON injection in `cmd/arkd-wallet/main.go`** — `fmt.Sprintf` for JSON payloads with user-supplied password/mnemonic. Use `json.Marshal`.
3. **🟡 `Lock` RPC still callable** but arkd no longer locks on shutdown — inconsistent behavior for operators.

### ⚠️ Protocol-critical flag still active
Wallet lifecycle changes affect operational security. Human sign-off required before merge.

No new findings from the rebase.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (changes requested):
## Arkana Follow-up Review — Commit `03bc417a`

**Scope:** Only the new commit since my last review.

---

### ✅ Issues addressed

All three fixes are correct:

1. **JSON injection (cmd/arkd-wallet/main.go:196-202)** — `jsonBody()` uses `json.Marshal` instead of `fmt.Sprintf`. Passwords/mnemonics with `"`, `\`, or other special chars are now properly escaped. Good.

2. **Lock RPC disabled (walletservice.go:110-115)** — Returns `Unimplemented` with the shared `errWalletManagedExternally` message. Correct — locking a shared wallet from arkd would be dangerous.

3. **Clean shutdown (cmd/arkd-wallet/main.go:127)** — `svc.Stop()` replaces `log.Exit(0)`. The old `log.Exit` would skip the deferred return. Now the signal handler calls `Stop()` directly and returns cleanly.

### ⚠️ Minor nit: double-Stop on fatal

`log.RegisterExitHandler(svc.Stop)` (line 124) is still registered, so if any code path triggers `log.Fatal()` after the signal handler already called `svc.Stop()`, `Stop()` runs twice. `GracefulStop()` and `server.Shutdown()` are safe to double-call, but `closeFn` (which closes wallet + scanner) may not be. Consider a `sync.Once` wrapper or nil-guard in `Stop()`. **Not a blocker** — just a hardening suggestion.

### 🔴 Previous concerns still outstanding

My cross-repo breakage findings from the initial review are **not addressed** by this commit. `kms-unlocker` and `arkade-kotlin` still call the now-Unimplemented `GenSeed`/`Create` RPCs. No companion PRs exist for those repos yet.

**This PR should not merge until those downstream consumers are updated or a migration plan is documented.**

ghost (changes requested):
## Aggressive Code Review — arkade-os/arkd#1098

**Reviewer**: Arkana (automated, protocol-adjacent review)  
**Verdict**: Request changes + **flag for human review** (wallet lifecycle = key-store security boundary)

---

### Architecture

The direction is sound: decoupling wallet lifecycle from arkd is the right prerequisite for multi-wallet. The split into `ensureWalletReady` (stateless check) + `autoUnlockMacaroons` (auth-only) + `tryStartAppServices` (join gate) is clean. The `onUnlock` → synchronous + returns error is a strict improvement over fire-and-forget goroutines.

---

### 🔴 Critical — Potential startup deadlock when wallet is already synced

**`internal/interface/grpc/service.go`** — `ensureWalletReady()` vs `walletReady` atomic

`ensureWalletReady()` confirms the wallet is initialized and unlocked but **does not set `s.walletReady.Store(true)`**. The only path to set `walletReady` is `onReady()`, which fires from `listenWhenReady()` reading the wallet's `GetReadyUpdate` channel/stream.

The wallet-side `GetReadyUpdate` implementation (in `pkg/arkd-wallet/core/application/wallet/service.go`) performs a **non-blocking send** with a `default` fallback when the wallet is already unlocked at subscription time:

```go
select {
case w.readyCh <- true:
case <-ctx.Done():
default:
    log.Warn("could not send event for ready update, channel full")
}
```

If the wallet is already fully synced when arkd subscribes, and the goroutine's send falls through to `default` before the listener reads:

1. `onReady()` never fires  
2. `walletReady` stays `false`  
3. `tryStartAppServices()` never proceeds  
4. **arkd appears started but never serves**

This was a latent race before, but this PR **dramatically increases its likelihood** because the wallet is now *always* expected to be ready before arkd starts.

**Fix**: Set `s.walletReady.Store(true)` at the end of `ensureWalletReady()`. This is safe — you just verified the wallet IS ready. The `tryStartAppServices` join-gate then works immediately from whichever of `onUnlock`/`onReady` fires first, and the atomic + `CompareAndSwap` in `startAppServices` still prevents double-start.

```go
func (s *service) ensureWalletReady() error {
    // ... existing checks ...
    s.walletReady.Store(true)   // <-- add this
    return nil
}
```

---

### 🔴 Critical — arkd-wallet CLI sends wallet password over plaintext HTTP

**`cmd/arkd-wallet/main.go`** — `walletHTTPClient()`, `walletPost()`

The new `create` and `unlock` CLI commands use a bare `http.Client` with no TLS support. The `--url` flag accepts arbitrary URLs (default `http://localhost:6060`), so if an operator points at a remote host, the wallet password transits in cleartext.

The old `arkd wallet create` used `getCredentials()` which included TLS config from macaroon/cert flags. The new CLI has no equivalent.

**Fix**: At minimum, support `https://` URLs by adding optional `--tls-cert` / `--insecure` flags. For localhost-only use, consider validating the URL and refusing non-TLS for non-loopback addresses.

---

### 🟡 High — Cross-repo breakage (downstream consumers)

`GenSeed`, `Create`, `Restore` now return `Unimplemented`; `Lock` also returns `Unimplemented`. Known consumers that will break:

| Consumer | Pattern | File |
|----------|---------|------|
| `emulator` | `POST /v1/admin/wallet/create`, `/unlock` | `test/tx_test.go` |
| `layerzero-usdt0-arkade-demo` | `bootstrap-server-wallet.sh` | `scripts/` |
| `arkd-pentester` | wallet create/unlock in plans | `docs/` |
| `rust-sdk` (`ark-rest`) | Generated `wallet_initializer_service_create`, `_gen_seed`, `_restore` | `src/apis/` |

I see no companion PRs open on those repos. If this merges without coordinated updates, CI on those repos breaks and any integration using the Rust SDK's `create`/`restore` calls will get `Unimplemented` at runtime.

**Action**: Open tracking issues or companion PRs on affected repos before merge, or add a migration/deprecation window (return `FailedPrecondition` with a descriptive message instead of `Unimplemented` for one release cycle).

---

### 🟡 High — Password visible in process list

**`cmd/arkd-wallet/main.go`** — `walletPasswordFlag`

`--password` as a required CLI flag means the password is visible in `ps aux`, `/proc/*/cmdline`, shell history, and process monitoring tools. This is the primary wallet password protecting the key store.

**Suggestion**: Accept password from stdin when the flag is omitted (e.g. prompt, or pipe), or support `--password-file`. Keep the flag for scripted use but document the risk.

---

### 🟡 Medium — e2e test still uses unsafe JSON construction

**`internal/test/e2e/utils_test.go:~line 583`** — `setupArkdWallet()`

```go
body := fmt.Sprintf(`{"seed": "%s", "password": "%s"}`, seed.Seed, password)
```

If `seed.Seed` ever contains a `"` or `\`, this produces malformed JSON. The new `cmd/arkd-wallet/main.go` correctly uses `json.Marshal` via `jsonBody()` — apply the same pattern in the test helper for consistency.

---

### 🟢 Low — Nits

1. **`cmd/arkd-wallet/main.go:118`** — `walletUrlFlag` should be `walletURLFlag` per Go naming conventions (`URL` is an initialism).

2. **`internal/interface/grpc/handlers/walletservice.go`** — `errWalletManagedExternally` says "initialize the arkd-wallet directly" but is also used by the `Lock` handler. The message is misleading for lock — consider a separate message like `"arkd no longer locks the wallet: lock the arkd-wallet directly"`.

3. **`internal/interface/grpc/service.go`** — `ensureWalletReady` logs a warning for zero balance but doesn't mention which wallet address. When multi-wallet lands (PR #1099+), this will be even less useful. Consider logging the wallet address.

4. **`docker-compose.regtest.yml`** — The `restart: unless-stopped` + comment is good operational documentation. 👍

---

### Protocol-Critical Flag

While this PR doesn't directly change VTXO/signing/forfeit/round/exit logic, it **redefines the security boundary around the key store**: who creates, unlocks, and locks the wallet. The startup deadlock in finding #1 could cause arkd to appear running but never start app services (no rounds, no exits, no sweeps). The plaintext password issue in finding #2 affects key-store security.

**⚠️ Requesting human review before merge.**

---

### Summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🔴 Critical | `ensureWalletReady` doesn't set `walletReady` atomic → startup deadlock |
| 2 | 🔴 Critical | arkd-wallet CLI sends password over plaintext HTTP to remote hosts |
| 3 | 🟡 High | Cross-repo breakage: 4+ consumers use now-Unimplemented RPCs |
| 4 | 🟡 High | Wallet password visible in process list via `--password` flag |
| 5 | 🟡 Medium | e2e test uses `fmt.Sprintf` for JSON (injection-unsafe) |
| 6 | 🟢 Low | Naming, error messages, log detail (3 nits) |

ghost (changes requested):
## Arkana Follow-up Review — Commit `513eca3f`

**Scope:** Only what changed since my last review (`517fe5c6` → `513eca3f`).

---

### ✅ Issues addressed by this commit

1. **JSON injection in `cmd/arkd/commands.go:309-312`** — `walletUnlockAction` now uses `json.Marshal` instead of `fmt.Sprintf`. ✅
2. **Separate Lock RPC error in `walletservice.go:36-38`** — New `errWalletLockManagedExternally` constant distinguishes lock from create/restore/seed errors. ✅
3. **Wallet address in startup logs (`service.go:650-656`)** — Balance warnings now include `s.appConfig.WalletAddr` for multi-wallet debugging. ✅
4. **e2e JSON construction (`utils_test.go:611-625`)** — `setupArkdWallet` uses `json.Marshal` for both create and unlock payloads. ✅
5. **e2e HTTP helper hardening (`utils_test.go:660-695`)** — `get` and `post` now check HTTP status codes and properly `defer resp.Body.Close()`. ✅

All five are clean fixes. No issues.

---

### 🔴 CRITICAL — Startup deadlock still present (unchanged from last review)

**The core issue**: `ensureWalletReady()` (`service.go:638-662`) validates the wallet is initialized and unlocked but **never sets `s.walletReady`**. The only thing that sets `walletReady = true` is `onReady()` (`service.go:609`), which fires from `listenWhenReady()` (`walletservice.go:103-112`) via `GetReadyUpdate()`.

**Trace the startup when the wallet is already synced:**

1. `newServer()` → `NewWalletInitializerHandler()` → spawns `go listenWhenReady()` → subscribes to `GetReadyUpdate()` stream
2. `Start()` → `ensureWalletReady()` → confirms wallet is init+unlocked → **does not set `walletReady`**
3. `Start()` → `autoUnlockMacaroons()` → `onUnlock()` → `tryStartAppServices()` → **`walletReady.Load()` is `false`** → returns immediately
4. If wallet was **already synced** before the `GetReadyUpdate` subscription in step 1, the stream never fires a transition event
5. `onReady()` is **never called** → `walletReady` stays `false` → **services never start**

The server appears running but never serves app traffic. This is a **silent failure** — no error is logged because `tryStartAppServices` returns early without error.

**Fix** (one line, at end of `ensureWalletReady`):

```go
s.walletReady.Store(true)
```

Since `ensureWalletReady()` already confirms the wallet is initialized and unlocked, and the comment says "sync state is intentionally not enforced here: it is transient and already gated by the readiness layer," setting `walletReady = true` here is semantically correct and eliminates the race. `tryStartAppServices()` will then proceed when `autoUnlockMacaroons()` calls it, and the readiness interceptor independently gates traffic until the wallet reports synced.

---

### 🟡 Previously flagged — still open (not blocking this commit)

- **Cross-repo breakage**: `GenSeed`/`Create`/`Restore` returning `Unimplemented` will break `kms-unlocker`, `arkade-regtest`, `emulator`, and other downstream consumers. Need companion PRs or a migration shim. This is an architectural concern beyond this commit.

---

### ⚠️ Protocol-critical flag

Wallet lifecycle changes affect key-store security boundaries. **Human review required** before merge, even after the deadlock fix.

ghost (changes requested):
## Arkana Follow-up Review — Commit `513eca3f`

**Scope:** Only what changed since my last review (`517fe5c6` → `513eca3f`).

---

### ✅ Issues addressed by this commit

1. **JSON injection in `cmd/arkd/commands.go:309-312`** — `walletUnlockAction` now uses `json.Marshal` instead of `fmt.Sprintf`. ✅
2. **Separate Lock RPC error in `walletservice.go:36-38`** — New `errWalletLockManagedExternally` constant distinguishes lock from create/restore/seed errors. ✅
3. **Wallet address in startup logs (`service.go:650-656`)** — Balance warnings now include `s.appConfig.WalletAddr` for multi-wallet debugging. ✅
4. **e2e JSON construction (`utils_test.go:611-625`)** — `setupArkdWallet` uses `json.Marshal` for both create and unlock payloads. ✅
5. **e2e HTTP helper hardening (`utils_test.go:660-695`)** — `get` and `post` now check HTTP status codes and properly `defer resp.Body.Close()`. ✅

All five are clean fixes. No issues.

---

### 🔴 CRITICAL — Startup deadlock still present (unchanged from last review)

**The core issue**: `ensureWalletReady()` (`service.go:638-662`) validates the wallet is initialized and unlocked but **never sets `s.walletReady`**. The only thing that sets `walletReady = true` is `onReady()` (`service.go:609`), which fires from `listenWhenReady()` (`walletservice.go:103-112`) via `GetReadyUpdate()`.

**Trace the startup when the wallet is already synced:**

1. `newServer()` → `NewWalletInitializerHandler()` → spawns `go listenWhenReady()` → subscribes to `GetReadyUpdate()` stream
2. `Start()` → `ensureWalletReady()` → confirms wallet is init+unlocked → **does not set `walletReady`**
3. `Start()` → `autoUnlockMacaroons()` → `onUnlock()` → `tryStartAppServices()` → **`walletReady.Load()` is `false`** → returns immediately
4. If wallet was **already synced** before the `GetReadyUpdate` subscription in step 1, the stream never fires a transition event
5. `onReady()` is **never called** → `walletReady` stays `false` → **services never start**

The server appears running but never serves app traffic. This is a **silent failure** — no error is logged because `tryStartAppServices` returns early without error.

**Fix** (one line, at end of `ensureWalletReady`):

```go
s.walletReady.Store(true)
```

Since `ensureWalletReady()` already confirms the wallet is initialized and unlocked, and the comment says "sync state is intentionally not enforced here: it is transient and already gated by the readiness layer," setting `walletReady = true` here is semantically correct and eliminates the race. `tryStartAppServices()` will then proceed when `autoUnlockMacaroons()` calls it, and the readiness interceptor independently gates traffic until the wallet reports synced.

---

### 🟡 Previously flagged — still open (not blocking this commit)

- **Cross-repo breakage**: `GenSeed`/`Create`/`Restore` returning `Unimplemented` will break `kms-unlocker`, `arkade-regtest`, `emulator`, and other downstream consumers. Need companion PRs or a migration shim. This is an architectural concern beyond this commit.

---

### ⚠️ Protocol-critical flag

Wallet lifecycle changes affect key-store security boundaries. **Human review required** before merge, even after the deadlock fix.

bitcoin-coder-bob:
> ## Arkana Follow-up Review — Commit `513eca3f`
> **Scope:** Only what changed since my last review (`517fe5c6` → `513eca3f`).
> 
> ### ✅ Issues addressed by this commit
> 1. **JSON injection in `cmd/arkd/commands.go:309-312`** — `walletUnlockAction` now uses `json.Marshal` instead of `fmt.Sprintf`. ✅
> 2. **Separate Lock RPC error in `walletservice.go:36-38`** — New `errWalletLockManagedExternally` constant distinguishes lock from create/restore/seed errors. ✅
> 3. **Wallet address in startup logs (`service.go:650-656`)** — Balance warnings now include `s.appConfig.WalletAddr` for multi-wallet debugging. ✅
> 4. **e2e JSON construction (`utils_test.go:611-625`)** — `setupArkdWallet` uses `json.Marshal` for both create and unlock payloads. ✅
> 5. **e2e HTTP helper hardening (`utils_test.go:660-695`)** — `get` and `post` now check HTTP status codes and properly `defer resp.Body.Close()`. ✅
> 
> All five are clean fixes. No issues.
> 
> ### 🔴 CRITICAL — Startup deadlock still present (unchanged from last review)
> **The core issue**: `ensureWalletReady()` (`service.go:638-662`) validates the wallet is initialized and unlocked but **never sets `s.walletReady`**. The only thing that sets `walletReady = true` is `onReady()` (`service.go:609`), which fires from `listenWhenReady()` (`walletservice.go:103-112`) via `GetReadyUpdate()`.
> 
> **Trace the startup when the wallet is already synced:**
> 
> 1. `newServer()` → `NewWalletInitializerHandler()` → spawns `go listenWhenReady()` → subscribes to `GetReadyUpdate()` stream
> 2. `Start()` → `ensureWalletReady()` → confirms wallet is init+unlocked → **does not set `walletReady`**
> 3. `Start()` → `autoUnlockMacaroons()` → `onUnlock()` → `tryStartAppServices()` → **`walletReady.Load()` is `false`** → returns immediately
> 4. If wallet was **already synced** before the `GetReadyUpdate` subscription in step 1, the stream never fires a transition event
> 5. `onReady()` is **never called** → `walletReady` stays `false` → **services never start**
> 
> The server appears running but never serves app traffic. This is a **silent failure** — no error is logged because `tryStartAppServices` returns early without error.
> 
> **Fix** (one line, at end of `ensureWalletReady`):
> 
> ```go
> s.walletReady.Store(true)
> ```
> 
> Since `ensureWalletReady()` already confirms the wallet is initialized and unlocked, and the comment says "sync state is intentionally not enforced here: it is transient and already gated by the readiness layer," setting `walletReady = true` here is semantically correct and eliminates the race. `tryStartAppServices()` will then proceed when `autoUnlockMacaroons()` calls it, and the readiness interceptor independently gates traffic until the wallet reports synced.
> 
> ### 🟡 Previously flagged — still open (not blocking this commit)
> * **Cross-repo breakage**: `GenSeed`/`Create`/`Restore` returning `Unimplemented` will break `kms-unlocker`, `arkade-regtest`, `emulator`, and other downstream consumers. Need companion PRs or a migration shim. This is an architectural concern beyond this commit.
> 
> ### ⚠️ Protocol-critical flag
> Wallet lifecycle changes affect key-store security boundaries. **Human review required** before merge, even after the deadlock fix.

Pushing back on "deadlock still present" .

The arkd-wallet server replays Ready: true to every new subscriber on connect when the wallet is already initialized+synced+unlocked (pkg/arkd-wallet/interface/grpc/handlers/wallet_handler.go:398-403), and the arkd client opens a fresh stream per consumer (internal/infrastructure/wallet/wallet_client.go:251). So listenWhenReady receives the event and onReady() fires → walletReady=true even when the wallet is already ready at boot. No deadlock. The review cited the wallet core's non-blocking default send but missed the per-subscriber replay at the gRPC boundary. Also, the proposed fix (walletReady.Store(true) inside ensureWalletReady) would regress: ensureWalletReady deliberately checks only init+unlocked, not synced, so it would start app services before the wallet is synced.

ghost (approved):
## Arkana Follow-up Review — Rebase `d8c8b92f..9e5e7cb0`

**Scope:** Branch was rebased onto master (all commits re-stamped `2026-06-13T01:45:29Z`). No new code changes since my last review of `513eca3f`.

---

### Previous findings — all resolved

1. ✅ JSON injection in CLI payloads — `json.Marshal` throughout
2. ✅ Lock RPC returns `Unimplemented` with distinct error
3. ✅ Wallet address in startup balance warnings
4. ✅ e2e JSON construction via `json.Marshal`
5. ✅ e2e HTTP helpers check status codes, `defer resp.Body.Close()`
6. ✅ `sync.Once` for arkd-wallet service stop

### Deadlock concern — retracted

My previous finding about a startup deadlock in `ensureWalletReady()` / `onReady()` was incorrect. The author correctly pointed out that `wallet_handler.go:398-403` replays `Ready: true` to every new `GetReadyUpdate` subscriber when the wallet is already in the ready state. Verified in source. The race I described cannot occur.

### Cross-repo impact — still relevant (not blocking)

`GenSeed`/`Create`/`Restore`/`Lock` returning `Unimplemented` will break any downstream consumer that calls these RPCs (kms-unlocker, arkade-regtest, emulator). Companion PRs or migration shims should land before or alongside this merge.

---

### ⚠️ Protocol-critical flag

Wallet lifecycle ownership changes affect key-store security boundaries. **Human approval required** before merge.

LGTM on the code. Approving to unblock, but a human must also approve per protocol.

ghost (approved):
## Arkana Follow-up Review — Rebase `d8c8b92f..9e5e7cb0`

**Scope:** Branch was rebased onto master (all commits re-stamped `2026-06-13T01:45:29Z`). No new code changes since my last review of `513eca3f`.

---

### Previous findings — all resolved

1. ✅ JSON injection in CLI payloads — `json.Marshal` throughout
2. ✅ Lock RPC returns `Unimplemented` with distinct error
3. ✅ Wallet address in startup balance warnings
4. ✅ e2e JSON construction via `json.Marshal`
5. ✅ e2e HTTP helpers check status codes, `defer resp.Body.Close()`
6. ✅ `sync.Once` for arkd-wallet service stop

### Deadlock concern — retracted

My previous finding about a startup deadlock in `ensureWalletReady()` / `onReady()` was incorrect. The author correctly pointed out that `wallet_handler.go:398-403` replays `Ready: true` to every new `GetReadyUpdate` subscriber when the wallet is already in the ready state. Verified in source. The race I described cannot occur.

### Cross-repo impact — still relevant (not blocking)

`GenSeed`/`Create`/`Restore`/`Lock` returning `Unimplemented` will break any downstream consumer that calls these RPCs (kms-unlocker, arkade-regtest, emulator). Companion PRs or migration shims should land before or alongside this merge.

---

### ⚠️ Protocol-critical flag

Wallet lifecycle ownership changes affect key-store security boundaries. **Human approval required** before merge.

LGTM on the code. Approving to unblock, but a human must also approve per protocol.

internal/interface/grpc/service.go
```diff
@@ -569,60 +578,56 @@ func (s *service) newServer(tlsConfig *tls.Config, withPprof bool) error {
...
+// tryStartAppServices starts the app services exactly once, when both the wallet
+// is ready and the macaroon service is unlocked (or macaroons are disabled). It
+// is safe to call repeatedly and from multiple goroutines.
+func (s *service) tryStartAppServices() {
+	if !s.walletReady.Load() {
+		return
+	}
+	if !s.config.NoMacaroons && s.macaroonSvc.IsLocked(context.Background()) {
+		return
+	}
```
aruokhai, line 638: I reckon there should  be warning logs here, just to indicate that the App wasnt started and why

cmd/arkd-wallet/main.go
```diff
@@ -44,5 +130,135 @@ func main() {
...
+	body, err := jsonBody(map[string]any{"password": password})
+	if err != nil {
+		return err
+	}
+	if err := walletPost(baseURL, "/v1/wallet/unlock", body, nil); err != nil {
+		return err
+	}
+	fmt.Println("wallet unlocked")
+	return nil
+}
```
aruokhai, line 193: I believe that It will be more expressive to have a separate restoreAction

ghost (commented):
## Arkana Follow-up Review — Commit `246b7433`

**Scope:** One new commit since last review (`9e5e7cb0`). Adds two `log.Warn` calls in `tryStartAppServices()` when either the wallet isn't ready or the macaroon service is still locked.

---

### 🟡 Nit: Warn log level too noisy for normal startup

**`internal/interface/grpc/service.go:627,631`**

`tryStartAppServices()` is called from both `onUnlock()` (:608) and `onReady()` (:618). During every normal startup, whichever fires first will hit one of these warn branches — the wallet may not be ready yet when macaroons unlock, or vice versa. This means **every healthy startup emits a warning**, which:

1. Creates noise in operator logs
2. Could desensitize operators to real warnings
3. May trigger alert rules keyed on `warn`-level output

Suggestion: use `log.Debug` (or `log.Info` at most) for the expected "waiting" case. Reserve `Warn` for situations that are genuinely unusual or require attention.

```go
log.Debug("app services not started yet: waiting for the wallet to become ready")
// ...
log.Debug("app services not started yet: waiting for the macaroon service to be unlocked")
```

This is a nit — no correctness or security issue. The code is functionally correct.

---

### Previous findings — all resolved ✅

No other changes in this commit. All prior findings remain addressed.

### ⚠️ Protocol-critical flag (standing)

Wallet lifecycle ownership changes affect key-store security boundaries. **Human approval required** before merge.

internal/interface/grpc/service.go
bitcoin-coder-bob: done in: [246b743](https://github.com/arkade-os/arkd/pull/1098/commits/246b7433dc03a903d9f3811968c2c9fcc4cb481b)

cmd/arkd-wallet/main.go
bitcoin-coder-bob: done in: [a14bcb3](https://github.com/arkade-os/arkd/pull/1098/commits/a14bcb39414dfd1a980b74debdd4b2969ba5b945)

ghost (approved):
## Arkana Follow-up Review — Commits `dd8d5b00`, `5d9ad687`

**Scope:** Two new commits since last review (`c968d253`). CLI-only changes in `cmd/arkd-wallet/main.go`.

### `dd8d5b00` — Split create/restore into separate CLI commands
Clean refactor. `create` no longer accepts `--mnemonic`; a new `restore` subcommand takes `--mnemonic` as required. Shared `unlockWallet()` helper eliminates duplication. Command registration in `main()` is correct (all four subcommands present). Flag sets are appropriate: `create` gets `url`+`password`, `restore` gets `url`+`password`+`mnemonic`+`gap-limit`. No issues.

### `5d9ad687` — Warn about seed phrase security before printing it
Good security UX — two clear warning lines before the seed hits stdout. Standard practice for CLI wallet tools.

**No protocol-critical changes.** No cross-repo impact (this is a standalone CLI binary). No security concerns in the diff.

✅ LGTM. Approving the incremental delta.

ghost (approved):
## Arkana Follow-up Review — Commits `dd8d5b00`, `5d9ad687`

**Scope:** Two new commits since last review (`c968d253`). CLI-only changes in `cmd/arkd-wallet/main.go`.

### `dd8d5b00` — Split create/restore into separate CLI commands
Clean refactor. `create` no longer accepts `--mnemonic`; a new `restore` subcommand takes `--mnemonic` as required. Shared `unlockWallet()` helper eliminates duplication. Command registration in `main()` is correct (all four subcommands present). Flag sets are appropriate: `create` gets `url`+`password`, `restore` gets `url`+`password`+`mnemonic`+`gap-limit`. No issues.

### `5d9ad687` — Warn about seed phrase security before printing it
Good security UX — two clear warning lines before the seed hits stdout. Standard practice for CLI wallet tools.

**No protocol-critical changes.** No cross-repo impact (this is a standalone CLI binary). No security concerns in the diff.

✅ LGTM. Approving the incremental delta.

bitcoin-coder-bob:
Addressed one item and I'm pushing back on the rest. Details below.

### Fixed in c6525e151

**`unlockArkdWallet()` unsafe `fmt.Sprintf` for JSON.** Correct catch, now uses `json.Marshal` and matches `setupArkdWallet`. `password` is a package const (`e2e_test.go:53`) so there was no live injection, but the inconsistency was real.

### Pushback: the CRITICAL "startup deadlock" does not exist

This was already answered on 2026-06-13 and re-raised unchanged, so restating it concretely.

The claim is that when the wallet is already synced the `GetReadyUpdate` stream never fires. It does. `pkg/arkd-wallet/interface/grpc/handlers/wallet_handler.go:398-403` replays to every new subscriber on connect:

```go
if status := h.wallet.Status(stream.Context()); status.IsInitialized &&
    status.IsSynced && status.IsUnlocked {
    if err := stream.Send(&arkwalletv1.GetReadyUpdateResponse{Ready: true}); err != nil {
```

Full path: `newServer` spawns `listenWhenReady` -> client opens a fresh stream per consumer (`wallet_client.go:251`) -> replay arrives -> `onReady()` -> `walletReady = true`. The already-synced case is precisely the case that works.

The review cites the `default:` in `listenToReadyUpdate`, which is the fan-out for *subsequent* transitions, not the initial subscribe. That branch is also effectively unreachable, the listener channel is buffered at 100 (`broker.go:17`).

More importantly, the suggested one-line fix would introduce a bug. `ensureWalletReady` deliberately checks initialized and unlocked but not synced, which is documented in its comment. Setting `walletReady.Store(true)` there would start app services against an unsynced wallet and defeat the sync gate. Not applying it.

### Pushback: the `Lock` RPC is already disabled

The follow-up says the fix "has not landed." It landed in `7366cec29` on 2026-06-12, before the review that reports it open. `internal/interface/grpc/handlers/walletservice.go:114-121`:

```go
func (a *walletHandler) Lock(
	_ context.Context, _ *arkv1.LockRequest,
) (*arkv1.LockResponse, error) {
	// arkd no longer manages the wallet lifecycle; locking the (possibly shared)
	// wallet via arkd would break every other consumer, so the RPC is disabled
	return nil, status.Error(codes.Unimplemented, errWalletLockManagedExternally)
}
```

It is on `walletHandler`, not `walletInitHandler`. The rationale the review asks to add is already the comment on the stub.

### Out of scope for this PR

**Cross-repo breakage.** `GenSeed`/`Create`/`Restore` returning `Unimplemented` is the purpose of this PR, not a defect in it. Real coordination work, tracked separately. `kms-unlocker` is the only non-test consumer in the list, the other two are test setup.

**Password on argv and TLS for the wallet CLI.** Both fair, neither blocking, both separate changes. On TLS the description is imprecise: `walletHTTPClient()` sets no custom Transport, so it uses `http.DefaultTransport` and handles `https://` with the system CA pool. The actual gaps are no `--tls-cert` for self-signed certs and no guard against plaintext to a non-loopback host, with the default being `http://localhost:6060`.

### One note on the reviews

The 2026-08-26 and 2026-08-27 passes contradict each other. The first lists four open issues plus one new and calls the deadlock the merge blocker. The second reports "2 prior issues, 0 fixed" and drops the deadlock and password items without comment. No code changed between them, only `Dockerfile` and `trivy.yaml`.

bitcoin-coder-bob:
> Changes were requested 82+ days ago. @bitcoin-coder-bob the arkd-wallet out-of-band init PR has been waiting a long time — is it blocked on something?

stop asking

PR #1099, opened 2026-06-12
Dial primary plus fallback arkd-wallets

bitcoin-coder-bob:
## Why

Next step towards letting a single `arkd` front more than one LP. This teaches `arkd` to connect to a **primary** `arkd-wallet` plus **N fallback** wallets belonging to additional liquidity providers. It is purely additive: the primary path is unchanged and fallbacks are dialed and validated but not yet used by sweep (that wiring lands in a follow-up). Very simple: we just support adding other wallets and try connecting to them, we dont actually use the wallets yet for sweeps/fallbacks. The PR is just a small step to introduce new wallet support before actually _using_ the wallets for functionality.

**Stacked PRs** (merge top-down):
- ~~#939~~ — ✅ merged into `master`
- #1098 — base `master`, merges first
- **#1099 (this PR)** — base `bob/arkd-require-unlocked-wallet`
- #1101 — sweep fallback, stacked on this branch (`bob/arkd-multi-wallet-dial`)
- #1102 — wallet addresses in settings (draft), stacked on #1101

This PR depends on #1098 landing first.

## What changes

**Config**
- New `ARKD_WALLET_FALLBACK_ADDRS` (comma-separated `host:port`). Parsed with an explicit comma split (`parseWalletFallbackAddrs`), not `viper.GetStringSlice`, because the latter splits on whitespace, not commas (verified empirically), and would silently mis-parse the list.
- `walletService()` dials the primary (unchanged), then `dialFallbackWallets()` dials each fallback via an indirected `newWalletClient` (so it is unit-testable) and **hard-fails** if a fallback is unreachable or on a different network than the primary, closing any partially-dialed connections to avoid leaks. It also rejects a fallback whose address duplicates the primary or another fallback.
- Dialed fallbacks are held in `Config.walletFallbacks` and exposed via `FallbackWallets()` (each pairing the dial address with the wallet client). The primary stays `c.wallet`, so GetInfo, the forfeit pubkey/address, connector address, scanning, signing, and sweep are all untouched.

**Startup / shutdown**
- `ensureWalletReady()` now also requires each fallback wallet to be initialized and unlocked (hard-fail); the zero-balance warning stays primary-only.
- `stop()` closes the fallback connections (the primary is closed by the app service).

**Exercised end to end**
[...]

PR #1101, opened 2026-06-12
Sweep fallback across primary and fallback arkd-wallets

bitcoin-coder-bob:
## Why

Phase-2 core of letting a single `arkd` front more than one LP. #1099 dials and validates primary + fallback `arkd-wallet`s but nothing uses the fallbacks yet. This PR makes **sweeping** use them: a batch (or checkpoint, or admin) sweep is now signed by whichever of the primary/fallback wallets can sign it, so arkd can sweep batches belonging to any of its LPs.

- #1098 — base `master`
- #1099 — dial primary + fallback wallets (base `bob/arkd-require-unlocked-wallet`)
- **this PR** — base `bob/arkd-multi-wallet-dial`; merges after #1099
- #1102 — wallet addresses in settings (draft), stacked on this branch

## What changes

**Build/sign split (`tx-builder/covenantless/sweep.go`, `ports/tx_builder.go`)**
- `sweepTransaction` is split into:
  - `buildSweepTransaction(ctx, wallet, inputs) -> (unsignedTx, txid)` — destination address, fee estimate and dust limit all come from the **primary** wallet, so every signing candidate sweeps to the same output and the txid is stable regardless of who signs.
  - `signSweepTransaction(ctx, wallet, unsignedTx) -> signedTx` — signs with an arbitrary wallet; the tapscript inputs that need signing are **re-derived from the psbt** (inputs carrying a `TaprootLeafScript`), so no index list has to be threaded through the API.
- `TxBuilder.BuildSweepTx` now returns the **unsigned** tx; new `TxBuilder.SignSweepTx(wallet, unsignedTx)`.

**Fallback iteration (`application/sweeper.go`, `application/admin.go`)**
- New `buildAndSignSweepTx(builder, wallets, inputs)` builds once, then tries signing with each wallet in order, returning on first success and aggregating errors (`errors.Join`) if none can sign.
- `sweeper` and `adminService` gained a `walletFallbacks` field and a `signingWallets()` helper returning `[primary, ...fallbacks]`. The three sweeper `BuildSweepTx` call sites (batch, event-rebuild, checkpoint) and `admin.Sweep` all use the new helper.
- **Broadcasting stays on the primary wallet.** `BroadcastTransaction` only relays raw bytes to the wallet daemon, so a fallback-signed (fully valid) tx broadcasts fine via the primary; the existing BIP68 non-final retry loop is untouched.
- **Graceful degradation on the event-rebuild path.** In `createBatchSweepTask`, when a batch's outputs are already spent and no primary/fallback wallet can sign the rebuilt sweep, it no longer errors out: it logs a warning and reconciles using the unsigned tx (re-deriving the txid) so the sweep event still completes.

**Wiring (`config.go`)**
- `c.fallbackWalletServices()` is threaded into `NewService` (→ sweeper) and `NewAdminService`. With no fallbacks configured, `signingWallets()` returns just `[primary]`, so single-wallet behavior is unchanged.

## Tests

- `TestBuildAndSignSweepTx` (application): first-success short-circuit, ordered fallback, all-fail aggregated error, build-error short-circuit.
- `TestSignSweepTransaction_DerivesTapscriptIndexes` / `_NoTapscriptInputs` (covenantless): the re-derived tapscript indexes match the inputs carrying a leaf script, and the tapscript-signed psbt is forwarded to the final sign.
- `go build ./...`, `make lint` (0 issues), `go test` on the touched non-infra packages pass with `-race`.

## Notes

- The fallback path only exercises the second wallet end to end once it's actually used; #1099 already wired `arkd-wallet-2` into `docker-compose.regtest.yml` and the e2e harness, so the regtest suite now runs with a fallback present.
- Selecting *which* LP a wallet serves (vs. just "whoever can sign") and moving the designation into the settings domain are follow-ups.


ghost (changes requested):
## Arkana Code Review — arkd#1101

**Reviewer:** Arkana (aggressive review mode)
**Verdict:** Request changes — one correctness concern + protocol-critical flag

---

### Overall Assessment

Well-structured PR. The build/sign split is clean, the fallback iteration logic is correct, and the test coverage is solid. The tapscript-index re-derivation from the PSBT (`signSweepTransaction`) is the right design — it avoids threading index lists across API boundaries.

That said, I have one medium-severity finding and several observations.

---

### 🔴 M1 — "All-spent" rebuild path now fails if no wallet can sign

**File:** `internal/core/application/sweeper.go` — the call site at the old line 669 (now `buildAndSignSweepTx(s.builder, s.signingWallets(), outputsToSweep)`)

**Context:** When all batch outputs are already spent, the sweeper builds a sweep tx *without broadcasting it* — solely to reconstruct the `sweepEvent` for the projection store. The comment on the old line 667-668 says exactly this:

> "if all outputs are spent, it means we missed to mark the batch as swept, build a sweep transaction without broadcasting it. we'll use it rebuild sweepEvent."

**Problem:** Before this PR, `BuildSweepTx` both built and signed with the primary wallet. If the primary wallet couldn't sign, this path failed — but that was already a latent issue. Now `buildAndSignSweepTx` iterates *all* wallets and returns a hard error (`"no wallet could sign sweep tx"`) if none can sign.

In the "all-spent" path, the **signed tx hex is never broadcast**. Only `sweepTxId` is used (for event reconstruction at line 676). The signing is entirely wasted work. Worse: if the batch was originally swept by a wallet that's no longer configured as a fallback, this path will now error out and prevent the sweeper from reconciling its state.

**Fix:** For the all-spent path, call `builder.BuildSweepTx(inputs)` directly (which returns `(unsignedTx, txid, err)`) and use only the `txid`. No signing needed.

```go
// all outputs already spent — only need the txid for event reconciliation
unsignedTx, sweepTxId, err := s.builder.BuildSweepTx(outputsToSweep)
_ = unsignedTx
if err != nil {
    return err
}
```

Or better yet, extract a `BuildSweepTxID` helper that doesn't sign at all.

---

### 🟡 O1 — Positional struct initialization in `newSweeper` is fragile

**File:** `internal/core/application/sweeper.go:52-54`

```go
return &sweeper{
    wallet, walletFallbacks, repoManager, builder, scheduler,
    &sync.Mutex{}, make(map[string]struct{}), nil,
}
```

8 positional fields. If anyone inserts a field between `scheduler` and `locker`, this silently misassigns values (and the compiler won't catch it if the types happen to align). Use named fields:

```go
return &sweeper{
    wallet:          wallet,
    walletFallbacks: walletFallbacks,
    repoManager:     repoManager,
    builder:         builder,
    scheduler:       scheduler,
    locker:          &sync.Mutex{},
    scheduledTasks:  make(map[string]struct{}),
}
```

This is pre-existing tech debt (it was positional before), but since you're already modifying the constructor and adding a field, now is the time to fix it.

---

### 🟡 O2 — `signingWallets()` is duplicated between sweeper and adminService

**Files:** `sweeper.go` and `admin.go` — identical implementations:

```go
func (s *sweeper) signingWallets() []ports.WalletService {
    return append([]ports.WalletService{s.wallet}, s.walletFallbacks...)
}
```

Consider extracting to a shared free function or embedding a common struct. Minor, but this is the kind of duplication that drifts.

---

### 🟡 O3 — `context.Background()` in `SignSweepTx` 

**File:** `internal/infrastructure/tx-builder/covenantless/builder.go:283-286`

```go
func (b *txBuilder) SignSweepTx(
    wallet ports.WalletService, unsignedTx string,
) (signedTx string, err error) {
    ctx := context.Background()
    return signSweepTransaction(ctx, wallet, unsignedTx)
}
```

Both `BuildSweepTx` and `SignSweepTx` use `context.Background()`. The sweeper checks `s.ctx.Done()` in its BIP68 retry loop, but the actual wallet RPC calls inside `signSweepTransaction` are not cancellable. If arkd is shutting down and a wallet RPC hangs, these calls block indefinitely. This is pre-existing (the old `sweepTransaction` also used `context.Background()` via `BuildSweepTx`), but since you're refactoring the interface, consider threading context through:

```go
SignSweepTx(ctx context.Context, wallet WalletService, unsignedTx string) (string, error)
```

Not blocking, but worth a follow-up.

---

### 🟢 Positive observations

1. **Return-value swap is handled correctly.** The dangerous `(txid, signed, err)` → `(unsigned, txid, err)` reorder in `BuildSweepTx` is safely encapsulated inside `buildAndSignSweepTx`, which maps it back to `(txid, signed, err)` for all callers. I verified all 4 call sites.

2. **Tapscript index re-derivation** (`signSweepTransaction` scanning for `TaprootLeafScript`) is correct and well-tested. Eliminates a fragile index-threading dependency.

3. **Broadcast stays on primary wallet** — correct, since `BroadcastTransaction` just relays raw bytes.

4. **Test coverage is good.** `TestBuildAndSignSweepTx` covers: primary-first short-circuit, ordered fallback, all-fail aggregation, no-wallets, build-error. `TestSignSweepTransaction_DerivesTapscriptIndexes` verifies the index derivation. Edge cases are handled.

5. **`errors.Join` for aggregated sign errors** is the right pattern — gives operators visibility into which wallet rejected.

---

### ⚠️ Protocol-Critical Flag

This PR modifies **sweep transaction signing** — a protocol-critical path. Sweep transactions recover expired VTXO funds back to the operator. A bug here means stuck funds or failed sweeps.

The code looks correct to me, but per protocol: **this requires explicit human sign-off before merge.**

---

**Summary:** Fix M1 (all-spent path shouldn't require signing), consider O1-O3 as improvements. Requesting human review for protocol-critical sweep signing changes.

ghost (changes requested):
## Arkana Follow-Up Review — arkd#1101 (commit b9db0bc3)

**Focus:** Incremental changes since my last review (ded6d98c → b9db0bc3).

---

### ✅ O1 and O2 — Fixed

- `newSweeper` now uses named struct fields — good. No more positional foot-gun.
- `primaryThenFallbacks` is extracted as a shared free function used by both `sweeper.signingWallets()` and `adminService.signingWallets()` — clean.

Both changes are correct. No issues.

---

### 🔴 M1 — Still open: all-spent path still signs unnecessarily

**File:** `internal/core/application/sweeper.go:722-729`

```go
// if all outputs are spent, it means we missed to mark the batch as swept,
// build a sweep transaction without broadcasting it. we'll use it rebuild sweepEvent.
sweepTxId, sweepTx, err = buildAndSignSweepTx(
    s.builder, s.signingWallets(), outputsToSweep,
)
```

This path only needs `sweepTxId` for event reconciliation — the signed tx is never broadcast. Signing is wasted work. More importantly: if the batch was originally swept by a wallet that's since been decommissioned (no longer primary or fallback), this path hard-fails with `"no wallet could sign sweep tx"`, preventing the sweeper from reconciling that batch.

**Fix:** Call `s.builder.BuildSweepTx(outputsToSweep)` directly — it returns `(unsignedTx, txid, err)` now, which is all this code path needs:

```go
_, sweepTxId, err = s.builder.BuildSweepTx(outputsToSweep)
if err != nil {
    return err
}
```

Or if `sweepTx` is actually used downstream for the event (I checked — it is passed to `NewBatchSweptEvent` at line 735), then this path truly does need a signed tx. In that case the current behavior is acceptable, but please confirm the intent and add a comment explaining why signing is required even though the tx isn't broadcast.

---

### ⚠️ Protocol-critical flag remains

Sweep signing path — requires human sign-off before merge.

**Summary:** O1/O2 look good. M1 still needs attention — either skip signing in the all-spent path or document why it's needed.

ghost (changes requested):
## Arkana Follow-Up Review — arkd#1101 (rebase: 48fbf228..44379b33)

**Focus:** Branch was rebased/squashed. Verified the rebase is clean — O1 (named struct fields) and O2 (`primaryThenFallbacks` extraction) are correctly incorporated. README update (44379b33) is accurate. No new issues introduced.

---

### 🔴 M1 — Still open (third review): all-spent reconciliation path signs unnecessarily

**File:** `internal/core/application/sweeper.go:721-725` (in PR diff)

```go
// if all outputs are spent, it means we missed to mark the batch as swept,
// build a sweep transaction without broadcasting it. we'll use it rebuild sweepEvent.
sweepTxId, sweepTx, err = buildAndSignSweepTx(
    s.builder, s.signingWallets(), outputsToSweep,
)
```

**Why this matters:** I checked — `round.Sweep(leafVtxos, preconfirmedVtxos, sweepTxId, sweepTx)` stores `sweepTx` in `r.SweepTxs[txid]` via the `BatchSwept` event handler. So `sweepTx` isn't discarded. However:

1. **The stored hex is wrong anyway.** The all-spent path means this batch was *already swept on-chain* by some wallet. The tx we sign here has different witness data than the actual on-chain tx (segwit txid stays the same, but the raw hex differs). We're storing a "fake" signed version.

2. **Hard failure when the original wallet is decommissioned.** If the batch was swept by a wallet that's no longer configured as primary or fallback, `buildAndSignSweepTx` returns `"no wallet could sign sweep tx"`, and the sweeper permanently fails to reconcile this batch. It retries every cycle, logs errors, never succeeds — a livelock.

**Suggested fix options (pick one):**

**(A)** Skip signing, store the unsigned PSBT (if downstream consumers of `SweepTxs` only need the txid for lookups):
```go
unsignedTx, sweepTxId, err = s.builder.BuildSweepTx(outputsToSweep)
// pass unsignedTx as sweepTx — never broadcast, only used for event storage
```

**(B)** Best-effort signing, fall through on failure:
```go
sweepTxId, sweepTx, err = buildAndSignSweepTx(
    s.builder, s.signingWallets(), outputsToSweep,
)
if err != nil {
    // All outputs already spent — signing failure is non-fatal.
    // Use unsigned tx for event reconciliation.
    var unsignedTx string
    unsignedTx, sweepTxId, err = s.builder.BuildSweepTx(outputsToSweep)
    if err != nil {
        return err
    }
    sweepTx = unsignedTx
}
```

Option B is more defensive — it tries to sign (preserving current behavior when possible) but doesn't block reconciliation when it can't.

---

### ⚠️ Protocol-critical flag remains

Sweep signing path — requires human sign-off before merge.

**Summary:** Rebase is clean. M1 is the only remaining blocker — the all-spent reconciliation path can livelock when a wallet is decommissioned. Everything else looks correct.

ghost (changes requested):
## Arkana Follow-Up Review — arkd#1101 (rebase: ef4ac4df)

Verified: pure rebase onto updated base, no content changes to PR files. No new issues introduced by the rebase.

### 🔴 M1 — Still open (4th review): all-spent reconciliation path signs unnecessarily

**File:** `internal/core/application/sweeper.go:722-729`

This remains the only blocker. The all-spent code path calls `buildAndSignSweepTx` when it only needs a txid for event reconciliation. When the original signing wallet is decommissioned (no longer primary or fallback), this path hard-fails with `"no wallet could sign sweep tx"` every cycle — a permanent livelock that prevents batch reconciliation.

Suggested fix (option B from my previous review — best-effort signing, graceful fallback):

```go
sweepTxId, sweepTx, err = buildAndSignSweepTx(
    s.builder, s.signingWallets(), outputsToSweep,
)
if err != nil {
    // All outputs already spent on-chain — signing failure is non-fatal.
    // Fall back to the unsigned tx for event reconciliation.
    var unsignedTx string
    unsignedTx, sweepTxId, err = s.builder.BuildSweepTx(outputsToSweep)
    if err != nil {
        return err
    }
    sweepTx = unsignedTx
}
```

### ⚠️ Protocol-critical flag

Sweep signing path — requires human sign-off before merge.

ghost (approved):
## Arkana Follow-Up Review — arkd#1101 (commit 89271cce)

**Focus:** New commit addressing M1 — all-spent reconciliation path.

### ✅ M1 — Resolved

The all-spent path now implements option B (best-effort signing with graceful fallback):

```go
sweepTxId, sweepTx, err = buildAndSignSweepTx(...)
if err != nil {
    log.WithError(err).Warnf(...)
    sweepTx, sweepTxId, err = s.builder.BuildSweepTx(outputsToSweep)
    if err != nil {
        return err
    }
}
```

Verified:
- Return value mapping is correct: `BuildSweepTx` returns `(unsignedTx, txid, err)` → assigned to `(sweepTx, sweepTxId, err)` ✓
- Signing failure no longer blocks reconciliation — livelock eliminated ✓
- Warning log includes batch ID and error context for operator visibility ✓
- Comment clearly explains why signing failure is non-fatal here ✓

All previous findings (O1, O2, M1) are now addressed. Code is correct.

### ⚠️ Protocol-critical flag remains

This PR modifies sweep transaction signing — a protocol-critical path. The code looks correct to me, but **this requires explicit human sign-off before merge.**

ghost (approved):
## Arkana Follow-Up Review — arkd#1101 (rebase: ccb8a46f)

Verified: pure rebase onto updated base, no content changes from previously approved commit `89271cce`. All prior findings (O1, O2, M1) remain addressed. Approval carries forward.

⚠️ **Protocol-critical flag remains** — sweep signing path requires human sign-off before merge.

ghost (approved):
## Arkana Follow-Up Review — arkd#1101 (merge: 45b1aa8d)

**Focus:** Base branch merge — 3 new commits from `bob/arkd-multi-wallet-dial` (#1099).

### Changes from base branch (not this PR's code)
- `FallbackWallet{Addr, Service}` struct replaces bare `[]ports.WalletService`
- Duplicate and primary-equals-fallback validation at startup
- Error messages now use `host:port` instead of positional index

### Integration verified ✅
- `fallbackWalletServices()` correctly extracts `Service` fields for the application layer
- gRPC shutdown/readiness uses `FallbackWallets()` (needs `Addr` for error context) — correct
- `appService()` / `adminService()` wiring unchanged — still passes `[]ports.WalletService` to application code
- No merge conflict artifacts, no logic changes to this PR's sweep fallback code

All previously reviewed PR logic (build/sign split, fallback iteration, M1 reconciliation fix) is untouched. Approval carries forward.

⚠️ **Protocol-critical flag remains** — sweep signing path requires human sign-off before merge.

ghost (approved):
## Arkana Follow-Up Review — arkd#1101 (merge: 45b1aa8d)

**Focus:** Base branch merge — 3 new commits from `bob/arkd-multi-wallet-dial` (#1099).

### Changes from base branch (not this PR's code)
- `FallbackWallet{Addr, Service}` struct replaces bare `[]ports.WalletService`
- Duplicate and primary-equals-fallback validation at startup
- Error messages now use `host:port` instead of positional index

### Integration verified ✅
- `fallbackWalletServices()` correctly extracts `Service` fields for the application layer
- gRPC shutdown/readiness uses `FallbackWallets()` (needs `Addr` for error context) — correct
- `appService()` / `adminService()` wiring unchanged — still passes `[]ports.WalletService` to application code
- No merge conflict artifacts, no logic changes to this PR's sweep fallback code

All previously reviewed PR logic (build/sign split, fallback iteration, M1 reconciliation fix) is untouched. Approval carries forward.

⚠️ **Protocol-critical flag remains** — sweep signing path requires human sign-off before merge.

ghost (approved):
## Arkana Follow-Up Review — arkd#1101 (commit be5ee751)

**Diff since last review (`45b1aa8d..be5ee751`):** Pure refactor/comment cleanup. No logic changes.

### Changes reviewed

1. **`primaryThenFallbacks` inlined** (`sweeper.go`, `admin.go`) — the helper was a one-liner wrapping `append([]WalletService{primary}, fallbacks...)`. Now inlined into both `signingWallets()` methods. Functionally identical; removes one level of indirection. ✅
2. **Test file renamed** — `sweep_fallback_internal_test.go` → `sweeper_test.go`. Git reports `similarity index 100%`. Content-preserving rename to match package convention. ✅
3. **Comment trimming** (`config.go:856`, `sweep.go:21-23,189-190`, `tx_builder.go:63-68`, `sweeper.go:715-719`) — shortened verbose doc comments. No semantic change. ✅

### Verdict

No protocol, security, or correctness concerns. Clean cosmetic pass addressing prior review feedback.

**Approving the refactor commit.** Prior reviews' requirement for human sign-off on the protocol-critical sweep-fallback logic still applies to the PR as a whole.

🤖 *Arkana — automated reviewer · aggressive mode*

ghost (changes requested):
## 🔍 Arkana Code Review — arkd#1101

**Reviewer**: Arkana (aggressive review, protocol-critical)
**Verdict**: Request changes (1 bug, 2 design concerns, protocol-critical → needs human sign-off)

---

### ⚠️ PROTOCOL-CRITICAL

This PR modifies sweep transaction building and signing — the mechanism that recovers expired VTXO batch outputs on-chain. Bugs here mean funds stuck forever or swept to wrong addresses. **Human review required regardless of code quality.**

---

### 🐛 BUG — Return-value order swap on `BuildSweepTx` is a silent foot-gun

**`internal/core/ports/tx_builder.go:63`** (old) → new interface

```go
// OLD: BuildSweepTx(inputs []TxInput) (txid string, signedSweepTx string, err error)
// NEW: BuildSweepTx(inputs []TxInput) (unsignedTx string, txid string, err error)
```

The first two returns swapped AND the semantics changed (signed → unsigned). Both are `string`, so the Go compiler will not catch any caller that destructures in the old order. All callers *in this PR* are updated correctly — I verified each one:

- `sweeper.go` broadcast path: `sweepTxId, sweepTx, err = buildAndSignSweepTx(...)` ✓ (buildAndSignSweepTx returns `txid, signed`)
- `sweeper.go` graceful-degradation fallback: `sweepTx, sweepTxId, err = s.builder.BuildSweepTx(...)` ✓ (reversed order matches new `unsignedTx, txid`)
- `admin.go` Sweep: `txid, txhex, err = buildAndSignSweepTx(...)` ✓
- `sweeper.go` checkpoint: `_, sweepTx, err := buildAndSignSweepTx(...)` ✓

No external consumers exist (confirmed by cross-repo search of all 140+ repos).

**However**: the asymmetry between `buildAndSignSweepTx` returning `(txid, signedTx, err)` and `BuildSweepTx` returning `(unsignedTx, txid, err)` is a maintenance hazard. A future developer calling `BuildSweepTx` and destructuring as `txid, tx, err :=` (the natural reading) will silently get an unsigned PSBT in `txid` and the real txid in `tx`. This PR literally has both orderings in the same function (`createBatchSweepTask`).

**Recommendation**: Make the return order consistent. Either:
1. Change `BuildSweepTx` to return `(txid, unsignedTx, err)` to match `buildAndSignSweepTx`, or
2. Return a struct: `type SweepResult struct { UnsignedTx, Txid string }` — makes it impossible to confuse.

At minimum, if keeping the current order, add a loud comment at the interface definition warning about the order.

---

### 🟡 DESIGN — Graceful degradation stores unsigned PSBT in `SweepTxs`

**`internal/core/application/sweeper.go` (fallback in `createBatchSweepTask`)**

When all outputs are already spent and no wallet can sign the rebuilt sweep, the fallback calls `BuildSweepTx` which now returns an **unsigned PSBT base64** string. This gets stored in `round.SweepTxs[txid]` via `round.Sweep()` → `BatchSwept.Tx` → persisted to sqlite/postgres/badger.

Previously, `SweepTxs` always contained signed raw tx hex. Now it may contain unsigned PSBT base64. While I confirmed `GetBatchSweepTxs` (indexer) only reads the map keys (txids), the values are persisted and could be consumed by future code that assumes signed tx hex.

**Not blocking**, but worth documenting the invariant change, or storing an empty string for `Tx` when unsigned (since it's never broadcast in this path anyway).

---

### 🟡 DESIGN — Double `BuildSweepTx` in graceful-degradation path

**`internal/core/application/sweeper.go` (fallback path)**

```go
sweepTxId, sweepTx, err = buildAndSignSweepTx(s.builder, s.signingWallets(), outputsToSweep)
if err != nil {
    // ...
    sweepTx, sweepTxId, err = s.builder.BuildSweepTx(outputsToSweep)  // builds AGAIN
```

`buildAndSignSweepTx` internally calls `builder.BuildSweepTx` to build the unsigned tx, then tries signing. If signing fails, the outer code calls `BuildSweepTx` a second time. Both calls go through `buildSweepTransaction` → `wallet.DeriveAddresses` → NBXplorer `GetNewUnusedAddress`.

I verified that `GetNewUnusedAddress` returns the *next unused* address (not a freshly-derived one), so the two builds should produce the same txid. Still, this is wasteful and fragile — if the address derivation behavior ever changes, the txids would diverge silently.

**Recommendation**: Have `buildAndSignSweepTx` return the unsigned tx and txid on signing failure (alongside the error), so the caller can reuse them without rebuilding. Something like:

```go
func buildAndSignSweepTx(...) (txid, signedOrUnsigned string, signed bool, err error)
```

Or split the call in the fallback path: build once, then try signing, then use the already-built unsigned tx if signing fails.

---

### ✅ What looks good

1. **Build/sign split is clean.** `buildSweepTransaction` + `signSweepTransaction` is a solid separation. The tapscript index re-derivation from the PSBT (`len(in.TaprootLeafScript) > 0`) in `signSweepTransaction` is correct and eliminates index-threading through the API.

2. **Fallback iteration in `buildAndSignSweepTx`** is well-structured: build once, try each wallet, short-circuit on first success, aggregate errors with wallet index for debuggability.

3. **Test coverage is solid.** `TestBuildAndSignSweepTx` covers all branches (primary succeeds, fallback succeeds, all fail, no wallets, build error). `TestSignSweepTransaction_DerivesTapscriptIndexes` validates the re-derivation logic.

4. **Broadcasting stays on primary wallet** — correct, since `BroadcastTransaction` only relays raw bytes.

5. **No cross-repo breakage.** `TxBuilder` interface is internal to arkd. Confirmed no external implementors or consumers across all 140+ repos.

6. **No race conditions.** `signingWallets()` creates a fresh slice per call. Wallet list is read-only after init.

7. **`NewService` / `NewAdminService` signature changes** are correctly updated at all call sites (config.go and tests).

---

### 📋 Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🟠 Medium | `ports/tx_builder.go` | Return-value swap is a maintenance trap — make order consistent or use struct |
| 2 | 🟡 Low | `sweeper.go` (fallback) | Unsigned PSBT stored in `SweepTxs` — document or use empty string |
| 3 | 🟡 Low | `sweeper.go` (fallback) | Double `BuildSweepTx` build — reuse unsigned tx from first call |
| 4 | 🔴 Required | — | **Protocol-critical: human must review and approve** |

The architecture is sound. The build/sign split and fallback iteration are well-designed. Requesting changes for the return-order consistency issue (#1) which is a real hazard in protocol-critical code — the rest are suggestions. Once #1 is addressed and a human has reviewed, this is good to merge.

ghost (approved):
## Arkana Follow-Up Review — arkd#1101 (commit e39ff38a)

All three issues from my last review are addressed:

| # | Issue | Status |
|---|-------|--------|
| 1 | Return-value order swap (`BuildSweepTx` vs `buildAndSignSweepTx`) | ✅ Fixed — both now return `(txid, tx, err)` |
| 2 | Unsigned PSBT stored in `SweepTxs` | ✅ Fixed — `sweepTx = ""` on signing failure |
| 3 | Double `BuildSweepTx` in graceful-degradation path | ✅ Fixed — `buildAndSignSweepTx` returns txid on signing failure; no rebuild needed |

Verified all four call sites (`createBatchSweepTask` broadcast path, already-spent reconcile path, checkpoint sweep, `admin.Sweep`) destructure consistently as `(txid, tx, err)`.

The reconcile path correctly distinguishes build failure (`sweepTxId == ""` → fatal) from signing failure (`sweepTxId != ""` → log warning, continue with txid-only reconciliation). Clean.

### ⚠️ PROTOCOL-CRITICAL

This PR modifies sweep transaction building and signing. **Human review and approval still required before merge.**

🤖 *Arkana — automated reviewer · aggressive mode*

PR #1105, opened 2026-06-12
fix sweep connectors

louisinger:
@altafan please review

## Summary by CodeRabbit

* **Breaking Changes**
  * Updated wallet connector UTXO queries to accept multiple connector addresses in a single request (`connectorAddresses: string[]` with `minItems: 1`) instead of a single address (`connectorAddress`), including API/gRPC and internal service interfaces.

* **Improvements**
  * Enhanced sweep processing to fetch connector UTXOs in one bulk call when enabled.
  * Added a maximum input limit per sweep (capped at 1000); once reached, remaining outputs are deferred for later sweeps and a warning is logged.


ghost (changes requested):
## Code Review — arkade-os/arkd#1105 "fix sweep connectors"

Reviewed the full diff (17 files, +316/−97) and surrounding codebase. This is a solid fix for the standardness weight limit problem — chunking sweep inputs across multiple txs is the right approach. However, there are issues that need addressing before merge, including a critical change in broadcast semantics.

---

### 🔴 CRITICAL — Fire-and-forget broadcast loses error visibility

**`internal/core/application/admin.go` (new `broadcastSweepTx` + sweep loop)**

The old code broadcast synchronously and returned the txid **after confirmed relay**:
```go
txid, err = a.walletSvc.BroadcastTransaction(ctx, txhex)
if err != nil { return }
```

The new code fires goroutines and returns immediately:
```go
go a.broadcastSweepTx(broadcastCtx, txhex)
txids = append(txids, txid) // txid from BuildSweepTx, not from broadcast
```

**Consequences:**
1. The admin API returns a "success" response with txids/hexes, but the transactions may silently fail to broadcast (mempool rejection, network error, BIP68 non-finality).
2. Connector UTXOs are locked for 1 minute based on the assumption the sweep succeeded — if broadcast fails, those UTXOs are locked for nothing.
3. `saveBatchSweptEvents` runs concurrently and generates `BatchSwept` domain events for txids that may never reach the mempool — potentially marking VTXOs as swept when they aren't.
4. Unlike the autonomous sweeper in `sweeper.go` (which has a BIP68 retry loop at line ~646), the admin path now has **zero retry logic** for non-final txs.

**Recommendation:** Either:
- (A) Keep broadcast synchronous (per-chunk) and return partial results on failure, or
- (B) If you need fire-and-forget, collect broadcast results via a channel/errgroup, wait for all broadcasts before returning, and only include successfully-broadcast txids in the response.

The old synchronous behavior was safer for an admin-triggered sweep where the operator expects confirmation.

---

### 🔴 HIGH — `broadcastSweepTx` logging bugs

**`internal/core/application/admin.go:~586-592`**

```go
func (a *adminService) broadcastSweepTx(ctx context.Context, tx string) {
    txid, err := a.walletSvc.BroadcastTransaction(ctx, tx)
    if err != nil {
        log.Warnf("failed to broadcast sweep tx: %s", tx) // logs FULL RAW TX HEX, not the error
    }
    log.Infof("admin sweep transaction %s broadcasted", txid) // runs even on error; txid is ""
}
```

Two bugs:
1. On error, the format arg is `tx` (the full raw hex), not `err`. This dumps potentially hundreds of KB of hex into logs and hides the actual error.
2. The `Infof` line is not in an `else` — on failure it logs `"admin sweep transaction  broadcasted"` (empty txid).

Fix:
```go
if err != nil {
    log.WithError(err).Warn("failed to broadcast admin sweep tx")
    return
}
log.Infof("admin sweep transaction %s broadcasted", txid)
```

---

### 🟡 MEDIUM — Nondeterministic preconfirmed VTXO attachment

**`internal/core/application/admin.go:~918-950` (saveBatchSweptEvents)**

```go
preconfirmedAttached := false
for sweepTxid, leafVtxos := range leafVtxosBySweepTxid {
    preconfirmed := make([]domain.Outpoint, 0)
    if !preconfirmedAttached {
        preconfirmed = preconfirmedVtxos
        preconfirmedAttached = true
    }
    events, err := round.Sweep(leafVtxos, preconfirmed, sweepTxid, ...)
```

Go map iteration is random. Which sweep tx gets the preconfirmed vtxos attached is nondeterministic. The comment says this doesn't affect the fully-swept computation, but:
- It makes debugging/auditing harder — the same inputs produce different event structures across runs
- If event projections downstream do anything with preconfirmed vtxos per-sweep-tx (e.g., indexing), the behavior is inconsistent

Consider sorting the sweep txids before iterating, or always attaching preconfirmed vtxos to the **first** txid lexicographically.

---

### 🟡 MEDIUM — Empty leaf groups may enter `leafVtxosBySweepTxid`

**`internal/core/application/admin.go:~840-885`**

When an input has no leaf vtxos (e.g., it's a connector UTXO), `inputLeafVtxos` is empty but the append still creates a map entry:
```go
leafVtxosBySweepTxid[sweepTxid] = append(
    leafVtxosBySweepTxid[sweepTxid], inputLeafVtxos...,
)
```

If `inputLeafVtxos` is empty and this is the first entry for `sweepTxid`, this creates `leafVtxosBySweepTxid[sweepTxid] = []`. Then in the event loop, `round.Sweep([]domain.Outpoint{}, ...)` is called with empty leaf vtxos — generating a `BatchSwept` event with 0 leaves. This is wasteful and could interact poorly with the `fullySwept` computation if `countSweptLeafVtxos` is called cumulatively.

Guard: skip the append when `inputLeafVtxos` is empty, or skip sweep tx iterations where `len(leafVtxos) == 0`.

---

### 🟡 MEDIUM — No unit tests for `chunkSweepInputs` or `tapscriptWitnessSize`

**`internal/core/application/sweep_chunk.go`** — This is new, protocol-critical code. The chunking logic and witness size estimation are non-trivial and have edge cases:
- Single oversized input
- Exactly-at-budget inputs
- Empty input list
- Scripts with 0 pubkeys (handled with `numSigs = 0 → 1`, but why?)

Table-driven tests would catch regressions and document expected behavior. The `numSigs == 0` fallback in `tapscriptWitnessSize` (line ~128) deserves a comment explaining when a CSVMultisigClosure has 0 pubkeys.

---

### 🟢 LOW — `tapscriptWitnessSize` returns bytes, typed as `WeightUnit`

**`internal/core/application/sweep_chunk.go:~105-137`**

The function calculates **serialized witness byte count** and returns it as `lntypes.WeightUnit`. This is correct because witness bytes are 1 WU each in segwit, and `AddWitnessInput` expects witness WU. But it's worth a one-line comment clarifying that witness bytes == weight units.

---

### 🟢 INFO — Breaking proto API change

`SweepResponse`: `string txid/hex` → `repeated string txids/hexes`
`ListConnectorUtxosRequest`: `string connector_address` → `repeated string connector_addresses`

Field numbers are preserved but field names and types changed. This is a wire-breaking change. The `arkade-kotlin` repo has the old protos. Confirm no other consumers are affected and coordinate the kotlin proto update.

---

### ✅ What looks good

- **Chunking approach** (`sweep_chunk.go`): Greedy bin-packing with 5% margin below 400k WU is sound. The estimator-per-chunk pattern is correct despite the non-obvious mutation-then-discard in the overflow path.
- **Batching `ListConnectorUtxos`**: Moving from N gRPC calls to 1 call with N addresses is a good optimization.
- **`outpointToSweepTxid` mapping**: Clean way to attribute swept inputs to their sweep tx for event generation.
- **Cumulative `fullySwept` via `countSweptLeafVtxos`**: Correctly handles multi-tx sweeps of a single round.
- **Weight budget constant**: `380_000 WU` (95% of 400k) is a reasonable safety margin.

---

### ⚠️ Protocol-Critical Flag

This PR modifies sweep transaction construction (splitting inputs across txs), connector UTXO handling, and `BatchSwept` event generation. All of these are **protocol-critical paths** — bugs here can lock funds or misreport sweep status. **Requires human review and sign-off** regardless of code quality.

ghost (changes requested):
## 🔍 Arkana Code Review — arkd#1105 (sweep-connector)

**PR does two things:**
1. Batches `ListConnectorUtxos` to accept multiple addresses in a single RPC call (optimization)
2. Reverts the multi-tx sweep chunking from the first commit (`a01811f1`), going back to a single sweep tx

---

### ⚠️ Protocol-Critical — Human review required

This PR touches **sweep paths** — the mechanism that reclaims expired VTXOs and connector UTXOs. Bugs here lose real money. Flagging for mandatory human sign-off regardless of code quality.

---

### Findings

#### 1. 🔴 Reverting sweep chunking re-introduces the standardness limit risk

**`internal/core/application/admin.go`** — The first commit added `chunkSweepInputs` to split large sweep inputs across multiple transactions when they would exceed Bitcoin's 400k weight standardness limit. The latest commit (`a01811f1`) reverts all of that, going back to `BuildSweepTx(inputs)` with all inputs in a single tx.

If an ASP accumulates enough expired rounds + connectors, the single sweep tx will exceed 400kWU and be rejected by mempools as non-standard. This is a **silent failure mode** — the broadcast will fail and funds remain unswept.

**Question for author:** Was the chunking logic buggy and this is a temporary revert? If so, there should be a tracking issue. If the intent is to rely on `BuildSweepTx` to handle weight limits internally, that needs to be verified.

#### 2. 🟡 Proto field rename is a breaking change for JSON/REST consumers

**`api-spec/protobuf/arkwallet/v1/bitcoin_wallet.proto:347`** — Field 1 changes from `string connector_address` to `repeated string connector_addresses`. While this is **wire-compatible** for binary gRPC (wire type 2 → wire type 2), the JSON field name changes from `"connectorAddress"` to `"connectorAddresses"` and the type changes from string to array.

Any REST/JSON consumer of the wallet API will break. The OpenAPI spec is updated, but downstream clients (e.g. **arkade-kotlin** still has the old proto at `arkade/src/commonMain/proto/arkwallet/bitcoin_wallet.proto`) need coordinated updates.

#### 3. 🟡 No server-side validation for empty `connectorAddresses`

**`pkg/arkd-wallet/interface/grpc/handlers/wallet_handler.go:222`** — The handler passes `req.GetConnectorAddresses()` directly to the service. If the slice is empty, `ListConnectorUtxos` builds an empty address set and returns zero UTXOs silently. The OpenAPI spec declares `minItems: 1` but this isn't enforced server-side.

Consider adding:
```go
if len(req.GetConnectorAddresses()) == 0 {
    return nil, status.Error(codes.InvalidArgument, "at least one connector address required")
}
```

#### 4. 🟢 The `ListConnectorUtxos` batching is clean

**`pkg/arkd-wallet/core/application/wallet/service.go:337-341`** — Building `addressSet` as `map[string]struct{}` for O(1) membership checks is the right approach. The single `GetUtxos` call replaces N calls, reducing RPC overhead. Implementation is correct.

#### 5. 🟢 Mock and client updates are consistent

**`internal/infrastructure/tx-builder/covenantless/mocks_test.go`**, **`internal/infrastructure/wallet/wallet_client.go`** — Both correctly updated to the new `[]string` signature. No mismatches found.

---

### Cross-repo impact

| Repo | Impact | Status |
|------|--------|--------|
| **arkade-kotlin** | Has own copy of `bitcoin_wallet.proto` with old `connector_address` field | ❌ Needs update |
| **rust-sdk** | Uses `ark/v1/admin.proto` (sweep response reverted to singular, matches existing) | ✅ OK |
| **ts-sdk** | Uses indexer APIs only, not wallet APIs | ✅ OK |

---

### Verdict

Requesting changes for:
1. **Clarify the sweep chunking revert** — is this intentional long-term or a temporary rollback? If permanent, document why the standardness limit risk is acceptable. If temporary, open a tracking issue.
2. **Coordinate the arkade-kotlin proto update** for `ListConnectorUtxosRequest`.
3. **Add server-side validation** for empty `connectorAddresses`.

🤖 *Reviewed by Arkana*

ghost (approved):
## 🔍 Arkana Code Review — arkd#1105 (sweep-connector) — Follow-up

🤖 *Reviewed by Arkana (follow-up on new commits: `revert chunk sweep tx`, `truncate sweep inputs if necessary`, `lint`)*

---

### ✅ Critical issue resolved

The fire-and-forget broadcast I flagged previously has been **reverted**. Sweep is back to synchronous broadcast with proper error handling. Good call.

The new approach — batch `ListConnectorUtxos` + cap at `maxSweepInputs=1000` with truncation — is a much safer design. Remaining UTXOs are left for subsequent sweep calls instead of risking oversized transactions.

---

### Minor findings (non-blocking)

**1. OpenAPI `style: "simple"` invalid for query parameters**
`api-spec/openapi/swagger/arkwallet/v1/bitcoin_wallet.openapi.json:118`

Per OpenAPI 3.0 spec, `"simple"` style is only valid for `path` and `header` parameters. For `query` parameters, valid styles are `form` (default), `spaceDelimited`, `pipeDelimited`, or `deepObject`. This may cause issues with OpenAPI code generators. Should likely be removed (default `form` + `explode: true` handles array query params correctly) or explicitly set to `"form"`.

**2. Cross-repo proto sync needed: arkade-kotlin**
`arkade-kotlin/arkade/src/commonMain/proto/arkwallet/bitcoin_wallet.proto:342` still has the old `string connector_address = 1`. Will need updating to `repeated string connector_addresses = 1`. The wire format is backward-compatible (single string → repeated string at same field number), so this isn't urgent, but the proto copy should be updated to stay in sync.

**3. No input validation on empty `connectorAddresses`**
`pkg/arkd-wallet/core/application/wallet/service.go` — If `connectorAddresses` is empty, the address set will be empty and *every* UTXO will be filtered out (correct behavior). But `GetSweptRoundsConnectorAddress` could return an empty slice, which means the `ListConnectorUtxos` call does a full `GetUtxos` RPC to NBXplorer only to discard everything. A short-circuit `if len(connectorAddresses) == 0 { return nil, nil }` would avoid the wasted RPC.

---

### Truncation logic review: ✅ Correct

Verified all three truncation points:
- Connector UTXOs loop (`admin.go:~483`): caps and breaks correctly
- Commitment txid outer loop (`admin.go:~505`): guard at top of loop catches cap from connector phase
- Batch outputs inner loop (`admin.go:~556`): breaks inner + sweepableOutputs loops, partial `batchInputsList` is correctly saved to `batchInputs`

The `saveBatchSweptEvents` only receives inputs that were actually included in the sweep tx. ✅

### Verdict

Previous critical issue addressed. Approving — the remaining items are non-blocking. Nice fix. 🟢

⚠️ **Protocol note**: This touches sweep paths (connector UTXOs, round lifecycle). Flagging for @altafan human review per protocol-critical policy.

ghost (approved):
## 🔍 Arkana Code Review — arkd#1105 (sweep-connector) — Follow-up

🤖 *Reviewed by Arkana (follow-up on new commits: `revert chunk sweep tx`, `truncate sweep inputs if necessary`, `lint`)*

---

### ✅ Critical issue resolved

The fire-and-forget broadcast I flagged previously has been **reverted**. Sweep is back to synchronous broadcast with proper error handling. Good call.

The new approach — batch `ListConnectorUtxos` + cap at `maxSweepInputs=1000` with truncation — is a much safer design. Remaining UTXOs are left for subsequent sweep calls instead of risking oversized transactions.

---

### Minor findings (non-blocking)

**1. OpenAPI `style: "simple"` invalid for query parameters**
`api-spec/openapi/swagger/arkwallet/v1/bitcoin_wallet.openapi.json:118`

Per OpenAPI 3.0 spec, `"simple"` style is only valid for `path` and `header` parameters. For `query` parameters, valid styles are `form` (default), `spaceDelimited`, `pipeDelimited`, or `deepObject`. This may cause issues with OpenAPI code generators. Should likely be removed (default `form` + `explode: true` handles array query params correctly) or explicitly set to `"form"`.

**2. Cross-repo proto sync needed: arkade-kotlin**
`arkade-kotlin/arkade/src/commonMain/proto/arkwallet/bitcoin_wallet.proto:342` still has the old `string connector_address = 1`. Will need updating to `repeated string connector_addresses = 1`. The wire format is backward-compatible (single string → repeated string at same field number), so this isn't urgent, but the proto copy should be updated to stay in sync.

**3. No input validation on empty `connectorAddresses`**
`pkg/arkd-wallet/core/application/wallet/service.go` — If `connectorAddresses` is empty, the address set will be empty and *every* UTXO will be filtered out (correct behavior). But `GetSweptRoundsConnectorAddress` could return an empty slice, which means the `ListConnectorUtxos` call does a full `GetUtxos` RPC to NBXplorer only to discard everything. A short-circuit `if len(connectorAddresses) == 0 { return nil, nil }` would avoid the wasted RPC.

---

### Truncation logic review: ✅ Correct

Verified all three truncation points:
- Connector UTXOs loop (`admin.go:~483`): caps and breaks correctly
- Commitment txid outer loop (`admin.go:~505`): guard at top of loop catches cap from connector phase
- Batch outputs inner loop (`admin.go:~556`): breaks inner + sweepableOutputs loops, partial `batchInputsList` is correctly saved to `batchInputs`

The `saveBatchSweptEvents` only receives inputs that were actually included in the sweep tx. ✅

### Verdict

Previous critical issue addressed. Approving — the remaining items are non-blocking. Nice fix. 🟢

⚠️ **Protocol note**: This touches sweep paths (connector UTXOs, round lifecycle). Flagging for @altafan human review per protocol-critical policy.

altafan (approved):

PR #1106, opened 2026-06-13
client-lib: Handle sending digest header

altafan:
Make grpc client fetch, cache and send digest header + handle DIGEST_MISMATCH api error.

## Summary by CodeRabbit

* **New Features**
  * Automatic digest-aware retry/refresh for gRPC calls so requests encountering stale-digest errors are retried after a refresh.

* **Tests**
  * Updated gRPC client reconnection tests to reflect enhanced client initialization and lifecycle handling.

### DIFF
--- a/pkg/client-lib/client/grpc/client.go
+++ b/pkg/client-lib/client/grpc/client.go
@@ -14,6 +14,7 @@ import (
 	"github.com/arkade-os/arkd/pkg/client-lib/client"
 	"github.com/arkade-os/arkd/pkg/client-lib/internal/utils"
 	"github.com/arkade-os/arkd/pkg/client-lib/types"
+	"github.com/arkade-os/arkd/pkg/errors"
 	"github.com/btcsuite/btcd/wire"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
@@ -29,9 +30,12 @@ var testDialOptions []grpc.DialOption
 
 type grpcClient struct {
 	conn       *grpc.ClientConn
-	connMu     *sync.RWMutex
+	connMu     *sync.Mutex
+	svc        arkv1.ArkServiceClient
 	listenerMu *sync.RWMutex
 	listenerId string
+	infoMu     *sync.RWMutex
+	digest     string
 }
 
 func NewClient(serverUrl string) (client.Client, error) {
@@ -51,6 +55,12 @@ func NewClient(serverUrl string) (client.Client, error) {
 		serverUrl = fmt.Sprintf("%s:%d", serverUrl, port)
 	}
 
+	client := &grpcClient{
+		connMu:     &sync.Mutex{},
+		listenerMu: &sync.RWMutex{},
+		infoMu:     &sync.RWMutex{},
+	}
+
 	options := []grpc.DialOption{
 		grpc.WithTransportCredentials(creds),
 		grpc.WithDisableServiceConfig(),
@@ -65,10 +75,10 @@ func NewClient(serverUrl string) (client.Client, error) {
 			MinConnectTimeout: 3 * time.Second,
 		}),
 		grpc.WithChainUnaryInterceptor(
-			unaryVersionInterceptor(),
+			unaryVersionInterceptor(), unaryDigestInterceptor(client.getDigest),
 		),
 		grpc.WithChainStreamInterceptor(
-			streamVersionInterceptor(),
+			streamVersionInterceptor(), streamDigestInterceptor(client.getDigest),
 		),
 	}
 	options = append(options, testDialOptions...)
@@ -77,20 +87,15 @@ func NewClient(serverUrl string) (client.Client, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	client := &grpcClient{
-		conn:       conn,
-		connMu:     &sync.RWMutex{},
-		listenerMu: &sync.RWMutex{},
-		listenerId: "",
-	}
+	client.conn = conn
+	client.svc = arkv1.NewArkServiceClient(conn)
 
 	return client, nil
 }
 
 func (a *grpcClient) GetInfo(ctx context.Context) (*client.Info, error) {
 	req := &arkv1.GetInfoRequest{}
-	resp, err := a.svc().GetInfo(ctx, req)
+	resp, err := a.svc.GetInfo(ctx, req)
 	if err != nil {
 		return nil, err
 	}
@@ -122,6 +127,13 @@ func (a *grpcClient) GetInfo(ctx context.Context) (*client.Info, error) {
 			CutoffDate: s.GetCutoffDate(),
 		})
 	}
+	// Cache the latest digest under the lock. This must NOT wrap the RPC above:
+	// the digest interceptor reads a.digest under the same mutex, so holding it
+	// across the call would self-deadlock the RWMutex.
+	a.infoMu.Lock()
+	a.digest = resp.GetDigest()
+	a.infoMu.Unlock()
+
 	return &client.Info{
 		SignerPubKey:              resp.GetSignerPubkey(),
 		ForfeitPubKey:             resp.GetForfeitPubkey(),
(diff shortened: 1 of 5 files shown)


ghost (changes requested):
## Code Review — arkade-os/arkd#1106

Reviewed the full diff, cross-referenced the server-side digest interceptor, checked TS-SDK and .NET-SDK implementations for consistency, and examined the errors package. Findings below.

---

### 🔴 P0 — Unbounded retry loops (potential infinite loop)

**`client.go` — every `for {}` loop (RegisterIntent, DeleteIntent, EstimateIntentFee, SubmitTx, FinalizeTx, GetPendingTx, GetTransactionsStream.Connect)**

None of the retry loops have a max-retry guard. The only protection is the caller's `ctx` deadline — but not all callers set deadlines (SDK consumers may pass `context.Background()`).

Pathological case: if the server config is being updated rapidly (digest rotates between GetInfo response and the retried RPC), the client loops forever.

**Fix:** Add a max retry constant (e.g., 3) and bail with the original error after exhaustion.

```go
const maxDigestRetries = 3

// in each method:
for attempt := 0; ; attempt++ {
    resp, err := a.svc.RegisterIntent(ctx, req)
    if err != nil {
        if attempt < maxDigestRetries && strings.Contains(err.Error(), errors.DIGEST_MISMATCH.Name) {
            if _, infoErr := a.GetInfo(ctx); infoErr != nil {
                return "", err
            }
            continue
        }
        return "", err
    }
    return resp.GetIntentId(), nil
}
```

---

### 🔴 P0 — Silent retry may submit stale transactions after config change

**`client.go:SubmitTx` (~line 370), `FinalizeTx` (~line 392)**

DIGEST_MISMATCH means the server's config has changed (fees, dust limits, VTXO amounts, etc.). Silently retrying `SubmitTx` with the **same `signedArkTx`** that was constructed under the OLD config is dangerous — the tx may be invalid under the new config, or worse, may succeed with unexpected economic parameters.

The TS-SDK and .NET-SDK both **throw** on DIGEST_MISMATCH and let the caller rebuild the transaction under the new config. This client silently retries, which is a semantic divergence.

**Recommendation:** For `SubmitTx` and `FinalizeTx` specifically, consider surfacing the error (possibly as a typed error) so the caller can rebuild the transaction with fresh params. At minimum, document this design choice and the risk.

---

### 🟡 P1 — Inconsistent retry coverage across methods

**`client.go` — these ArkService methods have NO digest retry but ARE subject to the server-side digest guard:**

| Method | Risk | Has Retry? |
|--------|------|------------|
| `ConfirmRegistration` (~line 249) | Round lifecycle | ❌ |
| `SubmitTreeNonces` (~line 257) | **Signing protocol** | ❌ |
| `SubmitTreeSignatures` (~line 271) | **Signing protocol** | ❌ |
| `SubmitSignedForfeitTxs` (~line 286) | **Forfeit paths** | ❌ |
| `GetEventStream` Connect (~line 304) | Stream setup | ❌ |
| `GetEventStream` Reconnect (~line 308) | Stream reconnect | ❌ |
| `GetTransactionsStream` Reconnect (~line 475) | Stream reconnect | ❌ |
| `ModifyStreamTopics` (~line 598) | Topic update | ❌ |
| `OverwriteStreamTopics` (~line 622) | Topic update | ❌ |

If the digest changes mid-round, `SubmitTreeNonces`, `SubmitTreeSignatures`, and `SubmitSignedForfeitTxs` will fail with `DIGEST_MISMATCH` and the round/forfeit will be permanently lost.

Either all ArkService RPCs should retry, or none should (surface the error to the caller). The inconsistency is a footgun.

---

### 🟡 P1 — Copy-paste error in doc comment

**`digest_header.go:34`:**
```go
// streamVersionInterceptor returns a gRPC stream client interceptor that sets dogestHeader=version
```
Should be:
```go
// streamDigestInterceptor returns a gRPC stream client interceptor that sets digestHeader=digest
```

Two errors: wrong function name (`streamVersionInterceptor` → `streamDigestInterceptor`), and `dogestHeader=version` → `digestHeader=digest`.

---

### 🟡 P1 — No tests for the retry logic

The retry-on-DIGEST_MISMATCH is the core behavioral change in this PR. It should have at least:
- A test verifying that a single retry succeeds after GetInfo refreshes the digest
- A test verifying the loop terminates (max retries or context cancellation)
- A test verifying non-DIGEST_MISMATCH errors pass through immediately

---

### 🟢 Minor — Error matching via string contains is fragile

**`client.go` — all DIGEST_MISMATCH checks:**
```go
strings.Contains(err.Error(), errors.DIGEST_MISMATCH.Name)
```

This matches the error name string inside the serialized gRPC error. If the error message format changes, or if a future error's name contains "DIGEST_MISMATCH" as a substring, this breaks. Consider using the gRPC status code + structured error details instead, or at minimum the `CodeName()` method from the errors package if available on the client side.

---

### ✅ Good

- Correct use of `sync.RWMutex` for digest caching — readers (interceptor) take RLock, writer (GetInfo) takes Lock.
- Good comment explaining why the lock must NOT wrap the GetInfo RPC (self-deadlock prevention).
- Proper handling of `metadata.FromOutgoingContext` → Copy → Set → `NewOutgoingContext` in both interceptors (preserves existing metadata).
- `svc` field replacing the `svc()` method is cleaner — `ArkServiceClient` is safe for concurrent use, no need to create it on every call.
- Test updates are correct.

---

**Verdict:** Request changes. The unbounded loops and inconsistent retry coverage are risks that should be addressed before merge. The silent retry on `SubmitTx`/`FinalizeTx` needs a design decision documented.

⚠️ **Protocol-critical flag:** This PR touches round lifecycle paths (`ConfirmRegistration`, `SubmitTreeNonces`, `SubmitTreeSignatures`, `SubmitSignedForfeitTxs`) and forfeit/exit paths. Requires human sign-off even after code issues are resolved.

ghost (changes requested):
## Incremental Review — New Commits

Previous P0 (unbounded retry loops) is **fixed** ✅ — `withDigestRefresh` retries exactly once. Good.

Two remaining issues, one of which is a real bug:

---

### 🔴 P1 — Reconnect callbacks missing `withDigestRefresh` → infinite reconnect loop

**`client.go` — `GetEventStream.Reconnect` (~L289) and `GetTransactionsStream.Reconnect` (~L425)**

Both `Reconnect` callbacks call `a.svc.GetEventStream/GetTransactionsStream` directly, bypassing `withDigestRefresh`:

```go
Reconnect: func(ctx context.Context) (string, arkv1.ArkService_GetEventStreamClient, error) {
    stream, err := a.svc.GetEventStream(ctx, req)   // ← stale digest
    return "", stream, err
},
```

When the server restarts with a new config (rotating the digest), the reconnecting stream enters its inner retry loop (`stream_retry.go:315`). `ShouldReconnect` treats `FailedPrecondition` — which is `DIGEST_MISMATCH`'s gRPC code — as retryable (`reconnect.go:61-62`), so it retries with exponential backoff up to 10s. **But the digest is never refreshed**, so the loop retries with the same stale digest indefinitely until context cancellation.

**Fix:** Wrap the Reconnect calls with `withDigestRefresh`:

```go
Reconnect: func(ctx context.Context) (string, arkv1.ArkService_GetEventStreamClient, error) {
    stream, err := withDigestRefresh(a, ctx, func() (arkv1.ArkService_GetEventStreamClient, error) {
        return a.svc.GetEventStream(ctx, req)
    })
    return "", stream, err
},
```

Same for `GetTransactionsStream.Reconnect`.

---

### 🟡 P2 — Doc comment typo (copy-paste from version interceptor)

**`digest_header.go:38-39`**

```go
// streamVersionInterceptor returns a gRPC stream client interceptor that sets dogestHeader=version
```

Should be:
```go
// streamDigestInterceptor returns a gRPC stream client interceptor that sets digestHeader=digest
```

Two errors: wrong function name (`streamVersionInterceptor` → `streamDigestInterceptor`) and wrong description (`dogestHeader=version` → `digestHeader=digest`).

---

### ℹ️ Observation — `ModifyStreamTopics` / `OverwriteStreamTopics` not wrapped

These two methods call `a.svc.UpdateStreamTopics` directly without `withDigestRefresh`. They're not in a retry loop so this isn't a loop bug, but they will fail with `DIGEST_MISMATCH` after a server config change without auto-recovery. Up to you whether to wrap them — mentioning for completeness.

---

Everything else looks correct: mutex usage in `GetInfo` avoids self-deadlock, `getDigest()` correctly uses `RLock`, the generic `withDigestRefresh` is clean, and the first-call flow (empty digest → mismatch → refresh → retry) works by design since `GetInfo` is server-side exempt.

🤖 Arkana

PR #1109, opened 2026-06-14
fix(guards): validate version and digest headers when provided

Kukks:
## Problem

Both header-compatibility interceptors skip **all** validation when their "required" flag is false, so a client that *advertises* a bad value is accepted:

- **Version guard** (`version.go`): when `BuildVersionHeaderRequired` is false, an `x-build-version` below the configured minimum was accepted.
- **Digest guard** (`digest.go`): when `DigestHeaderRequired` is false, a wrong `x-digest` was accepted (the comparison was gated behind `guardEnabled`).

This conflates two independent questions:

- **Is the header mandatory?** → the `*Required` flag
- **If a value is provided, is it valid?** → should always be enforced

Observed: with `MIN_BUILD_VERSION_HEADER=0.9.10` but `MIN_BUILD_VERSION_HEADER_REQUIRED` unset (default false), a client sending `x-build-version: 0.9.9` was not rejected.

## Change

The `*Required` flag now governs only the **absent / empty** header case. A **present** value is always validated, regardless of the flag.

### Version guard
`RequireHeader` gates only absent/empty/unparseable headers. A present, parseable version is always held to the minimum — still a **minimum** check (strict `<` via `isBehind`): `>= min` passes, only `< min` is rejected.

| | header absent | present & `< min` | present & `>= min` | present & unparseable |
|---|---|---|---|---|
| required = true | reject | reject | pass | reject |
| required = false (before) | pass | **pass** | pass | pass |
| required = false (after) | pass | **reject** | pass | pass |

Unparseable-when-not-required stays allowed: an unparseable string can't be proven below the minimum.

### Digest guard
A present digest is always compared to the expected one; an absent/empty header is rejected only when `DigestHeaderRequired` is true. Expressed as: reject iff `digest != expected && (guardEnabled || digest != "")`.

| | header absent | present & matches | present & mismatches |
|---|---|---|---|
| required = true | reject* | pass | reject |
| required = false (before) | pass | pass | **pass** |
| required = false (after) | pass | pass | **reject** |

\* except when no digest is configured (empty expected): a missing header still passes, since there is nothing to match. GetInfo remains exempt (it's how clients learn the digest).

## Tests

- `version_test.go`: the former "not required: below-min client passes" case now expects rejection; added "not required: equal/above min passes" cases.
- `digest_test.go`: the former "disabled guard passes despite mismatch" case now expects rejection; added "disabled guard passes with matching digest / empty header" cases. The empty-expected-digest and GetInfo-exemption cases are unchanged.

## Verification

⚠️ **Not run locally** — no Go toolchain in this environment, and `ci_unit` only triggers on `pull_request` to base branches matching `'*'`, which does not match the slashed base `feat/deprecated-keys`, so unit CI will not run on this PR. Both changes were verified by manually tracing every case in `version_test.go` and `digest_test.go` against the new logic. **Please run `go test ./internal/interface/grpc/interceptors/...` before merging** — that's the claim most likely to need confirmation.

## Notes / out of scope

- `VersionGuard` still names its parsed threshold fields `serverMajor/serverMinor/serverPatch` and a comment calls it "the server's own build version," but the value is actually the configured minimum (`settings.BuildVersionHeader`). Left as-is to keep the diff minimal; renaming is a reasonable follow-up.

### DIFF
--- a/internal/interface/grpc/interceptors/digest.go
+++ b/internal/interface/grpc/interceptors/digest.go
@@ -33,15 +33,17 @@ func unaryDigestHandler(getDigest func() (string, bool, error)) grpc.UnaryServer
 					"failed to verify digest header, retry later",
 				)
 			}
-			if guardEnabled {
-				digest := digestHeaderValue(ctx)
-				if digest != expectedDigest {
-					return nil, errors.DIGEST_MISMATCH.
-						New("invalid digest header").WithMetadata(errors.DigestMetadata{
-						ExpectedDigest: expectedDigest,
-						GotDigest:      digest,
-					})
-				}
+			// A present digest is always validated against the expected one;
+			// an absent (empty) digest is rejected only when the header is
+			// required. guardEnabled therefore governs only the missing-header
+			// case, not whether a provided digest is checked.
+			digest := digestHeaderValue(ctx)
+			if digest != expectedDigest && (guardEnabled || digest != "") {
+				return nil, errors.DIGEST_MISMATCH.
+					New("invalid digest header").WithMetadata(errors.DigestMetadata{
+					ExpectedDigest: expectedDigest,
+					GotDigest:      digest,
+				})
 			}
 		}
 		return handler(ctx, req)
@@ -60,15 +62,17 @@ func streamDigestHandler(
 			if err != nil {
 				return errors.INTERNAL_ERROR.New("failed to verify digest header, retry later")
 			}
-			if guardEnabled {
-				digest := digestHeaderValue(ss.Context())
-				if digest != expectedDigest {
-					return errors.DIGEST_MISMATCH.
-						New("invalid digest header").WithMetadata(errors.DigestMetadata{
-						ExpectedDigest: expectedDigest,
-						GotDigest:      digest,
-					})
-				}
+			// A present digest is always validated against the expected one;
+			// an absent (empty) digest is rejected only when the header is
+			// required. guardEnabled therefore governs only the missing-header
+			// case, not whether a provided digest is checked.
+			digest := digestHeaderValue(ss.Context())
+			if digest != expectedDigest && (guardEnabled || digest != "") {
+				return errors.DIGEST_MISMATCH.
+					New("invalid digest header").WithMetadata(errors.DigestMetadata{
+					ExpectedDigest: expectedDigest,
+					GotDigest:      digest,
+				})
 			}
 		}
 		return handler(srv, ss)
--- a/internal/interface/grpc/interceptors/digest_test.go
+++ b/internal/interface/grpc/interceptors/digest_test.go
@@ -86,18 +86,34 @@ func TestDigestCompat(t *testing.T) {
 			wantGot:        "abc123",
 		},
 
-		// --- Guard disabled: every request passes regardless of the header ---
+		// --- Guard disabled: a missing header passes, but a present digest is
+		// still validated against the expected one ---
 		{
-			description:    "disabled guard passes despite mismatch",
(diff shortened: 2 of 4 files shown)


ghost (approved):
## Code Review — arkana 🔍

**Verdict: Approve** ✅

Clean, well-reasoned fix. The logic change is correct: `RequireHeader` should only govern absent/empty/unparseable headers, not act as a kill-switch for the entire guard.

### What I checked

- **Logic correctness**: Traced every path through the refactored `checkVersionCompat`. The behavior matrix in the PR description accurately matches the code. The only cell that changes (`required=false` + `present & < min`) is the intended fix.
- **Test coverage**: All 6 cells in the matrix are covered. New cases for `not required: below-min rejected`, `equal to min passes`, and `above min passes` lock the semantics. Existing cases for missing/empty/unparseable are preserved.
- **Cross-repo impact**: Checked rust-sdk (`ark-grpc/src/client.rs:120`) and dotnet-sdk (`NArk.Core/Transport/ArkdVersion.cs`) — both send `x-build-version` with their target version. This fix only rejects clients that explicitly advertise a version below the minimum, which is the correct behavior. No breakage.
- **Admin/health bypass**: `TestVersionCompatSkipsAdminEndpoints` confirms non-ArkService methods are still unguarded — unchanged by this PR.
- **Unparseable-when-not-required passes**: Correct design choice, well-justified — an unparseable string can't be proven below the minimum.

### Minor note (non-blocking)

- `docs/version-compat.md` is now stale — it says "Malformed version string → Request allowed" and "Only the major version component is compared" (the guard actually compares major+minor+patch via `isBehind`). This predates the PR but the gap is now wider. Worth a follow-up to sync the docs.
- `BuilldVersion` typo (double L) in the struct field — pre-existing, not this PR.

### CI note

As the author notes, CI won't trigger on this base branch. **Please run `go test ./internal/interface/grpc/interceptors/...` locally before merging.**

LGTM.

ghost (approved):
## Code Review — arkana 🔍

**Verdict: Approve** ✅

Clean, well-reasoned fix. The logic change is correct: `RequireHeader` should only govern absent/empty/unparseable headers, not act as a kill-switch for the entire guard.

### What I checked

- **Logic correctness**: Traced every path through the refactored `checkVersionCompat`. The behavior matrix in the PR description accurately matches the code. The only cell that changes (`required=false` + `present & < min`) is the intended fix.
- **Test coverage**: All 6 cells in the matrix are covered. New cases for `not required: below-min rejected`, `equal to min passes`, and `above min passes` lock the semantics. Existing cases for missing/empty/unparseable are preserved.
- **Cross-repo impact**: Checked rust-sdk (`ark-grpc/src/client.rs:120`) and dotnet-sdk (`NArk.Core/Transport/ArkdVersion.cs`) — both send `x-build-version` with their target version. This fix only rejects clients that explicitly advertise a version below the minimum, which is the correct behavior. No breakage.
- **Admin/health bypass**: `TestVersionCompatSkipsAdminEndpoints` confirms non-ArkService methods are still unguarded — unchanged by this PR.
- **Unparseable-when-not-required passes**: Correct design choice, well-justified — an unparseable string can't be proven below the minimum.

### Minor note (non-blocking)

- `docs/version-compat.md` is now stale — it says "Malformed version string → Request allowed" and "Only the major version component is compared" (the guard actually compares major+minor+patch via `isBehind`). This predates the PR but the gap is now wider. Worth a follow-up to sync the docs.
- `BuilldVersion` typo (double L) in the struct field — pre-existing, not this PR.

### CI note

As the author notes, CI won't trigger on this base branch. **Please run `go test ./internal/interface/grpc/interceptors/...` locally before merging.**

LGTM.

PR #1110, opened 2026-06-14
Sign forfeit txs at collection-time

Kukks:
## Summary

Today arkd stores forfeit transactions carrying **only the user's signature** and adds the operator's signature lazily, at fraud‑broadcast time (`broadcastForfeitTx`, `fraud.go:167`). If the operator's signer were unavailable when fraud must be punished, the forfeit could not be broadcast. (The doc comment at `fraud.go:29` already *claims* forfeits are "created and signed during the batch execution" — this PR makes that true.)

This PR makes forfeit txs **broadcast-ready at rest**: arkd signs each forfeit at collection time, so every *future* forfeit is stored operator-signed.

**Nothing has to be migrated.** `broadcastForfeitTx` decides per forfeit whether the stored PSBT already carries every signature it needs and signs only when it does not, so forfeits written before this change keep working untouched. There is no on-boot migration, no migration-status table and no tool to run.

A standalone backfill was part of an earlier revision and has been dropped on review, along with the `RoundRepository.PatchForfeitTxs` method and `Config.RepoManager()` accessor that existed only to serve it. It is kept [in a gist](https://gist.github.com/bitcoin-coder-bob/f16d9f57144d356d1b2e508c152461c7) should anyone ever want to patch stored forfeits by hand; the gist also carries the one `UPDATE` it performed.

### 1. Sign forfeits on collection (`internal/core/application`)
- New `signForfeitTxs` helper signs each collected forfeit PSBT via `signer.SignTransactionTapscript` in the finalization path, before persisting.
- Signing only adds witness data, so the **txid is unchanged** — no downstream index churn.
- A signing error **fails the round**, consistent with the adjacent boarding-input signing.

### 2. Don't double-sign at fraud time (`internal/core/application/fraud.go`)
- Because forfeits are now operator-signed at collection, re-signing them at fraud-broadcast time would append a **second identical operator signature**, producing an invalid PSBT (**duplicate key**) that fails to finalize — caught by `TestReactToFraud` in e2e.
- `broadcastForfeitTx` now signs **only when a signature is still missing**, via `domain.ForfeitTxReadyToBroadcast`. Pre-signed forfeits are finalized directly.
- Readiness is read off the psbt's **own leaf** (a sig for every pubkey the leaf commits to, plus the connector key spend sig), not off the operator's current key set. A forfeit signed before a signer-key rotation therefore stays recognized as ready instead of being re-signed with a key its leaf does not accept, which is the property collection-time signing exists to provide.

### 3. Reject forfeits that arrive carrying operator signatures (`service.go`)
- `VerifyForfeitTxs` only asserts `len(TaprootScriptSpendSig) > 0` on the vtxo input, and its txid rebuild ignores witness data, so a client could submit a forfeit carrying a `TaprootScriptSpendSig` under the **operator's** x-only key over the real leaf. Both the key and the leaf are known to the client.
- Collection-time signing would then append a second entry for that same (key, leaf) pair, the psbt would stop parsing, and `signForfeitTxs` would **fail the whole round for every participant** — with no conviction recorded against the submitter.
- `SubmitForfeitTxs` now rejects a forfeit carrying an operator-key tapscript sig, or any key spend sig (the connector is a wallet output a client cannot sign). It fails closed when no operator key is configured, since `Settings.SignerPubkey` is nillable and an empty key set would compare against nothing.

## Test plan

TDD throughout (test written and watched fail before each implementation).

- **Unit** — `signForfeitTxs` (signs each forfeit, propagates signer errors); the submission guard (rejects planted operator/key-spend sigs, fails closed with no operator key); `ForfeitTxReadyToBroadcast` (ready only when every leaf pubkey has signed and the connector is signed, still ready when the signing key is no longer current);.
- **e2e (existing, run in CI)** — collection-time signing is on the critical path of *every* settlement (the round fails if it breaks), so the whole settle-based suite gates on it. `TestReactToFraud` is the targeted guard: it settles, forfeits, unrolls the forfeited vtxo on-chain, and asserts the server broadcasts the (now pre-signed) forfeit and claims the vtxo — and is what surfaced the duplicate-key bug fixed in change #2. `TestEagerForfeitSurvivesWalletRotation` settles, forfeits, rotates to a new signer key with **no deprecated key retained**, then unrolls the forfeited vtxo and asserts the server still punishes the fraud.

### Local verification
- `db` (sqlite+postgres), no race: ✅
- `db` (sqlite) + all other `./internal/...` packages under `-race`: ✅ (no data races)
- `internal/core/application`: ✅
- whole module builds (linux): ✅; `go vet`: ✅; `golangci-lint`: ✅
- e2e on regtest: `TestReactToFraud` ✅, `TestEagerForfeitSurvivesWalletRotation` ✅. During the fraud window the signer records **zero** `SignTransactionTapscript` calls, against one before change #2, confirming the stored forfeit is broadcast without the signer.

## Notes for reviewers
- Forfeit txs are operator-internal (no client/indexer API exposes them), so there is intentionally no new endpoint to assert forfeit-signing state directly in e2e.
- `internal/core/domain` now imports `btcutil/psbt` for the first time, for the shared readiness helper. If that layer should stay clear of transaction encoding, `pkg/ark-lib/script` next to `FinalizeVtxoScript` is the natural home, at the cost of a cross-module change.
- `SubmitForfeitTxs` now reads settings on every call and rejects if they are unavailable — a new failure mode on a hot path.

## Release Notes

* **New Features**
  * Added a new forfeit transaction backfill tool that signs historical forfeit transactions missing the operator’s signature.

* **Improvements / Bug Fixes**
  * Enhanced forfeit signing and broadcasting to detect when the operator’s tapscript signature is already present, preventing duplicate-signature PSBT invalidation and skipping unnecessary signing.

* **Infrastructure**
  * Added database support to patch stored forfeit transaction data during backfill operations.


ghost (changes requested):
## 🔍 Arkana Protocol Review — PR #1110

**Verdict: Request changes (protocol-critical — requires human sign-off)**

This PR changes the forfeit transaction signing path, which is the mechanism for punishing fraud. The overall design is sound: eager signing at collection time makes forfeits broadcast-ready at rest, eliminating signer availability as a single point of failure at fraud-reaction time. The backward-compatible approach (fraud.go re-sign is harmless due to `XOnlyPubKey` keying in `script/finalizer.go:41`) is well-reasoned.

Test coverage is solid — unit tests for `signForfeitTxs`, backfill core logic (signs unsigned, skips already-signed, idempotent re-runs, signer failure counting), and repo-level `PatchForfeitTxs` tests on sqlite+postgres. Good.

No cross-repo breakage: `RoundRepository` is arkd-internal, no SDK or consumer repo implements or imports it. `ForfeitTx` references in `emulator` and `introspector-review` are test-only.

---

### Issues

#### 1. 🔴 `UpdateForfeitTx` silently succeeds on non-existent txid (all SQL backends)
**`internal/infrastructure/db/postgres/sqlc/query.sql:414`** and **`sqlite/sqlc/query.sql:422`**

```sql
UPDATE tx SET tx = @tx WHERE txid = @txid AND type = 'forfeit';
```

This is `:exec` — `ExecContext` returns `nil` error even when 0 rows are affected. If a txid doesn't match (e.g., data corruption, a race where the round was deleted between scan and patch), the backfill silently counts it as signed (`res.Signed++` at `backfill.go:131`) but nothing was persisted. 

The `PatchCollectedFees` precedent has the same gap, but for forfeit txs this is protocol-critical: an operator who runs the backfill and sees `signed=50 failed=0` would believe their forfeits are broadcast-ready when they aren't.

**Fix:** Either change the sqlc annotation to `:execresult` and check `RowsAffected() == 0`, or add a verification read-back. The badger implementation is fine (Upsert always writes).

#### 2. 🟡 `GetAllVtxos` loads entire vtxo set into memory (`backfill.go:61`)

```go
allVtxos, err := vtxos.GetAllVtxos(ctx)
```

A production operator may have millions of vtxos. This loads all of them into memory just to filter down to unswept forfeited ones. For an offline tool this isn't a blocker, but it should be documented (or a filtered query added).

**Suggestion:** Add a comment noting the memory implications, or consider adding a `GetForfeitableVtxos` query that filters at the DB level.

#### 3. 🟡 SQLite concurrent access with a running arkd (`cmd/arkd-forfeit-backfill/main.go`)

The backfill tool opens the same SQLite database file as a running arkd process. SQLite is opened without WAL mode (`internal/infrastructure/db/sqlite/utils.go`), and the retry budget is only 5 × 100ms = 500ms. If arkd is finalizing a round while the backfill writes, `PatchForfeitTxs` will hit "database is locked" errors.

This is safe (idempotent re-run), but **operators should be warned**. The tool's doc comment and README should state: "For SQLite deployments, stop arkd or ensure low write activity before running."

#### 4. 🟡 No `ctx.Done()` check in backfill loops (`backfill.go:77-132`)

The `Run` function iterates all commitment groups and all vtxos within each group without checking context cancellation. For a large backfill (thousands of forfeits), ctrl+C won't interrupt cleanly — the signer call will eventually timeout, but the loop won't break.

```go
for commitmentTxid, group := range byCommitment {
    // Should check: if ctx.Err() != nil { return res, ctx.Err() }
```

#### 5. ℹ️ Minor: `findForfeitTx` is duplicated (`backfill.go:145` vs `fraud.go`)

The PR description notes this is intentional to avoid a dependency on the application package. That's a reasonable tradeoff, but the two implementations should stay in sync. Consider adding a comment in both locations referencing each other.

---

### What looks correct

- **`signForfeitTxs` (`service.go:2055-2075`)**: Correctly signs each forfeit tx via `SignTransactionTapscript` with `nil` input indexes (matching the fraud.go pattern), parses the signed PSBT to extract the unchanged txid, and wraps it as `domain.ForfeitTx`. Error on any single forfeit correctly fails the entire round — right call for collection-time signing.

- **`fraud.go` left unchanged**: The broadcast-time re-sign is harmless because `script/finalizer.go:41` keys witness args by `XOnlyPubKey`, so a duplicate operator signature is a no-op. Verified in source.

- **Backfill idempotency**: `forfeitOperatorSigned` checks for existing tapscript sigs matching the operator's `XOnlyPubKey` before calling the signer. Re-runs are safe. The `fakeRounds.PatchForfeitTxs` test double correctly applies patches to stored rounds so re-run assertions work.

- **DB implementations**: Postgres and SQLite use `execTx` for transactional writes. Badger uses direct `Upsert` on the `Tx` record (correct — forfeits are stored as standalone `Tx` entries, not embedded in the round document).

- **Build scripts**: `build-arkd-forfeit-backfill` follows the existing `build-arkd` pattern.

---

### Protocol-critical flag 🚨

This PR modifies the forfeit signing path — the operator's ability to punish fraud. Even though the code looks correct, **a human must review and approve** per protocol rules. Key points for human reviewer:

1. Verify that `SignTransactionTapscript(ctx, tx, nil)` at collection time signs the same inputs it would at broadcast time in `fraud.go:167`.
2. Verify that the backfill's `forfeitOperatorSigned` check correctly identifies the operator's signature (not a user signature that happens to be tapscript).
3. Confirm the silent-UPDATE issue (#1 above) is acceptable or should be fixed before merge.

ghost (commented):
### 🔍 Arkana — Incremental Review (new commit a71e6d5f)

**Commit:** `backfill: wrap long signatures to satisfy golines`

Purely cosmetic — line wrapping on function signatures and log statements in `internal/backfill/backfill.go`. No logic changes. LGTM on this commit.

**Previous findings still outstanding** — my earlier review's issues (#1 silent UPDATE on non-existent txid, #2 memory load, #3 SQLite concurrency warning, #4 missing ctx.Done check) remain unaddressed. The CHANGES_REQUESTED state stands until at least #1 is resolved or explicitly acknowledged as acceptable risk.

ghost (commented):
### 🔍 Arkana — Incremental Review (new commit dcbc0cea)

**Commit:** `fraud: skip re-signing already-signed forfeit txs to avoid duplicate-key PSBT`

**Good catch.** This fixes a real bug introduced by the collection-time signing in this same PR: `broadcastForfeitTx` unconditionally called `SignTransactionTapscript`, which for already-signed forfeits would append a duplicate operator signature → invalid PSBT → failed finalization → **forfeit can't broadcast** → fraud goes unpunished. That's a critical correctness issue and this commit correctly addresses it.

**Review of the fix:**

- `forfeitTxOperatorSigned` (`fraud.go:433-444`) — logic is correct: scans all inputs' `TaprootScriptSpendSig` for the operator's x-only pubkey. This matches the check in `backfill.go:forfeitOperatorSigned` (identical logic, different function name). ✅
- The `GetPubkey` call at `fraud.go:176` adds one signer RPC per fraud reaction. Fraud is rare; negligible overhead. ✅
- Conditional signing at `fraud.go:179-183` — correctly preserves the original code path for legacy unsigned forfeits while skipping for newly pre-signed ones. ✅
- Test coverage (`service_forfeit_test.go:99-131`) — covers signed, unsigned, and wrong-key cases. ✅

**One minor note:** `forfeitTxOperatorSigned` (fraud.go) and `forfeitOperatorSigned` (backfill.go) are identical logic with slightly different names. Consider extracting to a shared internal util to avoid drift, but this is non-blocking.

**Previous findings still outstanding** — issues #1 (silent UPDATE on non-existent txid), #2 (memory load), #3 (SQLite concurrency warning), #4 (missing ctx.Done check) from my initial review remain unaddressed. The CHANGES_REQUESTED state stands until at least #1 is resolved or explicitly acknowledged.

internal/core/application/fraud.go
```diff
@@ -164,9 +165,21 @@ func (s *service) broadcastForfeitTx(ctx context.Context, vtxo domain.Vtxo) erro
...
 	}
 
-	signedForfeitTx, err := s.signer.SignTransactionTapscript(ctx, forfeitTxB64, nil)
+	// Forfeit txs are signed by the operator at collection time, so the stored tx
+	// is usually already broadcast-ready. Re-signing would append a duplicate
+	// operator signature and produce an invalid PSBT (duplicate key), so we only
+	// sign here when the operator signature is still missing (e.g. forfeit txs
+	// collected before collection-time signing was introduced).
+	signedForfeitTx := forfeitTxB64
+	signerPubkey, err := s.signer.GetPubkey(ctx)
```
bitcoin-coder-bob: this does a grpc call, but the goal of the PR was "if the operator's signer were unavailable when fraud must be punished, the forfeit could not be broadcast." so if the signer is down, this will error and we will not broadcast. We need to pull the pubkey from a cache perhaps. So if signer is down when we need to broadcast a forfeit, we will fail here which directly goes against stated goal. Caching the pubkey and reading it from the cache here should fix this.

bitcoin-coder-bob (commented):
lgtm other than need a fixing of merge conflict and the one comment: https://github.com/arkade-os/arkd/pull/1110/changes#r3415820827

ghost (commented):
### 🔍 Arkana — Incremental Review (merge commit 97d5541d)

**Scope:** Single merge commit bringing `origin/master` into `presign-forfeit-txs`.

**Conflict resolution in `internal/infrastructure/db/postgres/sqlc/query.sql`:** ✅ Clean. The branch's `UpdateForfeitTx` query and master's `SelectAssetsWithUnspentAmountsByIds` query occupy distinct positions — both are preserved, ordering is correct.

**No PR code was modified** — the 18 changed files and +897/-10 delta are identical to my previous review. No new findings.

Previous review stance (request-changes for human sign-off on protocol-critical forfeit signing) still applies.

ghost (commented):
## Code Review

**Verdict: APPROVED** — This is a sound, well-structured fix to a real liveness risk in forfeit broadcasting. The core logic is correct, the duplicate-signature regression is cleanly handled, and the test coverage (unit, repo, e2e) is solid. Flagging two medium issues and some nits for the team.

---

### MEDIUM

**M1: Silent no-op on txid miss in `UpdateForfeitTx` (Postgres + SQLite)**

`UPDATE tx SET tx = $1 WHERE txid = $2 AND type = 'forfeit'` returns no error if zero rows are matched. If a txid is stale or wrong, `PatchForfeitTxs` succeeds, the backfill increments `res.Signed`, and no one knows the row was never written.

Suggested fix: check `RowsAffected()` after `ExecContext` and surface a warning or error on 0.

```go
result, err := q.db.ExecContext(ctx, updateForfeitTx, arg.Tx, arg.Txid)
if err != nil {
    return err
}
n, _ := result.RowsAffected()
if n == 0 {
    return fmt.Errorf("forfeit tx %s not found in db", arg.Txid)
}
```

**M2: `forfeitOperatorSigned` cross-input false-positive risk**

Both implementations iterate over all inputs looking for the operator's x-only key in any `TaprootScriptSpendSig`. A forfeit tx has exactly two inputs: the vtxo (index 0, operator-signed) and the connector (index 1). If a future format change moves the vtxo input, or if the connector tapscript happens to reference the operator's key (e.g. via key reuse in the connector branch), the check could return `true` for an actually-unsigned forfeit and skip re-signing, causing a failed broadcast.

Tightening to check only `ptx.Inputs[0].TaprootScriptSpendSig` would be strictly correct for the current forfeit design and would self-document the assumption.

---

### LOW / Nits

- **Badger `PatchForfeitTxs` uses `Upsert`, not `Update`**: SQL backends guard with `AND type='forfeit'`; badger doesn't distinguish by type in the key. Cosmetically inconsistent; practically safe since Bitcoin txids are globally unique.
- **`GetAllVtxos` full table scan**: acceptable for a one-shot backfill, but operators with a large vtxo set may see high memory usage. A targeted filtered query would be cleaner long-term.
- **No timeout on `context.Background()` in `main.go`**: a stuck signer gRPC call hangs the tool forever. Consider `signal.NotifyContext` or a deadline.
- **`forfeitTxOperatorSigned` is duplicated**: `fraud.go` and `backfill.go` have identical implementations. The local copy is justified to avoid the application import cycle. Fine for now; consider a shared `internal/forfeit` package if a third callsite appears.
- **Unquoted path vars in `build-arkd-forfeit-backfill`**: `pushd $PARENT_PATH` will break for paths with spaces. Cosmetic but inconsistent with shell best practices.
- **Test assertion reuse of `f3`**: `test_patch_forfeit_txs` uses the `f3` constant both as patched content for txida and as the expected-unchanged value of f3Tx, which is confusing on first read.

---

This touches forfeit broadcasting and round finalization — **protocol-critical paths**. The logic is correct and the e2e coverage via `TestReactToFraud` provides a strong functional gate. The medium issues are safety-net improvements, not correctness blockers for this PR. Human sign-off recommended before merge per policy.

ghost (commented):
## Code Review

**Verdict: APPROVED** — This is a sound, well-structured fix to a real liveness risk in forfeit broadcasting. The core logic is correct, the duplicate-signature regression is cleanly handled, and the test coverage (unit, repo, e2e) is solid. Flagging two medium issues and some nits for the team.

---

### MEDIUM

**M1: Silent no-op on txid miss in `UpdateForfeitTx` (Postgres + SQLite)**

`UPDATE tx SET tx = $1 WHERE txid = $2 AND type = 'forfeit'` returns no error if zero rows are matched. If a txid is stale or wrong, `PatchForfeitTxs` succeeds, the backfill increments `res.Signed`, and no one knows the row was never written.

Suggested fix: check `RowsAffected()` after `ExecContext` and surface a warning or error on 0.

```go
result, err := q.db.ExecContext(ctx, updateForfeitTx, arg.Tx, arg.Txid)
if err != nil {
    return err
}
n, _ := result.RowsAffected()
if n == 0 {
    return fmt.Errorf("forfeit tx %s not found in db", arg.Txid)
}
```

**M2: `forfeitOperatorSigned` cross-input false-positive risk**

Both implementations iterate over all inputs looking for the operator's x-only key in any `TaprootScriptSpendSig`. A forfeit tx has exactly two inputs: the vtxo (index 0, operator-signed) and the connector (index 1). If a future format change moves the vtxo input, or if the connector tapscript happens to reference the operator's key, the check could return `true` for an actually-unsigned forfeit and skip re-signing, causing a failed broadcast.

Tightening to check only `ptx.Inputs[0].TaprootScriptSpendSig` would be strictly correct for the current forfeit design and self-document the assumption.

---

### LOW / Nits

- **Badger `PatchForfeitTxs` uses `Upsert`, not `Update`**: SQL backends guard with `AND type='forfeit'`; badger doesn't distinguish by type in the key. Cosmetically inconsistent; practically safe since Bitcoin txids are globally unique.
- **`GetAllVtxos` full table scan**: acceptable for a one-shot backfill, but operators with a large vtxo set may see high memory usage. A targeted filtered query would be cleaner long-term.
- **No timeout on `context.Background()` in `main.go`**: a stuck signer gRPC call hangs the tool forever. Consider `signal.NotifyContext` or a deadline.
- **`forfeitTxOperatorSigned` is duplicated**: `fraud.go` and `backfill.go` have identical implementations. The local copy is justified to avoid the application import cycle. Fine for now; consider a shared `internal/forfeit` package if a third callsite appears.
- **Unquoted path vars in `build-arkd-forfeit-backfill`**: `pushd $PARENT_PATH` will break for paths with spaces.
- **Test assertion reuse of `f3`**: `test_patch_forfeit_txs` uses the `f3` constant both as patched content for txida and as the expected-unchanged value of f3Tx, which is confusing on first read.

---

This touches forfeit broadcasting and round finalization — **protocol-critical paths**. The logic is correct and the e2e coverage via `TestReactToFraud` provides a strong functional gate. The medium issues are safety-net improvements, not correctness blockers for this PR. Human sign-off recommended before merge per policy.

ghost (approved):
### 🔍 Arkana — Incremental Review (commits 16830b5f, 06aa0d8e)

**Verdict: APPROVED** — Both new commits are clean, well-tested fixes that directly address issues from the previous review.

---

#### Commit `16830b5f` — fraud: broadcast pre-signed forfeits without the live signer

This is the right fix. Reading the operator key set from the cached settings instead of the live signer eliminates the dependency on signer availability at fraud-broadcast time — which was the entire motivation for pre-signing at collection time.

**What I checked:**
- `fraud.go:179-194`: Settings cache is populated at service init (`service.go:128`), so `settings` is non-nil on the fraud path. The nil guard at `:183` is correct defensive code.
- `fraud.go:186-194`: Deprecated key inclusion prevents re-signing a forfeit whose tapscript references an old (rotated-out) key, which would produce an invalid witness. Logic mirrors the backfill path.
- `backfill.go:71-83`: Same deprecated-key pattern, sourced from the signer (correct for a standalone tool that doesn't have the settings cache).
- `forfeitTxOperatorSigned` signature updated consistently in both `fraud.go` and `backfill.go` — the duplication is justified to avoid an import cycle.
- Test coverage: `TestBackfillSkipsForfeitsSignedWithDeprecatedKey` and the updated `TestForfeitTxOperatorSigned` (multi-key set assertion at `:127`) both cover the new path.

**Edge case verified:** If `operatorKeys` is empty (settings.SignerPubkey nil, no deprecated keys), `forfeitTxOperatorSigned` returns `false`, and the code falls through to the live signer. This is correct — it degrades to the legacy path rather than silently skipping signing.

No issues found.

---

#### Commit `06aa0d8e` — fix(db): fail loudly when PatchForfeitTxs targets a missing txid

This directly addresses my M1 finding from the first review. Clean implementation across all three backends:

- **sqlite/postgres** (`sqlite/round_repo.go:634-643`, `postgres/round_repo.go:507-516`): `:exec` → `:execrows`, check `RowsAffected() == 0`. Correct.
- **badger** (`badger/ark_repo.go:109-115`): `Upsert` → `Update`, mapping `ErrNotFound` to a consistent error message. Correct — prevents inserting a stray record.
- **Tests**: `service_test.go:768-769` asserts the not-found path on sqlite/postgres; new `badger/ark_repo_test.go` covers badger specifically. Both verify the stray-record scenario doesn't occur.

No issues found.

---

Both previous MEDIUM findings are now resolved. The PR is in good shape. **Protocol-critical paths (forfeit broadcast, round finalization) — human sign-off still recommended before merge.**

Co-Reviewed-By: Arkana 🔍

Dunsin-cyber (commented):
utACK 06aa0d8e, read through the forfeit signing path, logic looks sound.

louisinger:
let's simplify the PR by removing the tool to backfill and sign old forfeits
let's keep it in some gist to do it manually via DB if necessary but the logic of checking forfeit signing or not should be enough to not be a breaking change

cc @bitcoin-coder-bob

bitcoin-coder-bob:
> let's simplify the PR by removing the tool to backfill and sign old forfeits let's keep it in some gist to do it manually via DB if necessary but the logic of checking forfeit signing or not should be enough to not be a breaking change
> 
> cc @bitcoin-coder-bob

backfill code removed in: [97e2ab1](https://github.com/arkade-os/arkd/pull/1110/commits/97e2ab16afbef6249f3dd72534675ec025046383)

Gist: https://gist.github.com/bitcoin-coder-bob/f16d9f57144d356d1b2e508c152461c7

louisinger (approved):

bitcoin-coder-bob:
> Still open — `UpdateForfeitTx` silently succeeds on no-match (protocol-critical)

is your comment on this issue out of date?

bitcoin-coder-bob:
> Still open — `UpdateForfeitTx` silently succeeds on no-match (protocol-critical)

is your comment on this issue out of date?

@arkana-ai-bot

PR #1111, opened 2026-06-14
Apply asset-query-churn fixes (#1043) on top of deprecated signer keys (#1097)

Kukks:
Stacks the conflict-resolved version of #1043 (stable asset reads under churn + SQLite read/write pool split) on top of #1097 (`feat/deprecated-keys`).

## What this is
- Base branch: `feat/deprecated-keys` (PR #1097), so this is a **stacked PR**.
- Brings in #1043's changes: single-query asset reads, separate SQLite read/write paths (WAL, pinned connections), and discarding interrupt-tainted connections.

## How it was produced
1. Merged current `master` into #1043's branch and resolved conflicts (sqlc query collisions, the settings/scheduled-session refactor that deleted `intent_fees_repo.go`/`scheduled_session_repo.go`, and the `testFeeRepository`/`testScheduledSessionRepository` removals).
2. Adapted master-side code written against the old `*sql.DB` API to #1043's new `SQLiteDB` read/write-pool API (`round_repo.go`, `vtxo_repo.go`, `service.go`, `settings_repo.go`, and the `settings_seed`/`vtxo_repo_batching` tests — the latter now uses `file::memory:` + shared cache so reads see writes across the split pools).
3. Merged that resolved branch onto `feat/deprecated-keys` — **clean, zero conflicts** (the two PRs touch disjoint subsystems).

## Testing
- `GOOS=linux go build ./...` and `GOOS=linux go vet ./...` (compiles all packages incl. tests): clean.
- Native run: `internal/infrastructure/db/sqlite` package tests and `TestService` (sqlite config, exercising the resolved asset/settings repository code): pass.

## Note
- A pre-existing failure in #1097 (`TestSignTransaction` nil `keyMgr` at `pkg/arkd-wallet/.../service.go:665`) reproduces identically on a clean `feat/deprecated-keys` checkout and is unrelated to these changes — left untouched.
- Because the base is a slashed branch (`feat/*`), repo CI workflows (triggered on `branches: ['*']`) won't run against this PR until it's rebased onto `master` or the parent merges.


ghost (changes requested):
## 🔍 Arkana Code Review — arkade-os/arkd#1111

**Scope:** SQLite read/write pool split, asset query optimization (N+1 → single JOIN), interrupt-tainted connection discard, client-side digest header mechanism, new e2e churn test.

**Verdict:** Well-engineered change. The core infrastructure (pool split, WAL, connection pinning/discard) is sound. A few items need attention before merge — one correctness issue in conviction_repo, one performance concern, and a comment typo.

---

### ⚠️ Issues to Fix

**1. `conviction_repo.go:45,79` — bare `==` comparison for `sql.ErrNoRows` will break under error wrapping**

```go
if err == sql.ErrNoRows {
```

With the new `withReadQuerier`, if `closeConn` also fails, errors get wrapped via `fmt.Errorf("%w: %w", err, closeErr)` and the bare `==` comparison silently misses the sentinel. The rest of the codebase (asset_repo, round_repo, vtxo_repo, settings_repo) was correctly updated to use `errors.Is(err, sql.ErrNoRows)` — conviction_repo was missed.

**Fix:** Replace both occurrences with `errors.Is(err, sql.ErrNoRows)`.

**2. `conviction_repo.go:105` — `withWriteQuerier` called per-conviction in a loop (non-transactional batch)**

```go
for _, conviction := range convictions {
    // ...
    if err := withWriteQuerier(ctx, r.db, func(q *queries.Queries) error {
        return q.UpsertConviction(ctx, params)
    }); err != nil {
```

Each iteration acquires a connection, runs one INSERT, releases. For N convictions, that's N connection round-trips. The old code had the same per-item pattern through the cached querier, but the connection overhead was lower. Consider wrapping the loop in a single `execTx` call for atomicity and performance. Not blocking, but worth noting — partial failures leave partial writes.

**3. `digest_header.go:34` — typo in comment**

```go
// streamVersionInterceptor returns a gRPC stream client interceptor that sets dogestHeader=version
```

`dogestHeader` → `digestHeader`

---

### 📝 Observations (non-blocking)

**4. `client.go:30` — `connMu *sync.Mutex` is now vestigial**

With the old `svc()` method removed and `svc` set once at construction, `connMu` is only locked in `Close()`. Nothing else acquires it. Consider removing it or documenting its purpose.

**5. `digest_header.go:55-57` — `isDigestMismatch` uses fragile string matching**

```go
func isDigestMismatch(err error) bool {
    return err != nil && strings.Contains(err.Error(), errors.DIGEST_MISMATCH.Name)
}
```

The existing TODO acknowledges this. When the structured error details path works end-to-end, this should check `status.Code(err) == codes.FailedPrecondition` + parse `status.Details()` for the specific error code. Not blocking since the TODO is tracked.

**6. `withDigestRefresh` — no deduplication of concurrent GetInfo calls**

When N concurrent RPCs all get `DIGEST_MISMATCH`, each independently calls `GetInfo`. The calls are idempotent and cheap, but a `singleflight.Group` would be cleaner. Not blocking.

**7. Metadata handling change is a correctness improvement**

Old code assigned metadata even on parse error (latent bug). New code only assigns on success, matching the stated intent in the comment. Good fix.

**8. Supply calculation correctness — verified**

The `SelectAssetsWithUnspentAmountsByIds` LEFT JOIN + client-side `big.Int.Add` accumulation produces the same result as the old N+1 `SelectAssetAmounts` approach. Edge case (asset with zero unspent VTXOs) handled correctly via COALESCE. Postgres and SQLite implementations have full semantic parity.

---

### ✅ What looks good

- **Pool split architecture**: WAL mode + separate read/write pools with `MaxOpenConns(1)` for writes is the standard SQLite concurrency pattern. Read pool correctly sized to `runtime.NumCPU()`.
- **Connection discard via `driver.ErrBadConn` + `Raw()`**: Standard Go idiom. The `closeConn` implementation correctly handles the `sql.ErrConnDone` follow-up.
- **`isInterruptError` detection**: Covers `context.Canceled`, `context.DeadlineExceeded`, and `sqlite3.SQLITE_INTERRUPT` (including masked codes). Thorough.
- **`withReadQuerier`/`withWriteQuerier` connection lifecycle**: All paths properly close connections. No leaks found.
- **`queries.New(*sql.Conn)` is valid**: `*sql.Conn` satisfies the sqlc-generated `DBTX` interface.
- **No cross-repo breakage**: `SQLiteDB` interface is in an `internal` package. `grpcClient` is unexported. Client-lib public API unchanged.
- **Test coverage**: `TestCanceledReadQueryDiscardsConnection`, `testAssetRepositorySpentOnlySupply`, `TestGetAssetQueryChurn` cover the key new code paths.

---

### 🔒 Protocol-Critical Flag

This PR modifies VTXO read/write paths (`vtxo_repo.go`), round lifecycle reads (`round_repo.go`), and forfeit tx retrieval (`GetRoundForfeitTxs`). The connection pooling change is sound under WAL mode, but **requires human sign-off** given the blast radius on protocol-critical data paths.

**Requesting changes for items 1–3 above.**

PR #1112, opened 2026-06-14
Deprecate keys with sqlite fix

altafan:


ghost (approved):
## Review: Deprecate keys with sqlite fix

**Verdict: Approve** — solid, well-tested read/write pool split. The SQLite concurrency model is correct (one writer, many readers via WAL). The supply query optimization and interrupt-aware connection handling are well-implemented. No protocol-critical concerns.

### Architecture ✅

The read/write pool separation (`SQLiteDB` interface with `Read()` / `Write()`) is the canonical approach for SQLite concurrency. WAL mode + busy timeout in production, shared-cache for in-memory test DBs — all correct. The `closeConn` + `driver.ErrBadConn` discard trick is the standard way to evict tainted connections from `database/sql`'s pool.

### Supply query change ✅

`SelectAssetsWithUnspentAmountsByIds` replaces the old N+1 pattern (`SelectAssetsByIds` + per-asset `SelectAssetSupply`/`SelectAssetAmounts`) with a single LEFT JOIN to `vtxo_vw`. Semantics are preserved: both filter `spent = false` and sum amounts. The additional `asset_amount > 0` filter is correct — vtxos without asset projections have `COALESCE(amount, 0)` which contributes nothing to supply. Go-side accumulation via `big.Int` matches the column types (INTEGER in SQLite, NUMERIC(20,0) in Postgres).

### Issues to address

**1. Bug — `conviction_repo.go:45,79`: bare `==` comparison with `sql.ErrNoRows`**

```go
// conviction_repo.go:45
if err == sql.ErrNoRows {
// conviction_repo.go:79
if err == sql.ErrNoRows {
```

These errors now flow through `withReadQuerier` which can wrap them via `fmt.Errorf("%w: %w", err, closeErr)` (utils.go:241). If `closeConn` ever returns a non-nil error alongside `sql.ErrNoRows`, the `==` check will miss it. The rest of the PR correctly uses `errors.Is(err, sql.ErrNoRows)` (asset_repo.go:179,195; round_repo.go:337,553,596; vtxo_repo.go:429,637; settings_repo.go:53). These two spots should be updated for consistency.

**2. Nit — `settings_seed_test.go`: read pool leaked in test cleanup**

All three sub-tests (`fresh install`, `backfill legacy`, `idempotent`) do:
```go
dbSvc, err := sqlitedb.OpenDb(":memory:")
db := dbSvc.Write()
t.Cleanup(func() { _ = db.Close() })
```

This closes only the write pool. `dbSvc.Close()` should be called instead to close both pools. (For bare `:memory:` without shared-cache the read pool is a separate empty database anyway, so this is harmless — but it's a resource leak and sets a bad pattern.)

**3. Observation — per-item `withWriteQuerier` in `conviction_repo.go:101-103`**

```go
for _, c := range convictions {
    // ...
    if err := withWriteQuerier(ctx, r.db, func(q *queries.Queries) error {
        return q.UpsertConviction(ctx, params)
    }); err != nil {
```

Each conviction acquires and releases a write connection individually. Pre-existing pattern, but now has more overhead with explicit connection pinning. Consider batching into `execTx` if conviction counts are large. Not a blocker.

### What's good

- Comprehensive E2E test (`TestGetAssetQueryChurn`) — stress-tests cancelled `GetAsset` queries concurrent with offchain sends and periodic batches. This directly validates the interrupt-handling code.
- `TestCanceledReadQueryDiscardsConnection` unit test validates the core discard mechanism.
- `testAssetRepositorySpentOnlySupply` covers the edge case of zero supply when all vtxos are spent.
- No cross-repo breakage — `internal/infrastructure/db/sqlite` is not imported by any external repo.
- Postgres and SQLite queries are semantically identical (dialect differences only).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #1118, opened 2026-06-19
Extract standalone arkd-signer + shared txsigner lib (BREAKING: ARKD_SIGNER_ADDR required)

Kukks:
## Summary

Splits `arkd-wallet`'s two responsibilities. `arkd-wallet` becomes an onchain-wallet-only service; a new standalone **`arkd-signer`** service owns the operator signing key and `SignerService`. The shared tapscript-signing primitive is factored into **`pkg/ark-lib/txsigner`**, reused by the wallet and signer (and adoptable by the [emulator](https://github.com/arkade-os/emulator) — see follow-ups).

## What's new (additive)

- `pkg/ark-lib/txsigner` — chain-free tapscript PSBT signing primitive (`BuildPrevoutFetcher`, `SignTapscriptInput`, `ExtractFinalizedTx`).
- `pkg/arkd-signer` — standalone service (own module, `cmd/arkd-signer`, Dockerfile, compose service). Pure signer: holds the injected operator key, requires complete PSBTs (no chain access). Full parity with the wallet's old signer, including **deprecated signer keys** (`ARKD_SIGNER_DEPRECATED_KEYS`, per-leaf key selection, `GetPubkey` deprecated-signers).

## Clean break

- `arkd-wallet` no longer serves `SignerService`, holds no operator key, and its `LoadSignerKey` RPC is now `Unimplemented`.
- `arkd` signer-management: runtime key injection removed; the external-signer URL path is kept; `ARKD_SIGNER_ADDR` is now **required** (the silent `WALLET_ADDR` fallback is gone).
- CLI `arkd signer load` is URL-only (`--signer-prvkey` removed).

## :warning: Breaking changes — deploy migration

- Run the new `arkd-signer` service and set `ARKD_SIGNER_ADDR` on `arkd` (no more fallback).
- Move the operator key: `ARKD_WALLET_SIGNER_KEY` -> `ARKD_SIGNER_SECRET_KEY`; deprecated keys `ARKD_WALLET_DEPRECATED_SIGNER_KEYS` -> `ARKD_SIGNER_DEPRECATED_KEYS`.
- `arkd signer load --signer-prvkey` is removed; use `--signer-url`.

`docker-compose.regtest.yml`, the Makefile (`build-signer`, `run-signer`), `envs/signer.dev.env`, and the README are updated accordingly.

## Test plan

- [x] Unit tests: `ark-lib/txsigner`, `arkd-signer` (incl. deprecated-key selection + `GetDeprecatedPubkeys`), `arkd-wallet` — all green.
- [x] Whole arkd module cross-compiles for linux and `go vet ./internal/...` is clean (incl. test files).
- [x] `arkd-signer` is dependency-neutral (no change to arkd's `go.mod`/`go.sum`).
- [ ] CI: e2e/integration suite now runs on Docker+Linux. `arkdsigner.Dockerfile` was bumped to `golang:1.26.5` to match the modules' toolchain (it had lagged at `1.26.4`, breaking the docker build); the `sqlite/badger` suite is green. `postgres/redis` intermittently flakes on timing-heavy flows (`TestReactToFraud`, `TestUnilateralExit`), not a deterministic failure. The e2e harness recreates `arkd-signer` (not the wallet) for key rotation.

## Follow-ups (not in this PR)

- `arkd-wallet` adopting `txsigner` for its own LP-mode tapscript branch (removes the last duplicated copy).
- Emulator adopting `txsigner` via an `ark-lib` release (the "share a library" consolidation).
- Optional otel/pyroscope telemetry for `arkd-signer` (dropped here to keep the PR dependency-neutral).

## Summary by CodeRabbit

* **New Features**
  * Added `arkd-signer` as a standalone signing service (with health checks, key rotation, and transaction/Taproot script signing).
  * Updated local and containerized regtest workflows to run wallet and signer separately.
  * Added signer-focused build support (including a dedicated Docker image and build/run make targets).
* **Bug Fixes**
  * Improved forfeit transaction signing so connector inputs are signed correctly.
* **Documentation**
  * Updated guidance so `arkd-signer` is the signer service: `ARKD_SIGNER_ADDR` is required, and `arkd signer load` accepts only a signer URL.
  * Updated setup/testing instructions to include the full wallet+signer regtest stack.


ghost (changes requested):
## 🔍 Arkana Code Review — arkd-signer extraction

Clean architectural split. The decomposition is well-motivated and the implementation follows existing patterns. A few issues to address before merge, one of which is protocol-critical.

---

### 🚨 P1 — Protocol-Critical: nil-pointer dereference in signer signing loop

**`pkg/arkd-signer/core/application/signer.go:157`**

```go
for i, in := range ptx.Inputs {
    if bytes.Equal(in.WitnessUtxo.PkScript, txutils.ANCHOR_PKSCRIPT) {
```

If `in.WitnessUtxo` is nil, this panics. Yes, `BuildPrevoutFetcher` earlier will error on missing WitnessUtxo, but that's an implicit contract — a refactor could easily break it. A signing service that panics on malformed input is a DoS vector. Add an explicit nil guard:

```go
if in.WitnessUtxo == nil {
    continue // or return error
}
```

This same pattern also applies in the wallet's existing `SignTransaction` (pre-existing issue, not introduced here, but worth noting).

---

### 🚨 P1 — Protocol-Critical: fraud.go forfeit signing order needs verification

**`internal/core/application/fraud.go:167-174`**

```go
signedForfeitTx, err := s.signer.SignTransactionTapscript(ctx, forfeitTxB64, nil)
// ...
signedForfeitTx, err = s.wallet.SignTransaction(ctx, signedForfeitTx, false)
```

The dual-signing flow (signer signs VTXO tapscript input, wallet signs connector key-path input) is architecturally correct for the split. However:

1. **Does `s.wallet.SignTransaction` handle a partially-signed PSBT correctly?** The wallet's `SignTransaction` in LP mode skips taproot script-path inputs that already have signatures, right? Confirm the wallet won't try to re-sign or reject the already-signed tapscript input.

2. **Ordering matters**: If the wallet's signing depends on seeing the signer's signature in the PSBT (e.g., for sighash computation on the other input), this order is correct. If not, parallel signing might be possible but that's a follow-up. Just confirming the serial order is intentional and correct.

⚠️ **This is protocol-critical (forfeit path = money recovery). Requires human review sign-off regardless of code correctness.**

---

### ⚠️ P2 — Health handler always returns SERVING

**`pkg/arkd-signer/interface/grpc/handlers/healthservice.go:14-20`**

The health check unconditionally returns `SERVING`. It should delegate to `signer.IsReady(ctx)` so that orchestrators (Docker healthcheck, k8s liveness probe) can detect a misconfigured signer (e.g., missing secret key that somehow bypassed config validation).

```go
func (h *healthHandler) Check(ctx context.Context, ...) (...) {
    if !h.signer.IsReady(ctx) {
        return &grpchealth.HealthCheckResponse{Status: grpchealth.HealthCheckResponse_NOT_SERVING}, nil
    }
    // ...
}
```

---

### ⚠️ P2 — No TLS / mTLS support on the signer service

**`pkg/arkd-signer/interface/grpc/service.go:89`**

```go
grpc.Creds(insecure.NewCredentials()),
```

The signer holds the operator private key — the most sensitive secret in the system. In production, this service MUST NOT be reachable over cleartext. Even if the plan is "same-host networking only", a single misconfigured firewall rule exposes the signing key. At minimum:

- Add a `ARKD_SIGNER_NO_TLS` flag (default false) mirroring arkd's pattern.
- Or document explicitly that the signer MUST be deployed behind a sidecar proxy (e.g., Envoy with mTLS).

If this is intentionally deferred, add a TODO and a startup warning log.

---

### ⚠️ P2 — CORS wildcard on a signing service

**`pkg/arkd-signer/interface/grpc/service.go:132-137`**

```go
w.Header().Set("Access-Control-Allow-Origin", "*")
w.Header().Set("Access-Control-Allow-Headers", "*")
```

`Access-Control-Allow-Origin: *` on a signing endpoint is concerning. The signer should never be browser-accessible. If the grpc-gateway is only for internal health/status probes, restrict CORS or remove it entirely.

---

### 📋 P3 — Cross-repo breakage: 6+ downstream repos need migration

The following repos still reference `ARKD_WALLET_SIGNER_KEY` in their docker-compose files and will break:
- `arkade-os/solver`
- `arkade-os/introspector-review`
- `arkade-os/arkd-pentester`
- `arkade-os/emulator`
- `arkade-os/arkade-kotlin`
- `arkade-os/bancod`
- `arkade-os/layerzero-usdt0-arkade-demo`

None of these set `ARKD_SIGNER_ADDR`, so `arkd` will fail to start (`ARKD_SIGNER_ADDR is required`). Consider opening tracking issues or a follow-up PR batch for these repos.

Also, `arkade-kotlin` still has the `LoadSignerKey` RPC in its vendored proto — this will need updating.

---

### 📋 P3 — Config key length validation missing for current key

**`pkg/arkd-signer/config/config.go:61-63`**

```go
buf, err := hex.DecodeString(c.SecretKey)
// ...
prvkey, _ := btcec.PrivKeyFromBytes(buf)
```

The deprecated key parser validates `len(buf) != 32` but the current key does not. `btcec.PrivKeyFromBytes` will silently truncate/pad non-32-byte inputs. Add the same check:

```go
if len(buf) != 32 {
    return fmt.Errorf("invalid signer secret key: must be 32 bytes (64 hex chars)")
}
```

---

### ✅ What looks good

- `txsigner` library is clean, well-tested, chain-free by design — good reusable primitive.
- Deprecated key selection logic faithfully ported from wallet with full parity.
- E2e harness correctly rewired to recreate `arkd-signer` instead of `arkd-wallet` for key rotation tests.
- Clean removal of `SignModeSigner` from wallet — no dead code left behind.
- `LoadSignerKey` on wallet returns `Unimplemented` rather than silently succeeding — correct migration signal.
- Config `String()` method properly redacts secrets.

---

**Verdict**: Request changes on P1 items. The forfeit signing change is protocol-critical and requires human sign-off even after code fixes.

bitcoin-coder-bob:
unclear if i need those case checks on `CSVMultisigClosure` and `ConditionCSVMultisigClosure` in `pkg/arkd-signer/core/application/signer.go` I'm thinking these are not really cases that we will encounter. maybe fine to keep defensively?  Those checks probably not needed, they were removed in PR: https://github.com/arkade-os/arkd/pull/1121 in commit [d00197d](https://github.com/arkade-os/arkd/pull/1121/commits/d00197d9848202c7c1f3548ab571c7cd068c8c75)

bitcoin-coder-bob:
@coderabbitai review

internal/core/application/fraud.go
louisinger:
```suggestion
	// Sign the connector input (a wallet-owned key-path output) with the wallet:
	// the operator signer holds no wallet keys, so it cannot sign it.
	signedForfeitTx, err = s.wallet.SignTransaction(ctx, signedForfeitTx, true)
	if err != nil {
		return fmt.Errorf("failed to sign forfeit connector input: %s", err)
	}
```

in theory we can simplify and remove the FinalizeAndExtract call (if we pass `true` to SignTransaction)

pkg/ark-lib/txsigner/txsigner_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/require"
+)
+
+// taprootPkScript returns a syntactically valid P2TR scriptPubKey for the given
+// key. The signing primitive only needs a P2TR-shaped prevout; it does not
+// verify that the script commits to the leaf, so this is sufficient for a unit
+// test of SignTapscriptInput.
+func taprootPkScript(t *testing.T, key *btcec.PublicKey) []byte {
```
louisinger, line 97: nit : helper at the bottom of the file

pkg/ark-lib/txsigner/txsigner.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+	)
+	return nil
+}
+
+// ExtractFinalizedTx finalizes every input and returns the hex-encoded raw tx.
+// Tapscript inputs are finalized by assembling their closure witness from the
+// gathered TaprootScriptSpendSig entries (plus any condition-witness PSBT field);
+// other inputs are finalized via psbt.Finalize.
+func ExtractFinalizedTx(ptx *psbt.Packet) (string, error) {
```
louisinger, line 61: this function can be simplified thanks to `script.FinalizeVtxoScript`. also if we export this in ark-lib, we can remove `FinalizeAndExtract` method from `TxBuilder` and use this utility instead ?

pkg/ark-lib/txsigner/txsigner.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil/psbt"
+	"github.com/btcsuite/btcd/txscript"
+)
+
+// BuildPrevoutFetcher builds a prevout fetcher from the PSBT's WitnessUtxo
+// fields. It errors if any input is missing its WitnessUtxo (no chain fallback).
+func BuildPrevoutFetcher(ptx *psbt.Packet) (txscript.PrevOutputFetcher, error) {
+	return txutils.GetPrevOutputFetcher(ptx)
+}
```
louisinger: useless ? let's use directly txutils function ?

louisinger (changes requested):

pkg/ark-lib/txsigner/txsigner.go
bitcoin-coder-bob: done in: [059f353](https://github.com/arkade-os/arkd/pull/1118/commits/059f353284269478b32e038ad7cac32bba72c589)

pkg/ark-lib/txsigner/txsigner.go
bitcoin-coder-bob: done in: [059f353](https://github.com/arkade-os/arkd/pull/1118/commits/059f353284269478b32e038ad7cac32bba72c589)

pkg/ark-lib/txsigner/txsigner_test.go
bitcoin-coder-bob: done in: [059f353](https://github.com/arkade-os/arkd/pull/1118/commits/059f353284269478b32e038ad7cac32bba72c589)

internal/core/application/fraud.go
bitcoin-coder-bob: done in: [059f353](https://github.com/arkade-os/arkd/pull/1118/commits/059f353284269478b32e038ad7cac32bba72c589)

internal/interface/grpc/handlers/signer_manager.go
```diff
@@ -4,42 +4,38 @@ import (
...
+	// Runtime key injection into arkd-wallet is no longer supported: the operator
+	// key now lives in arkd-signer (ARKD_SIGNER_SECRET_KEY). Only repointing arkd
+	// at a signer URL is supported.
+	if req.GetSignerPrivateKey() != "" {
 		return nil, status.Error(
-			codes.InvalidArgument, "address and private key are mutually exclusive",
+			codes.InvalidArgument,
+			"runtime signer key injection is no longer supported; "+
+				"configure ARKD_SIGNER_SECRET_KEY on arkd-signer and provide signer_url",
 		)
```
louisinger, line 33: if not supported, should we remove the signerPrivateKey from the proto request ?

louisinger (commented):
LGTM but let's wait for @altafan to double check because he worked on the signer initially

internal/interface/grpc/handlers/signer_manager.go
bitcoin-coder-bob: i think if we keep this it is a good way for deprecation. older clients will get this error message.

bitcoin-coder-bob:
Addressing the review findings that have been carried across several passes. Three are now fixed in `2d3754da`; the rest are deliberate decisions rather than oversights, so recording the reasoning here so it stops being re-raised each pass.

## Fixed

**CORS wildcard.** Valid, and the sharpest part of the "no auth" finding. The gateway answered preflight with `Access-Control-Allow-Origin: *` on a service that signs with the operator key. Since gateway JSON posts are not CORS-simple, that wildcard was precisely what would let a page an operator happens to be visiting reach a signer it can route to. The signer's callers are arkd and operator tooling, never a browser, so it now sends no CORS headers and refuses preflight. Covered by `TestRouterSendsNoCORSHeaders`, which also pins that `/healthz` stays reachable for the container healthcheck.

**Handler-level tests for the signing paths.** `TestSignerHandlerSignTransactionTapscript` and `TestSignerHandlerSignTransaction`. Worth noting the indexes are `int32` on the wire, not `uint32`, so a negative index is representable and reaches the conversion. The out-of-range case covers `-1` alongside past-the-end values. Also asserts a selected subset signs only the requested input, an empty list signs every eligible input, and anchor inputs are skipped rather than signed.

**Deterministic dual-signing test.** `TestForfeitWitnessIsAuthorizedByOperatorKey` pins the invariant `broadcastForfeitTx` depends on: after the wallet's unscoped second pass appends a signature from its forfeit key to the vtxo input, the witness that finalizes must still be authorized by the operator key, and the wallet's stray signature must not appear in it. I verified the test fails when the wallet forfeit pubkey is placed inside the closure, which is exactly the hazard the comment in `fraud.go` describes, so it is a real guard rather than a passing assertion.

## Not doing, with reasons

**Nil guard in the signing loop.** `GetPrevOutputFetcher` runs before the loop, iterates every input, and returns an error on any nil `WitnessUtxo`. Its return value is what builds `sigHashes`, so no signing path can skip it and still sign. The coupling is structural rather than incidental, which makes the proposed guard unreachable code. The originally suggested form, `continue`, would additionally have been a regression: it would silently skip an input that currently produces a clear error, and `TestSignRejectsMissingWitnessUtxo` already pins that behaviour.

**`insecure.NewCredentials()` as a blocker on this PR.** `arkd-wallet` uses the identical posture, `insecure.NewCredentials()` plus a CORS wildcard, and it held this same operator key before the split. So this is not a trust boundary this PR weakens, it is the existing posture carried across. The concrete exposures have been closed: the regtest compose no longer publishes the port to the host, and the CORS wildcard is gone. Real transport auth belongs to both services together, in its own change, rather than being bolted onto one side here.

**Scoping the wallet's second signing pass to the connector input.** Correct in principle, but it needs input indexes on the wallet's `SignTransaction` RPC, and only `SignTransactionTapscript` carries them today while hardcoding `extractRawTx=false`. That is a proto change on an already large breaking PR. The invariant is documented at the call site and now has a test that fails if it is ever violated, which covers the risk until the RPC can carry the parameter.

**`--signer-prvkey` removed without a stub.** Intentional. This PR is explicitly breaking, `ARKD_SIGNER_ADDR` is now required, and the migration is documented in the README. A hidden deprecated flag is polish rather than a correctness issue.

## Note on prior passes

Several of these were decided in earlier rounds but never written down publicly, which is why each pass reports them as untouched. That is on me for not recording them here sooner.

louisinger:
replaced PR

PR #1120, opened 2026-06-19
Wallet rotation safety (integration preview): eager forfeit + multi-wallet sweeps + signer-key-per-leaf

bitcoin-coder-bob:
> [!WARNING]
> **Integration / preview branch, not meant to be merged as a unit.** It bundles several in-flight PRs plus one new fix so the full "what happens when you switch out a wallet with active batches/vtxos" story can be reviewed and exercised by CI together. The individual pieces land through their own PRs (linked below).

## Why

Rotating the signer key, or switching the arkd wallet, while batches and vtxos are still live is high risk. On-chain funds and vtxo scripts stay locked to the old signer key, and several operations still need that key:

- **Forfeit txs** (fraud punishment) were signed lazily, at fraud-broadcast time. After a rotation the wallet signed the old-key forfeit leaf with the new key, the tx could not be finalized, and the fraud went unpunished.
- **Sweeps** of expired old-key batch outputs and **settles** of old-key coins also need the old key.

## What's bundled

- **#1110 eager forfeit** (`presign-forfeit-txs`): sign forfeit txs at collection time so they are broadcast-ready without the live signer. Core fix for the fraud-after-rotation gap.
- **#1099 / #1101 / #1102 multi-wallet** (`dial` / `sweep-fallback` / `wallet-settings`): dial a primary plus fallback arkd-wallets, sign sweeps with the primary or any fallback, and persist the wallet addresses in the settings domain. Lets an old wallet stay reachable as a sweep fallback across a switch.

## New in this branch

- **`signerKeyForLeaf` hardening** (`pkg/arkd-wallet`): select the signer key per tapscript leaf across **all** multisig closures, including CSV and condition-CSV. Previously CSV (sweep) leaves fell through to the current key, so deprecated-key sweeps of old-key outputs used the wrong key. Now an explicitly required leaf that references none of the wallet's keys is a hard error instead of a silent wrong-key signature, and the signer no longer dereferences `keyMgr` in signer mode (also fixes a nil-pointer panic in the existing wallet unit test).
- **`TestEagerForfeitSurvivesWalletRotation`** (e2e): regression test proving #1110's eager forfeit lets the server punish fraud across a hard signer rotation. It mirrors the pre-fix repro exactly (hard rotation, no deprecated key retained) but asserts the opposite outcome: the server broadcasts the pre-signed forfeit and claims the fraudulently unrolled vtxo.

## Validation (live regtest)

All green on this combined branch:

- `TestEagerForfeitSurvivesWalletRotation`: eager forfeit survives a hard rotation
- `TestDeprecatedSignerKey`: rotation + deprecated keys + boarding (the explicit-index path through the new `signerKeyForLeaf`)
- `TestReactToFraud`: normal forfeit / sweep / fraud reactions with everything combined

## Base and how the pieces actually land

Targeted at `master` so CI runs the whole stack. Because it bundles #1110 and #1099/#1101/#1102, it should not be merged here; those PRs merge individually. The only genuinely new work, the `signerKeyForLeaf` fix and the verify test, should land as their own PRs once the others are in (`signerKeyForLeaf` into `master`; the verify test onto #1110).


bitcoin-coder-bob:
Closing this integration/CI preview branch. It bundled several in-flight PRs so the full wallet-rotation-safety story could be reviewed and exercised by CI together; it was never meant to merge as a unit.

The individual pieces live in their own PRs:
- eager forfeit: #1110
- multi-wallet support: #1099 / #1101 / #1102
- signer-key-per-leaf fix: #1121

The TestEagerForfeitSurvivesWalletRotation e2e test added on this branch is a regression test for #1110's eager-forfeit fix and should be added there.

PR #1123, opened 2026-06-22
fix: SubmitTx concurrent calls error with "duplicated tx" instead of "vtxo spent"

louisinger:
check for duplicated offchain also in cache before processing the SubmitOffchainTx.

@altafan @sekulicd please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offchain transaction filtering to correctly track transactions through their lifecycle stages (acceptance and finalization). Previously, transactions with failures were excluded from queries; they are now properly included based on their processing stage.

* **Tests**
  * Expanded test coverage for offchain transaction workflows to validate both successful and failure paths through the system.


tiero (commented):
The early cache check takes `offchainTxMu` and releases it before validation, while the real write happens under a separate lock later. So the lock doesn't span the check-then-act, and the race it targets can still slip through (both pass the early check, one wins, the other gets `VTXO_ALREADY_SPENT`). Re-checking under the second lock would patch it, but at that point we're guarding a uniqueness invariant with read-then-write across a dropped lock, which is what got us here.

This smells like we're working around the symptom rather than fixing it. `txid` is unique, so the question is whether dedup should be one atomic operation at the store/cache layer instead of a check. You both know the arkd patterns far better than I do, so I'll leave the how to you, but I wanted to flag that this doesn't look like it closes the underlying race.

PR #1131, opened 2026-06-29
Fix: cap metadata of interest values size in logging interceptor

chris-ricketts:
**Summary**
- Adds a 100-byte limit for single incoming gRPC metadata values selected for debug logging.
- Logs a warning with the metadata key and value length when a selected value is too large.
- Replaces oversized values with `arklabs/invalid` before writing metadata to logs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logging safety by limiting the maximum size of metadata values included in debug output.
  * Added warnings when a metadata value exceeds the allowed size, replacing oversized entries with a fixed placeholder instead of logging the full content.
  * Improved multi-value metadata handling so sanitized results reflect either a single value or a list based on what was received.
* **Tests**
  * Added coverage for metadata sanitization scenarios, including empty/non-matching metadata and oversized value replacement.

### DIFF
--- a/internal/interface/grpc/interceptors/logger.go
+++ b/internal/interface/grpc/interceptors/logger.go
@@ -13,6 +13,11 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+const (
+	maxMetadataValueSizeBytes = 100
+	invalidMetadataValue      = "arklabs/invalid"
+)
+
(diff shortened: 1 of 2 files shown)


altafan (commented):
@chris-ricketts please add unit tests

PR #1132, opened 2026-06-29
feat(intent): set BIP-322 PSBT_GLOBAL_GENERIC_SIGNED_MESSAGE (0x09)

afsheenb:
Intent-proof PSBTs never set BIP-322's global 0x09 field (the encoded message being signed). Without it a co-signer cannot recompute the to_spend commitment from PSBT-internal data alone, making it impossible to reliably distinguish a genuine ownership proof from an ordinary fund-moving spend before contributing a partial signature.

Append the 0x09 global Unknown to the toSign PSBT inside New(), using the message string that is already available there. The key is a single byte (0x09); the value is the UTF-8 message bytes.

Update proof fixture expected values to match the new PSBT wire format and add a BIP-322_global_0x09_field sub-test that asserts the field is present and equal to the message across all valid fixtures.

## Summary by CodeRabbit

* **New Features**
  * PSBTs now embed the original signed message metadata so downstream co-signers can reconstruct commitments from shared intent data.
* **Tests**
  * Added assertions that the produced PSBT includes the signed message payload and matches the expected intent bytes across valid scenarios.
* **Chores**
  * Updated expected fixture outputs to reflect the revised PSBT encoding.
* **Documentation**
  * Added clarifying doc comments for relevant intent APIs.


afsheenb:
@coderabbitai I've updated the docstrings. All good now?

altafan (approved):
@afsheenb thank you for the contribution, clean!

PR #1078, opened 2026-05-21
Split client-lib and client-wallet

altafan:
This PR refactors the client-lib and extracts the stateless client wallet used for integration tests and as the base for the client-facing CLI.

The main purpose of this refactor is to break the dependency between the stateful client (go-sdk) and the stateless one (what's now pkg/client-wallet), and make them reuse shared types, functions, and interfaces instead for interoperability instead.

The secondary goal was to make the shared code agnostic to the tx signing so that funds locked by any kind of script can be provided. For this the original `Vtxo` and `VtxoWithTaptree` types have been merged into the first one, and a new `SigningClosure script.Closure` has been added. This field is required when using the vtxo/utxo as input of a tx and represents the very tapleaf of the given `Tapscripts` used for signing.  The type `type SignTx func(ctx context.Context, tx string) (string, error)` is also introduced to generalize and abstract the signing of a tx that ust always be provided by callers.

In client-lib, all types and interfaces are now at "root" level, so, for example, they can be imported with `clientlib.Vtxo` instead of `types.Vtxo`.

In client-lib's folders we can find:
- `client/` - grpc implementation of the `Client` interface
- `indexer/` - grpc implementation of the `Indexer` interface
- `explorer/` - mempool implementation of the `Explorer` interface
- `batch-session/` - collection of functions to join a batch (low-level `JoinBatch`, `RegisterIntent`, `DeleteIntent` and high-level orchestrators `Settle`, `CollaborativeExit`, `RedeemNotes`)
- `offchain-tx/` - collection of functions to spend offchain (low-level builders for any kind of tx, high-level orchestrators for any tx)
- `unroll/` - collection of functions for unilateral exit

This adds a brand new `pkg/wallet` that contains the stateless client wallet built on top of the client-lib types, interfaces and functions and that is used for integration tests and as base for the client-dacing CLI.

Please @louisinger @sekulicd review.

## Release Notes
- **New Features**
  - Added new batch-session APIs (join, settle, redeem notes, collaborative exits) and a refreshed batch-session orchestration flow.
  - Added off-chain transaction support for payments and asset operations (issue/reissue/burn).
  - Improved the explorer client with unified typed WebSocket + polling event streams.
- **Refactoring**
  - Migrated the CLI and SDK usage to the newer consolidated client libraries.
- **Breaking/Behavior Changes**
  - SDK: removed legacy batch-session and asset operations from the older client surface.
  - CLI: `--datadir` no longer sets a wallet environment variable.
- **Chores**
[...]

PR #1134, opened 2026-07-01
Fix: fix two bugs in client-lib (send on closed channel panic and deprecated URL)

guggero:
Two issues I ran into after updating to the latest version of `arkd` and the Go SDK (`pkg/client-lib`).

1. If the caller passes in a replay channel but closes it before the anonymous goroutine could write to it, a "send on closed channel" panic occurs. Not sending in a goroutine fixes the issue.
2. The `/api/fee-estimates` endpoint of Mempool is deprecated and currently not served anymore on https://mempool.mutinynet.arkade.sh/api/fee-estimates, so we switch the client library to use the `/api/v1/fees/recommended` endpoint instead, which works on [mutinynet](https://mempool.mutinynet.arkade.sh/api/v1/fees/recommended).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved batch session shutdown to prevent rare “send on closed channel” panics when replay events are still in flight.
  - Updated mempool fee rate fetching to use the recommended-fee endpoint and field, with safe fallback behavior for missing endpoints.
- **Tests**
  - Added a regression test to ensure batch sessions terminate cleanly even if the caller closes the replay channel immediately after return.
  - Updated fee-rate service tests to match the latest recommended-fee API contract and error responses.

### DIFF
--- a/pkg/client-lib/batch_session_handler.go
+++ b/pkg/client-lib/batch_session_handler.go
@@ -132,13 +132,19 @@ func JoinBatchSession(
 				continue
 			}
 
+			// Forward to the caller's replay channel inline and
+			// non-blocking. This must NOT run in a detached goroutine:
+			// the caller owns replayEventsCh and closes it once
+			// JoinBatchSession returns, so a goroutine outliving the
+			// return would race that close and panic ("send on closed
+			// channel"). An inline non-blocking send keeps the loop
+			// responsive (a slow/unread consumer just drops the event)
+			// while guaranteeing every send happens-before the return.
 			if options.replayEventsCh != nil {
-				go func() {
-					select {
-					case options.replayEventsCh <- notify.Event:
-					default:
-					}
-				}()
+				select {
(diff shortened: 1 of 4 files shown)


altafan (changes requested):
@guggero thank you for the fixes.

CI atm is red because of the breaking changes to `GetFeeRate`.

I opened a [PR](https://github.com/thalia-finance/arkd/pull/1) against your branch with fixes to the backward compatibility of that method + some cleanup. Please merge it to make this ready for the merge on master.

guggero:
Thanks for the fixes, @altafan! I merged the PR.

altafan (approved):
@guggero merging. Thank you again.

Please, be aware of #1078 that is gonna be merged soon as well: it refactors the client-lib making it a pure collection of reusable types, functions and services, and also moves the stateless wallet in a dedicated module.

The import path for explorer, indexer and client services will slightly change, so I suggest you to give it a read. I'm available to sync somewhere else to guide you through all changes if you prefer

PR #1135, opened 2026-07-02
fix: CSVMultisigClosure small int decoding logic

louisinger:
fix the small int decoding logic. the current version is failing to decode block locktime from 81 to 96. this PR is fixing it.

@altafan @d4rp4t please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relative locktime decoding consistency, including cases using small-integer encodings, so the interpreted locktime type/value is preserved.
  * Fixed CSV multisig locktime token parsing to retain the correct numeric value for small locktime tokens.
* **Tests**
  * Expanded round-trip test coverage for block-based relative locktimes across the full supported range, including decoding verification.

### DIFF
--- a/pkg/ark-lib/script/closure.go
+++ b/pkg/ark-lib/script/closure.go
@@ -355,14 +355,17 @@ func (d *CSVMultisigClosure) Decode(script []byte) (bool, error) {
 	}
 
 	var sequence []byte
-	if txscript.IsSmallInt(tokenizer.Opcode()) {
-		if tokenizer.Opcode() == txscript.OP_0 {
-			// minimal encoding for a zero scriptnum is an empty slice
-			sequence = []byte{}
+	
+	scriptNumOpcode := tokenizer.Opcode()
+	// small int (0-16) are encoded with 1 OP_x opcode
+	if txscript.IsSmallInt(scriptNumOpcode) {
+		if scriptNumOpcode == txscript.OP_0 {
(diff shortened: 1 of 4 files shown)


altafan (changes requested):
Please add a new round-trip unit test that ensures **ALL** block-locktime are encoded/decoded the right way

pkg/ark-lib/script/closure.go
```diff
@@ -355,14 +355,17 @@ func (d *CSVMultisigClosure) Decode(script []byte) (bool, error) {
...
+		if scriptNumOpcode == txscript.OP_0 {
+			sequence = []byte{} // minimal encoding policy forces empty byte slice for OP_0
 		} else {
-			sequence = []byte{tokenizer.Opcode()}
+			sequence = []byte{scriptNumOpcode - (txscript.OP_1 - 1)}
 		}
 	} else {
+		// if bigger than 16, int is encoded as pushdata byte slice
 		sequence = tokenizer.Data()
 	}
```
altafan, line 370: does this mean we have to do the same wherever we use `BIP68DecodeSequenceFromBytes`?

louisinger: no, this is only needed while parsing the raw script opcodes. BIP68DecodeSequenceFromBytes is handling it in `MakeScriptNum`

PR #1142, opened 2026-07-07
Implement optional `StreamMaxLifetime` for server streams

s373nZ:
Implements an optional `StreamMaxLifetime` setting which bounds the duration a stream should live for. When set, clients are expected to re-connect when the connection is reaped.

## Summary by CodeRabbit

* **New Features**
  * Added `STREAM_MAX_LIFETIME` configuration to cap maximum lifetime of server-streaming RPCs (default: 1800s).
  * Supports `STREAM_MAX_LIFETIME=0` to disable the cap.
* **Bug Fixes**
  * Improved cleanup of abandoned/disconnected streaming sessions by terminating streams when the configured lifetime elapses.
* **Documentation**
  * Updated the Environment Variables section in the README with the new setting and its behavior.
* **Tests**
  * Added unit tests verifying stream lifetime expiration and bounded context timeout handling.

PR #1143, opened 2026-07-07
fix(indexer): terminate abandoned GetSubscription streams on reconnect

Kukks:
`GetSubscription` streams currently have no tie to their listener's lifecycle. Reconnecting with the same subscription id attaches a second stream to the same listener channel while the previous one stays parked in its select loop: if the server never observes the old client's disconnect (typical behind a load balancer that keeps the connection alive), every reconnect leaks the previous stream, and the leaked streams keep competing for events on the shared channel, so the reconnected client randomly loses events. A stream also survives its own subscription's removal (unsubscribe or reconnect-timeout expiry), since the loop never watches `listener.done` — unlike `GetEventStream` and `GetTransactionsStream`.

This gives the broker an explicit notion of a listener's active consumer: `attach` registers the calling stream as the sole consumer (cancelling any pending reap timeout) and displaces a previously attached stream by closing its `displaced` channel; `detach` releases it and reports whether the caller still owned the listener, so a displaced stream no longer arms the reap timer or removes the listener out from under its successor. `GetSubscription`'s select now also exits on `listener.done` and on displacement, and the new and old flows share the same attach/detach path. Events buffered on the listener channel survive a reconnect and are delivered to the new stream.

One behavioral note: a subscription id now serves exactly one stream at a time, newest wins. Two clients deliberately sharing an id previously got a random split of events (each event went to exactly one of them), which doesn't seem like behavior anyone could rely on; with this change the latest stream takes over and earlier ones end cleanly. Happy to adjust if concurrent consumers per subscription are meant to be supported.

Complements #1142: a max stream lifetime bounds streams whose clients never come back; this change removes the leak on the reconnect path and ties stream lifetime to the subscription itself.

Covered by new handler-level tests (reconnect displaces the previous stream and events flow to the successor only, streams end on listener removal, displaced streams do not reap or remove the listener under the successor) plus broker-level tests for the attach/detach contract. `go test -race ./internal/interface/grpc/handlers/` and `golangci-lint run --tests=false` are green.

## Summary by CodeRabbit

* **New Features**
  * Added explicit subscription take-over so reconnecting streams become the single active consumer.
  * Stream setup now treats blank subscription IDs as new and immediately starts sending from an inline listener.
* **Bug Fixes**
  * Older/displaced streams no longer receive events after takeover.
  * Improved timeout handling to avoid premature listener removal and ensure correct cleanup on detachment/reconnect.
* **Tests**
  * Expanded coverage for attach/release ownership, timeout/no-op rules, and high-concurrency reconnect races with deterministic ordering.
* **Chores**
  * Enabled gRPC keepalive pings to better detect and close dead connections.

### DIFF
--- a/internal/interface/grpc/handlers/broker.go
+++ b/internal/interface/grpc/handlers/broker.go
@@ -243,11 +243,10 @@ func (h *broker[T]) attach(id string) (*listener[T], *attachment, error) {
 	return l, l.attached, nil
 }
 
-// detach releases an attachment taken with attach and reports whether the
-// caller was still the listener's active consumer. A false return means the
-// stream was displaced (or the listener is already gone) and must not touch
-// the listener's lifecycle: cleanup belongs to the current holder.
-func (h *broker[T]) detach(id string, att *attachment) bool {
+// release ends att's hold on the listener and settles its lifecycle atomically:
+// kept for reconnectWindow if it still has filters, removed otherwise. Returns
+// false if att was displaced — the successor owns the lifecycle.
+func (h *broker[T]) release(id string, att *attachment, reconnectWindow time.Duration) bool {
 	h.lock.Lock()
 	defer h.lock.Unlock()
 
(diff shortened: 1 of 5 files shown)


Kukks:
Addressed in 2d7307d0. The select now checks the exit signals (ctx done / listener removed / displaced) in a non-blocking select before the blocking one, so a displaced stream stops draining `listener.ch` and leaves buffered events for its successor.

Since `select` picks a ready case at random and the channel is shared, this can't be a hard guarantee at the exact instant an event and the displacement become ready together — but that boundary is already best-effort (the dispatch goroutine drops on a full buffer), and the priority check removes the case that actually mattered: a displaced stream steadily draining events into a dead connection. New test `old flow displaced stream does not consume buffered events` parks the predecessor mid-`Send`, buffers an event, displaces it, then asserts the event survives for the successor; it fails reliably without the priority check.

PR #1152, opened 2026-07-13
fix: duplicated swept vtxos in sweep tx event

### LINKED ISSUE #1138: Bug: SweepTx event message contains duplicated swept vtxo
While testing on go-sdk, i discovered that  tests like TestBalance/recoverable_(swept) receive sweep tx events that contain duplicated list of swept vtxos. Harmless, but requires to be fixed

Dunsin-cyber:
fixes #1138 

the sweep event listed the same vtxo twice because the queries used to collect preconfirmed vtxos also returned the leaves.

### Fix: make the queries return what their names say, instead of filtering at call sites:

- GetSweepablePreconfirmedVtxosByCommitmentTxid (renamed): adds preconfirmed = true, leaves excluded
- GetDescendantVtxos (new): same walk as `GetAllChildrenVtxos` but depth > 0, seed excluded. 
- GetAllChildrenVtxos untouched (checkpoint sweep needs the seed)
This covers both event producers (sweeper + admin sweep) with no filter code to forget.

Also needed: leaf collection now takes every vtxo output of a leaf tx (not just vout 0), and `FullySwept` counts distinct leaf txids, otherwise multi-receiver leaves would be dropped from the event / break the accounting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved batch sweep handling to include all VTXO outpoints from each leaf transaction (skipping anchor/extension outputs).
  * Refined descendant VTXO expansion used during sweeps to improve coverage and avoid repeats.
  * Updated full-round sweep completion detection to rely on distinct swept transaction IDs.
  * Tightened commitment-scoped sweep selection to eligible preconfirmed VTXOs.
* **Tests**
  * Enhanced sweep event validation to ensure no duplicate swept VTXO outpoints are reported.

### DIFF
--- a/internal/core/application/admin.go
+++ b/internal/core/application/admin.go
@@ -865,34 +865,9 @@ func (a *adminService) saveBatchSweptEvents(
 			}
 		}
 
-		// get preconfirmed vtxos
-		preconfirmedVtxos := make([]domain.Outpoint, 0)
-		if commitmentRootSwept {
-			var err error
-			preconfirmedVtxos, err = vtxoRepo.GetSweepablePreconfirmedVtxosByCommitmentTxid(
-				ctx,
-				commitmentTxid,
-			)
-			if err != nil {
-				log.WithError(err).
-					Error("error while getting sweepable vtxos by commitment txid")
-			}
-		} else {
-			seen := make(map[string]struct{})
-			for _, leafVtxo := range leafVtxos {
-				descendants, err := vtxoRepo.GetDescendantVtxos(ctx, leafVtxo)
-				if err != nil {
-					log.WithError(err).Error("error while getting descendant vtxos")
-					continue
-				}
-				for _, descendant := range descendants {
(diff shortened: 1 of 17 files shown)


Dunsin-cyber:
Hi @altafan, quick context on the approach.

The simple fix for #1138 would be to just filter the leaf vtxos out of the preconfirmed list in `sweeper.go` and `admin.go`. But that treats the symptom. The real issue is that the repository functions weren't narrow enough. `GetSweepableVtxosByCommitmentTxid` returned all the batch vtxos, leaves included, even though both of its callers only wanted the preconfirmed ones.

So instead of filtering after the fact, I made the queries honest: a renamed `GetSweepablePreconfirmedVtxosByCommitmentTxid` that returns only preconfirmed vtxos, and a `GetDescendantVtxos` that excludes the seed. With the sources correct, the two lists can't overlap and no call-site filtering is needed.

Happy to go the simpler route if this looks invasive.

PR #1164, opened 2026-07-29
deps: bump golang.org/x/text to v0.39.0 (CVE-2026-56852)

bitcoin-coder-bob:
Trivy started failing on every open PR with `CVE-2026-56852` (HIGH): a `norm.Iter`
in `golang.org/x/text` can enter an infinite loop on malformed input. Fixed in
v0.39.0.

The advisory landed after master's last scan, so master is affected too and will
go red on its next push. `golang.org/x/text` is an indirect dependency in 7 of
the 9 modules, so all 7 move together to keep `check-go-sync` happy.

The Docker image that Trivy scans builds two binaries, `arkd` from the root
module and `ark` from `pkg/ark-cli`, which is why the scan reported the finding
twice.

### Transitive bumps

`golang.org/x/text` v0.39.0 requires `x/tools` v0.47.0, `x/mod` v0.37.0 and
`x/sync` v0.21.0 in its own `go.mod`, which in turn pull up `x/crypto` v0.53.0,
`x/net` v0.56.0, `x/sys` v0.46.0 and `x/term` v0.44.0 in the root module. That is
minimal version selection, not a deliberate widening of scope. No code changes,
only `go.mod` and `go.sum`.

### Verification

- All 9 modules build.
- `go vet ./...` clean.
- `make lint` reports 0 issues.
- `go test ./internal/core/...` passes, plus `pkg/ark-lib`, `pkg/errors`,
  `pkg/macaroons` and `pkg/kvdb`.
- `go mod tidy` is idempotent, so `check-go-sync` passes.

The container-backed suite was not run, since nothing but dependency versions
changed.

### Note on overlap

This same commit is already cherry-picked onto #1160, #1161, #1162 and #1163 so
those PRs are not blocked waiting on this one. Whichever lands first will make
the others conflict on `go.sum`. Resolve with `git rebase master` followed by
`go mod tidy` rather than hand-merging the hash lines.

## Summary by CodeRabbit

* **Chores**
  * Updated indirect Go module dependencies across the project and related packages.
  * Includes refreshed security, networking, synchronization, terminal, text-processing, and development tooling components.
  * No user-facing features or public APIs changed.

PR #1168, opened 2026-08-18
docs: give the generated OpenAPI documents info and servers

gringokiwi:
Documentation metadata for the generated OpenAPI specs. No endpoint, message, or handler behaviour changes.

## Problem

Every spec under `api-spec/openapi/swagger/ark/v1/` is generated with placeholder metadata:

```json
"info": { "title": "", "version": "version not set" },
```

and **no `servers` block at all**. Two consequences:

1. An OpenAPI document without `servers` has no base URL. Any tool that renders an interactive console has nothing to send a request to, so "Try it" cannot work.
2. An empty `title` renders as a blank heading, and `"version not set"` is visible to anyone reading the spec.

This is the blocker for the API playground the docs site wants to publish — arkade-os/docs#60.

## Approach

Set the metadata through the `meshapi.gateway.openapi_doc` annotation, which the protos already have available (they import `meshapi/gateway/annotations.proto` for their HTTP bindings). Keeping it in the proto means it regenerates correctly and nobody has to remember to patch generated JSON.

| Spec | Title | Servers |
|---|---|---|
| `service` | Arkade Service API | all four |
| `indexer` | Arkade Indexer API | all four |
| `admin` | Arkade Admin API | none |
| `wallet` | Arkade Wallet Initializer API | none |
| `signer_manager` | Arkade Signer Manager API | none |
| `types` | Arkade Shared Types | none |

The four servers are `https://arkade.computer` (mainnet), `https://mutinynet.arkade.sh`, `https://signet.arkade.sh`, and `http://localhost:7070` (arkade-regtest).

**`admin`, `wallet`, and `signer_manager` deliberately get no `servers`.** They are operator-internal, and a spec that lists a public URL next to administrative endpoints invites someone to point a console at it. Their descriptions say plainly that they are not for application developers.

`version` is `"v1"`, matching the `ark.v1` proto package. That describes the API surface and stays correct, rather than drifting with each arkd release the way an embedded release number would.

The service and indexer descriptions both point at https://docs.arkadeos.com and say most applications should reach for an SDK first. These endpoints are for debugging and for understanding the wire format.

## Verification

Regenerated with `make proto`, then:

```
go build ./...   # passes
```

```
service:  title 'Arkade Service API',  version 'v1', 4 servers, 15 paths
indexer:  title 'Arkade Indexer API',  version 'v1', 4 servers, 15 paths
types:    title 'Arkade Shared Types', version 'v1', no servers
admin:    title 'Arkade Admin API',    version 'v1', no servers
```

Documents remain OpenAPI 3.1.0 and path counts are unchanged.

## Notes for review

- The `.pb.go` diffs are the embedded file descriptors picking up the new option — expected, and the reason those files move at all.
- `make proto` also reordered a `trie.New(...)` argument list in `indexer.pb.rgw.go`. That is generator nondeterminism over a set, unrelated to this change, so I reverted it to keep the diff honest. It will reappear for whoever regenerates next.
- `buf lint` emits a pre-existing warning that the `DEFAULT` lint category is deprecated in favour of `STANDARD`. Untouched here, but worth a follow-up in `api-spec/protobuf/buf.yaml`.

## Summary by CodeRabbit

* **Documentation**
  * Improved API documentation with clear titles, descriptions, and `v1` version information across Arkade services.
  * Documented available environments, including mainnet, Mutinynet, Signet, and local regtest where applicable.
  * Added operator-internal visibility notes for administrative, signer management, and wallet initialization APIs.
  * Added OpenAPI metadata to service definitions to keep generated documentation consistent with published API specifications.

### DIFF
--- a/api-spec/openapi/swagger/ark/v1/admin.openapi.json
+++ b/api-spec/openapi/swagger/ark/v1/admin.openapi.json
@@ -1,8 +1,9 @@
 {
   "openapi": "3.1.0",
   "info": {
-    "title": "",
-    "version": "version not set"
+    "title": "Arkade Admin API",
+    "description": "Operator-internal administration API. Not intended for application developers, and not exposed on the public operator endpoints.",
+    "version": "v1"
   },
   "paths": {
     "/v1/admin/auth/revoke": {
--- a/api-spec/openapi/swagger/ark/v1/indexer.openapi.json
+++ b/api-spec/openapi/swagger/ark/v1/indexer.openapi.json
@@ -1,9 +1,28 @@
 {
   "openapi": "3.1.0",
   "info": {
-    "title": "",
-    "version": "version not set"
+    "title": "Arkade Indexer API",
+    "description": "The read API of the Arkade operator. Use it to look up virtual outputs, commitment transactions, batch trees, and to subscribe to script activity.\n\nMost applications should use an SDK rather than these endpoints directly: https://docs.arkadeos.com",
+    "version": "v1"
   },
+  "servers": [
+    {
+      "url": "https://arkade.computer",
+      "description": "Bitcoin mainnet"
+    },
+    {
+      "url": "https://mutinynet.arkade.sh",
+      "description": "Mutinynet"
+    },
+    {
+      "url": "https://signet.arkade.sh",
+      "description": "Signet"
+    },
+    {
+      "url": "http://localhost:7070",
+      "description": "Local regtest, through arkade-regtest"
+    }
+  ],
   "paths": {
     "/v1/indexer/asset/{asset_id}": {
       "get": {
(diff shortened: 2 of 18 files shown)


gringokiwi:
The `Build and Scan` failure is not from this change — it is Trivy flagging dependency CVEs, and this PR touches proto annotations and regenerated specs only.

```
Total: 9 (HIGH: 9, CRITICAL: 0)
golang.org/x/net  CVE-2026-46600  HIGH  v0.55.0 → fixed in 0.56.0
stdlib            CVE-2026-33818        v1.26.5 → fixed in 1.25.13, 1.26.6, 1.27.0-rc.3
stdlib            CVE-2026-39821
```

The scanner is configured for `CRITICAL,HIGH` and exits non-zero on a hit. Both findings are in dependencies that master carries too — the reason master's scan is green is that it last ran before these advisories were published, and Trivy's database updates daily. Any PR opened today inherits this.

Worth its own dependency bump: `golang.org/x/net` to 0.56.0 and the Go toolchain to 1.26.6. Note #1164 already bumped `golang.org/x/text` for CVE-2026-56852, so this is the same class of maintenance.

Everything else here is green, including `verify-proto`, which is the check that matters for a regenerated-spec change.

PR #1171, opened 2026-08-26
feat(expiry): expire batches on shared epoch boundaries, behind a flag

Kukks:
Makes batches expire on a shared epoch boundary instead of a per-batch relative
timelock. **`epoch_expiry_enabled` defaults to false**, so merging changes
nothing; flipping it is a separate operational decision that needs the SDKs
shipped first.

**Retargeted at `master` so CI runs.** GitHub's workflow globs (`branches: ['*']`) do
not match a slashed base, so while this was based on `chore/epoch-expiry-foundations`
it got no CI at all - and the epoch e2e tests are the only coverage the sweep path
has. The tradeoff is that this diff now contains #1170's 11 commits as well.

**Review #1170 first** - everything up to and including `fix(sweep): keep the epoch
decode error, tighten the non-final match` belongs to it, and merging it will shrink
this diff to the epoch work alone.

### Why

Today a batch's maturity is a function of user behaviour, not the calendar:
every tree node carries the same CSV sweep leaf, and the sweeper computes each
node's maturity from *its parent's confirmation*, so broadcasting one tree level
near expiry restarts the operator's clock on that whole subtree by a full
expiry period. Every batch also matures at a different instant, so sweeps cannot
be combined, and consolidating vtxos with different expiries strands time value.

With a shared boundary `E`:

- every batch in the window matures at the same instant, so one transaction
  sweeps the epoch;
- at most **two** distinct expiry dates are live at once, so vtxos minted in the
  same window are expiry-fungible and consolidating them strands nothing;
- renewing mid-epoch returns *the same date you already had*, so the
  early-renewal drain on operator capital disappears by incentive rather than by
  policy.

### How

The sweep leaf becomes `CLTV(E) + CSV(Î”)`. An untouched node has satisfied Î” long
before `E` and matures exactly at `E`. A node created by a mid-flight unroll at
`u` matures at `max(E, u+Î”)` â€” so a unilateral exiter earns Î” **per tree level**,
with one tapscript root for the whole tree and no per-level script variation.
Griefing is bounded at `E + depthÂ·Î”` and only on branches somebody actually
unrolled, versus today's `E + depthÂ·expiry`.

`E` is computed **once** at round start and pinned on the `Round` aggregate.
Re-deriving it from the clock would let a round straddling a boundary advertise
one date to clients, build its tree against another, and schedule its sweep
against a third. The unroll grace is pinned the same way, but read back from the
batch's own tree rather than stored a second time: every node's PSBT carries the
grace it was built with, it is the source the sweep execution path already uses,
and unlike a copy it cannot drift from what the script enforces.

`epoch_anchor` is frozen while the flag is on. It only sets the phase of the
boundary grid, so moving it does not disturb existing batches â€” they carry their
own date and their own leaves â€” but new batches would land on a different grid,
putting more than two expiry dates in flight and breaking the fungibility the
scheme exists to provide.

### Admission

Settles are admitted only inside `[E âˆ’ rolloverWindow, E âˆ’ settlementCutoff]`.
The lower bound is a safety bound on every vtxo input including exits: forfeit
collection is evaluated at finalization, so a vtxo expiring mid-round slips
through unforfeited. The upper bound applies to renewals only â€” exits and
boarding stay available all epoch.

The window applies only to vtxos actually on the schedule. A vtxo from a
pre-cutover batch expires on its own relative schedule, typically further out
than the rollover window, so holding it to that bound would reject renewing it â€”
on the day of the cutover, that is every vtxo in circulation, and renewal is
precisely how one migrates onto the epoch schedule. `EpochSchedule.Governs`
tests boundary alignment, which is exact rather than heuristic: every epoch vtxo
inherits its batch's date and that date is always a boundary.

### Compatibility

- `batch_expiry` (field 3) is **not** reused. A client predating this reads it as
  a relative locktime and would fail BIP68 encoding on a timestamp; leaving it
  carrying the unroll grace keeps such a client failing closed on tree
  validation instead, which is the same outcome with a clearer cause.
- `ValidateVtxoTree` and `BuildVtxoTree` now take `tree.SweepParams`. **Breaking
  for the SDKs** â€” deliberately one type rather than an overload pair, since an
  overload would let a caller validate an epoch tree with legacy parameters and
  get a green result.
- Epoch expiry is refused on a block-height deployment. An epoch date is a unix
  timestamp; a block-height scheduler would read it as a block ~1.8e9 away and
  the batch would never sweep. If one somehow reached the sweeper anyway, that
  tree is logged and skipped rather than erroring â€” an error there aborts the
  scan for *every* batch, not just the bad one.

### Cutover (not part of this PR)

Ship the SDKs, set `build_version_header_required`, then flip
`epoch_expiry_enabled`. Pre-cutover batches keep their own relative schedule and
the sweeper handles both until the last drains. arkd logs the epoch length,
anchor, next boundary and admission window at startup so the operator can see
the dates the configuration implies.

### Test plan

- `EpochSchedule` arithmetic, including the property the design rests on: at most
  two live expiry dates across a full epoch.
- Tree build/validate in all four scheme-mismatch directions, plus a wrong-date
  case.
- Sweep maturity, including the bound the design claims â€” griefing costs
  `depthÂ·grace` past the date, not `depthÂ·expiry`.
- Grace read back from the tree, and pinned not to follow a settings change.
- Legacy-vtxo exemption across renewal, exit and swept.
- Anchor freeze, including that setting it in the same update that enables the
  flag still works and that re-submitting the same value is not an error.
- Settings round-trip through both backends (the repo fills a generated params
  struct with a struct literal, so a forgotten field writes a zero and still
  compiles â€” only a round trip catches it).
- Full suite against real Postgres and Redis; `-race` clean; linter clean; every
  commit in the series builds independently.

### Found in review

Checking my own claim that boundary alignment identifies an epoch vtxo turned up
a real bug, fixed in `03347bfb`: `createBatchSweepTask` rescheduled epoch
batches at `commitmentConfirmedAt + unrollGrace` - hours instead of weeks -
because it re-derived the sweep time from the relative psbt field instead of
reusing the maturity `findSweepableOutputs` had just computed. That field
carries the unroll grace for an epoch tree. The early sweep is rejected
`non-final`, exhausts its retries and is not re-armed until the next restart,
and the same wrong value is written onto every leaf vtxo's expiry. It fired on
every restart for every live epoch batch. Pinned by
`TestBatchSweepTaskSchedulesEpochBatchAtItsDate`.

### Not verified

**The e2e suite now runs and passes**, in its own CI job on every PR
(`make docker-run-epoch` / `make epochtest`): shared expiry dates, both halves of
the admission window, legacy/epoch coexistence, and a batch built, signed,
scheduled and swept at a real epoch boundary.

Getting there found three defects nothing else would have: the admin CLI could
not encode `epoch_expiry_enabled` at all, the redis live-store dropped every
epoch setting on the cache round-trip, and `unroll_grace >= rollover_window` was
unvalidated. The first two left the feature inert in production while the entire
unit suite stayed green.

**One design claim remains unverified onchain**: that a branch unrolled within
a grace of the boundary survives past `E` while its siblings sweep at `E` - the
unilateral-exit property the griefing bound rests on. The scheduler's half is
pinned by `TestBatchSweepTaskSchedulesUnrolledBranchAfterItsGrace`; the onchain
half is argued, not observed. Left open deliberately and noted in the test file;
worth closing before the flag is flipped anywhere griefing matters.

An earlier version of this section claimed CI would exercise these on this PR.
That was wrong twice over: no CI runs against a slashed base, and the tests were
gated behind a `-epoch` flag that `make integrationtest` never passes, so they
would have been skipped even on a green run. The gate is removed in `63d71e79`;
they now run with the rest of the integration suite.

### Review

`arkana-ai-bot` reviewed and requested changes. Seven of its nine findings are
fixed here (grace source, legacy admission, anchor freeze, event field type,
sweep-scan containment, boundary overflow, startup visibility); the remaining
two are cross-repo cutover items rather than code in this PR. The reply on the
thread also corrects one stated failure mode â€” `AdmitsSettle` never reads the
anchor, so an anchor change does not misjudge existing vtxos â€” and says plainly
where I stopped short of the suggested fix and why.

## Summary by CodeRabbit

* **New Features**
  * Added configurable epoch-based batch expiry with rollover windows, settlement cutoffs, and unroll grace periods.
  * Added CLI and environment settings for epoch expiry.
  * Exposed epoch expiry details through service information and batch-start events.
  * Added client support for epoch-based batch processing and sweeping.
  * Persisted epoch expiry settings and batch data.

* **Bug Fixes**
  * Improved retries for transactions delayed by relative or absolute timelocks.
  * Prevented invalid scheduler configurations and incorrect boolean serialization.

* **Documentation**
  * Documented epoch expiry configuration and version compatibility behavior.


Kukks:
Thanks — this was a genuinely useful review. Seven of the nine are fixed; two are cross-repo cutover items rather than code here. Pushed as `af5708d4` (rebased on the updated base).

**#1 — grace read from live settings. Confirmed, fixed.** Worse than described: the comment directly above that line claims "a settings change between finalization and this call must not move an existing batch's sweep schedule", and then the next line reads live settings for the grace. Fixed by reading it back from the tree rather than storing a second copy on the `Round` — every node's PSBT already carries the grace it was built with, it is the same source `GetSweepableBatchOutputs` uses on the execution path, and unlike a stored copy it cannot drift from what the script actually enforces. No migration needed. `TestEpochUnrollGrace` pins it.

**#2 — anchor mutability. Gap confirmed, mechanism corrected.** The anchor is now frozen while `epoch_expiry_enabled` is true; setting it while off, including in the same update that turns it on, stays allowed, and re-submitting the same value is a no-op rather than an error.

But the stated failure mode does not hold: `AdmitsSettle` never reads the anchor. It works off the vtxo's own `ExpiresAt` and the two window durations, so admission windows are anchor-independent and pre-existing vtxos are not misjudged by a change. The real damage is narrower and worth stating precisely — existing batches keep their pinned date and their own sweep leaves, but *new* batches land on a different grid, so more than two expiry dates go live at once and vtxos minted either side of the change stop being expiry-fungible. That is the property the whole scheme exists to provide, so it is still worth freezing.

I did not add the repository existence check you suggested. The residual hole is deliberate and narrow: disable the flag, change the anchor, re-enable, while epoch vtxos are still live. That is three explicit operator actions against a logged setting, not a silent footgun, and the alternative is a new repo method across two backends plus sqlc codegen for a guard on a rare admin call. Happy to add it if you think the hole matters.

**#3 — legacy renewals blocked. Confirmed, fixed.** This was the most user-visible one. `EpochSchedule.Governs` now tests whether a vtxo is on the boundary grid, and the admission window applies only to those. Exact discriminator, not a heuristic: every epoch vtxo inherits its batch's date and that date is always a boundary, while a legacy vtxo expires at `parentConfirmedAt+VtxoTreeExpiry`. A legacy vtxo landing on a boundary by coincidence is possible at second resolution and harmless — it is then held to the stricter policy. `TestCheckEpochAdmissionExemptsLegacyVtxos` covers renewal, exit and swept.

**#4 — rust-sdk.** Correct, and it is the intended fail-closed behaviour. Not fixable here; it is a cutover ordering item — SDKs ship first, then `build_version_header_required`, then the flag. Called out in the PR body.

**#5 — `uint32` vs wire `int64`.** Fixed; the event field is now `int64`. Nothing truncated today (`AbsoluteLocktime` is itself a `uint32`) but the asymmetry was pointless.

**#6 — block-height scheduler aborting the scan.** Good catch, and the consequence is worse than the diagnosis: returning an error from the `Apply` callback aborts `findSweepableOutputs` for the whole run, so one misconfigured tree stops *every* batch from being swept. Now logged at error level and skipped.

**#7 — boundary overflow.** Added. `BoundaryAfter` returns the zero time rather than wrapping, which `ExpiryFor` already rejects through its range check — the point being that a wrapped boundary lands in the past, the one failure here that would not look wrong.

**#8 — e2e note.** Added to the test file, including why a failure there should be read as design-level first: a unit-mixing bug survived every unit test precisely because they all work in seconds.

**#9 — anchor in the past.** Added a startup line naming the epoch length, the anchor, the boundary batches created now will land on, and the admission window. The default anchor is a fixed past date by design — it is only a phase offset — but an operator had no way to see the dates it implies.

Gate: full suite against real Postgres and Redis, matching baseline; `golangci-lint` clean on an LF checkout; every commit in the series builds independently.

The e2e tests are still unrun — CodeRabbit reports "reviews are disabled for this base branch" here and no CI runs against a slashed base, so #8 remains the honest gap until this is retargeted or the base merges.

Kukks:
Found a real bug in this PR while checking my own claim from the last comment — that boundary alignment is an exact test for "minted under epoch expiry". Fixed in `03347bfb`.

**`createBatchSweepTask` rescheduled epoch batches at the unroll grace, not the epoch date.**

`findSweepableOutputs` correctly computes `epochMaturity(E, parentConfirmedAt, Δ)`. The reschedule branch then threw that value away and derived its own via `getVtxoTreeExpiry`, which reads the relative PSBT field — the field that, for an epoch tree, carries **Δ (the unroll grace)** rather than a tree expiry. So it scheduled at `commitmentConfirmedAt + Δ`.

Measured, not argued: with a 21-day epoch date and the default 7168s grace, the sweep was scheduled `1787781824` — about two hours after the commitment confirmed instead of three weeks. That is `TestBatchSweepTaskSchedulesEpochBatchAtItsDate`, which fails on the old code with exactly that number and passes on the new.

Two consequences, both bad:

1. **The sweep fires early, then stops.** The sweep tx carries `nLockTime = E` from the leaf, so an early broadcast is rejected `non-final`. It exhausts its bounded retries and nothing re-arms it until the next restart. Batch outputs left unswept — the failure this whole scheme exists to prevent.
2. **Every leaf vtxo's `ExpiresAt` is overwritten** with the same wrong value, because the reschedule also drives `updateVtxoExpirationTime`. Clients see vtxos expiring in hours, and since that value is not on the boundary grid, `Governs` returns false and the epoch admission window silently stops applying to them.

**This was not a corner case.** `sweeper.start` runs this task immediately for every sweepable batch, so it fired on every operator restart for every live epoch batch.

The fix is to reuse the maturity `findSweepableOutputs` already derived, and delete `getVtxoTreeExpiry` (its only caller). That value is the correct one — it comes from the subtree's own sweep leaf and the confirmation of the node directly above it.

It also corrects the **legacy** path for a partially unrolled tree, which measured from the batch root's parent rather than the frontier's. Untouched trees are unaffected — there the frontier is the root and the two values are identical — which is why this is on this PR rather than #1170, where the "no observable behaviour change" claim holds.

I would treat this as the strongest argument yet for running the e2e suite before the flag is flipped: it is a scheduling bug that every unit test missed and that only appears once a batch is scheduled, restarted, and swept for real.

Gate: full suite against real Postgres and Redis, matching baseline; linter clean.

Kukks:
Second bug from the same audit, and this one is worse — `dac05880`. **Epoch mode could not have produced a valid batch at all.**

`startFinalization` builds the tree via `BuildCommitmentTx(..., sweepParams)`, so the batch output commits to the epoch sweep root `CLTV(E)+CSV(Δ)`. Forty lines later it built the coordinator's root separately:

```go
root, _, err := tree.BuildLegacySweepTapTreeRoot(forfeitPubkey, vtxoTreeExpiry)
...
coordinator, err := tree.NewTreeCoordinatorSession(root.CloneBytes(), batchOutputAmount, vtxoTree)
```

Unconditionally legacy, from `vtxoTreeExpiry` — which for an epoch batch is the **unroll grace**, not a tree expiry.

That root is not incidental. `prevOutFetcherFactory` does:

```go
aggregateKey, err := AggregateKeys(cosignerPubkeys, batchOutSweepClosure)
pkscript, err := script.P2TRScript(aggregateKey.FinalKey)
```

So the root is the taproot tweak on the tree's aggregate cosigner key, and that key *is* the batch output. Seeding the coordinator with the legacy root makes it derive a pkScript the commitment tx does not contain, and every sighash in the tree is then computed over the wrong prevout. The signatures would not verify against the batch the operator actually built.

Fixed by using `sweepParams.Root(forfeitPubkey)` — which is precisely what the single dispatching constructor exists for. This call site was simply missed when the rest moved onto it. Legacy batches are byte-identical, since for non-epoch params `Root` delegates to `BuildLegacySweepTapTreeRoot` with the same value. Also changed the bare `return` on error into a round failure, so the round no longer hangs with no reason recorded.

`TestSweepRootDrivesBatchOutputScript` pins the consequence rather than the mechanism: same cosigners, same relative value, legacy vs epoch root — different batch output scripts.

---

Worth being blunt about what this says: two critical bugs in two hours, both in epoch mode only, both invisible to the entire unit suite, and both found by pulling on one loose thread rather than by any test. The unit tests verify the pieces; nothing exercises a batch being built, signed, restarted and swept end to end under epoch expiry. **The e2e suite is not a nice-to-have on this PR — it is the only thing that would have caught either of these**, and it still has not run because CI does not build a PR with a slashed base.

I would not flip `epoch_expiry_enabled` anywhere until this has had a real e2e run.

Kukks:
Since two bugs of the same shape turned up back to back — "epoch params exist, but this call site still uses the legacy value" — I swept for the rest of the class rather than assuming two was all of it. Result: no further instances.

**Sweep-root construction.** Every site now goes through the dispatching `SweepParams.Root`: `service.go:3258` (fixed in `dac05880`), `builder.go:551` (`BuildCommitmentTx`), `builder.go:1227` (`extractSweepLeaf`), `validation.go:78` (`ValidateVtxoTree`), and `default_handler.go:176` client-side. The only remaining direct calls to `BuildLegacySweepTapTreeRoot` / `BuildEpochSweepTapTreeRoot` are inside `Root` itself.

**Batch maturity.** `scheduleSweepBatchOutput` takes `case round.EpochExpiry > 0` first, so the block-height and legacy arms are unreachable for an epoch batch; `findSweepableOutputs` overrides the legacy line with `epochMaturity` under `IsEpoch()`; the reschedule path now reuses that value (`03347bfb`); `Round.ExpiryTimestamp` prefers `EpochExpiry`; `StartFinalization` receives the epoch date alongside the relative one.

**Deliberately relative, and correct.** `scheduleCheckpointSweep` reads the checkpoint's own CSV closure — checkpoints are not epoch trees. `epochUnrollGrace` reads only the relative field, because for an epoch tree that field *is* the grace. The `BatchStarted` event carries the grace in `batch_expiry` by design, which is the compatibility story in the description.

**Client side.** `sweepParams()` builds `{Expiry: batchExpiry, BatchExpiry: batchExpiryDate}` from the event, so a client validates against the same scheme the operator built.

That is the whole class as far as I can find it by reading. It is not a substitute for the e2e run — both bugs I found were live-path bugs that no unit test touched, and a reading-based audit has the same blind spot.

Kukks:
One more, and it undercuts something I claimed earlier — `63d71e79`.

The epoch e2e tests were gated behind `skipUnlessEpoch`, which skips unless `-epoch` is passed. `make integrationtest` runs:

```
go test -v -count 1 -timeout 3600s github.com/arkade-os/arkd/internal/test/e2e
```

No arguments. So **CI would have skipped every epoch e2e test while reporting the suite green** — they looked like coverage without being coverage. My earlier note that "CI's nigiri stack will exercise them on this PR" was wrong on two counts: no CI runs against a slashed base, and even with CI they would not have run.

Gate removed; they now run with the rest of the integration suite. The cost is real — `TestEpochBatchSweepsAtTheBoundary` waits for an actual epoch boundary, so expect several minutes added to the job. The parameters are already as small as settings validation permits (`0 < cutoff < rollover < length`), and the boundary cannot be mined past because epoch expiry is timestamp-based by design. If the team would rather not pay that on every integration run, the alternative is a separate workflow job for it — say the word and I will split it out rather than leave it skipped.

Current state of the two branches:

- **#1170** — `unit tests`, `check-go-sync`, `postgres/redis` green. `Build and Scan` fails on `CVE-2026-14456` in the Alpine base image, which predates and is independent of this PR.
- **#1171** — no CI, by base-branch glob. Still the honest gap.

Kukks:
Retargeted at `master`, and the first CI run immediately earned its keep — `b5435153`.

**Third bug, and this one CI found rather than me.** All five epoch e2e tests failed in `enableEpochExpiry` with:

```
proto: (line 1:60): invalid value for bool field epochExpiryEnabled: 0
```

`epochExpiryEnabledFlagName` was listed among the **int-valued** flags in `updateSettingsAction`, so a `cli.BoolFlag` was read with `ctx.Int` and posted as `"epochExpiryEnabled":0`. protojson rejects that for a bool field and rejects the entire request — so `arkd settings update --epoch-expiry-enabled` failed outright, and **epoch expiry could not be turned on from the CLI at all**. That is the only way an operator turns it on, so the feature was unreachable in practice regardless of everything else in this PR.

Fixed, and the flag-to-JSON mapping is split out as `buildSettingsUpdate` so it can be tested without a running arkd. These are the first tests in `cmd/arkd`; they pin the encoded type of every epoch setting, that disabling sends `false` rather than omitting the field, and that an unset flag stays absent so updates remain partial. Putting the old line back makes them fail with `expected: bool(true), actual: int(0)` — the exact value the server refused.

**Run status at `63d71e79`:** `unit tests`, `check-go-sync` and `verify-proto` all green — `verify-proto` had never run against this branch before, and confirms the `.proto` edits and generated Go are in sync. `postgres/redis` failed only on the five epoch tests; the rest of the e2e suite passed. `Build and Scan` is the unrelated Alpine CVE.

Tally so far: three bugs, all epoch-only, all invisible to the unit suite. Two found by reading, one by the first e2e run. The e2e job is now the thing I would gate the cutover on.

Kukks:
Fourth bug, and it changes how the e2e has to run — `daf2b29f`.

**The block-height refusal was bypassable at runtime.** `Settings.Validate` rejects epoch expiry when the *settings* locktime type is block-based. But the sweep scheduler is chosen **once at startup** from `Config.VtxoTreeExpiry.Type` (`config.go:926`) and never moves. So on a deployment that booted block-based, an operator could switch `vtxo_tree_expiry` to seconds, enable epoch expiry, pass validation — and still be left with a scheduler that reads an epoch date as a block height ~1.8 billion ahead. Batches that silently never sweep: precisely the outcome the validation exists to prevent, reachable by two ordinary admin calls.

`UpdateSettings` now checks `sweeperTimeUnit`, the unit the sweeper is actually running on. Tested in all three directions, including that unrelated updates still work on a block-height deployment — which is every deployment that exists today.

I found this while trying to make the e2e work, and it settles that question too. Since the scheduler cannot be moved at runtime, the epoch tests need a deployment that **booted** seconds-based. The default e2e stack is block-based on purpose, so the rest of the suite can mine past its timelocks rather than wait real seconds. So they get their own environment and job:

- `docker-compose.regtest.yml` — the four hardcoded exit delays become overridable alongside `ARKD_VTXO_TREE_EXPIRY`. They have to move together; config validation requires one type.
- `make docker-run-epoch` — boots arkd with seconds-based delays (multiples of 512, satisfying the same-type and ordering rules).
- `make epochtest` — runs only the epoch tests.
- `integration.yaml` — a third matrix entry, `epoch`, with the test target parameterised.
- On the default stack the epoch tests now **skip with the reason** rather than fail, via `skipUnlessSecondsBased`. Not a convenience skip: epoch expiry is genuinely unavailable there, and the admin API says so.

**This adds a CI job to every PR in the repo** — roughly 20 minutes, most of it waiting for a real epoch boundary. That is a team cost I am imposing from inside this PR, so say the word if you would rather it ran only on merges to master (like `sqlite/badger` does) and I will move it.

Also fixed two CodeRabbit findings in the same commit:

- **`service.go:4017`** — the `switch` evaluates `round.EpochExpiry > 0` before the `BlockHeight` case, so an epoch round under a block-height scheduler produced a unix timestamp that `scheduleBatchSweep` treats as a block height: never due, never swept, no log line. Same guard as `findSweepableOutputs` now. Good catch, and the same defect class as two of the bugs above — I had checked this switch during my own audit and read it the wrong way round.
- **`docs/settings.md`** — I documented the anchor freeze as "once batches exist" while the code refuses any change whenever the flag is on. Documented what the code does, plus the new scheduler precondition.

Still not verified: no epoch assertion has been reached yet. This puts the environment in place; the first run that gets past setup is still ahead.

Kukks:
The epoch job ran, and the headline is that **`TestEpochBatchSweepsAtTheBoundary` passed** — 1038s, a real batch built, signed, scheduled, and swept at a real epoch boundary on a real chain. That single test exercises the hybrid `CLTV(E)+CSV(Δ)` leaf, the epoch sweep root, the boundary schedule, MTP lag and the non-final retry, and it is the thing both scheduling bugs above would have broken. `TestEpochExitIsAllowedOutsideRolloverWindow` passed too.

Also on that run: **`postgres/redis` is green**, which confirms `TestDeprecatedSignerKey` was a flake, and the epoch tests skip cleanly there. `Build and Scan` has gone green on its own — the Alpine advisory was fixed upstream, as expected for something this PR never touched.

The other three epoch tests failed, all from one cause, and it is a **fifth real bug** — `4cf191c9`.

**Nothing validated that `unroll_grace` must be shorter than `rollover_window`.**

A node matures at `max(E, appearedAt + grace)`. Every batch is minted at least a rollover window before its date, so a grace under that window leaves an untouched node maturing exactly at `E`. Let the grace exceed the window and the `max` flips for untouched nodes too: every batch matures a grace period after *its own* commitment confirmed. Consequences, all silent:

- batches stop sharing a date, so vtxos are no longer expiry-fungible — the entire point of the scheme;
- those dates land **off the boundary grid**, so `Governs` returns false and the settlement admission window stops applying to them altogether.

All three failures are that, in three costumes: two vtxos minted together differed by 2 seconds; a post-cutover vtxo carried a date 651s off its boundary; and a renewal that should have been refused outside the rollover window was **admitted**, because its off-grid expiry made `Governs` return false.

That last one is worth dwelling on. It is exactly the interaction I named a few comments ago as the claim I would most expect to be wrong — that boundary alignment is a sound test for "epoch-managed". It is sound, but only while the grace stays under the window. Now enforced, so the premise holds by construction rather than by luck.

The violation was in my own test parameters — a 512s grace against a 300s window. They are now the smallest set that satisfies the constraint (BIP68 encodes seconds in 512s steps, so 512 is the floor for a grace, which puts the window at 1024 and the epoch at 2048), and the anchor is placed deliberately instead of rounded down so batches minted moments apart provably share a boundary. Production defaults were always fine — a 2 hour grace against a 7 day window — and a test now pins that.

Five bugs: three found by reading, two by these tests. Every one epoch-only, every one invisible to the unit suite.

Kukks:
Two more from the epoch job — `e4f0e4bc`. Both are mine, and one is a regression I introduced two commits ago.

**The tests were measuring legacy batches.** A round reads its settings once, when it starts. `enableEpochExpiry` flipped the flag and minted immediately, so the batch was almost always one already in flight — built against the pre-update settings, handing back a plain relative expiry. Every assertion then failed in a way that looked like an epoch bug: vtxos whose dates differed by exactly the gap between two rounds, an admission window that never applied, a post-cutover vtxo `now+1026` instead of on its boundary. All of it was `EndingTimestamp + vtxo_tree_expiry`.

`requirePairedBatch` already waits a session duration after a settings change for precisely this reason, with a comment explaining it; `enableEpochExpiry` now does the same and reads the flag back afterwards, so a settings path that silently drops it fails there rather than as a puzzling assertion three tests later.

This also means **`TestEpochBatchSweepsAtTheBoundary` was passing for the wrong reason** on the last two runs. It was sweeping a legacy batch. I would not have caught that from the green tick — only from the arithmetic on the failures next to it.

**And I regressed restart latency.** The reschedule fix inlined the subtree scheduling, which I did not think about at the time. `scheduleBatchSweep` also rewrites the leaf vtxo expiries — a repository round trip per subtree — and `sweeper.start` waits for that whole scan before arkd reports ready, so startup cost now grew with the number of live batches. `TestDeprecatedSignerKey` restarts arkd repeatedly, and its settle timed out at four minutes in **two of three runs on this branch while passing on #1170**, which does not carry the change. I had written that off as the known integration flakiness on the first occurrence; the second is what made me look properly. Dispatched on goroutines again, as it was before; the corrected expiry value is untouched.

Moving those writes back off the test goroutine made the regression test race, which `-race` caught — the harness, not the sweeper. Fixed with a mutex.

Current state: `unit tests`, `check-go-sync`, `verify-proto`, `Build and Scan` green. `postgres/redis` green apart from that flake, with the epoch tests skipping cleanly as intended. The epoch job is the one still to prove itself, and I would not read a green tick on it as meaningful until the shared-date assertions actually pass — those are what distinguish an epoch batch from a legacy one.

Kukks:
Sixth bug, and it is the one that was making every epoch e2e assertion fail — `c7cfc3ff`.

**The redis live-store dropped every epoch setting.**

`internal/infrastructure/live-store/redis/settings.go` serializes through a hand-enumerated `settingsDTO`, and none of `epoch_expiry_enabled`, `epoch_anchor`, `epoch_length`, `rollover_window`, `settlement_cutoff` or `unroll_grace` were in it. So on any redis-backed deployment:

- the admin API writes the **repository** and reports success;
- the settings-update handler refreshes the **cache** through that DTO, silently zeroing the epoch fields;
- `startRound` reads the **cache**, sees `epoch_expiry_enabled` false, and builds a plain relative batch.

Epoch expiry could be switched on, confirmed on, and still never activate. The inmemory store copies the struct whole and is unaffected, which is exactly why no local run caught it.

What made this hard to see is that **the admin read and the round read are different stores**. The read-back verification I added last round queried the admin API, got `true`, and passed — while every minted vtxo carried `EndingTimestamp + vtxo_tree_expiry`. The tell was two vtxos minted six seconds apart differing by exactly the gap between their rounds; a shared epoch date cannot do that, and nothing else in the system produces that shape.

Worth noting for the reviewer: **this is the second time these fields have been dropped in this exact file.** `TestSettingsDTORoundTrip` already carries a comment saying it exists because a runtime build-version-header change went invisible to the version guard the same way. I added six settings and did not extend that test. Now extended — including that an unset anchor stays unset rather than becoming the unix epoch — and it fails without the DTO change.

I would suggest this file deserves a structural fix rather than another entry in a list, but that is a change to code this PR does not otherwise touch, so I have left it as a follow-up rather than widening the diff.

Also: `postgres/redis` went green on the last run, confirming the restart-latency fix resolved `TestDeprecatedSignerKey`.

Kukks:
**The epoch e2e suite is green.** `d57a1f14` — every check on this PR passes.

```
epoch            pass  23m27s
postgres/redis   pass  19m57s
unit tests       pass   6m42s
Build and Scan   pass
check-go-sync    pass
verify-proto     pass
CodeRabbit       pass
```

All five epoch tests, run against a real chain:

- `TestEpochVtxosShareAnExpiryDate` — the fungibility claim the whole design rests on
- `TestEpochRenewalOutsideRolloverWindowIsRefused` / `TestEpochExitIsAllowedOutsideRolloverWindow` — both halves of the admission window
- `TestLegacyAndEpochBatchesCoexist` — the cutover drain
- `TestEpochBatchSweepsAtTheBoundary` (1155s) — a batch built, signed, scheduled and **swept at a real epoch boundary**

The last commit confirms the `TestDeprecatedSignerKey` diagnosis: dropping `waitForConfirmation` was collateral from the reschedule fix, and it turned a staggered set of restart schedules into one burst across every live batch — 82 of them in one run. Restoring it took that job from failing four of six runs (including a re-run of the same commit) to green.

---

**Seven bugs, all epoch-only, none visible to the unit suite:**

| # | Bug | Found by |
|---|-----|----------|
| 1 | Sweep scheduled against live settings rather than the batch's own grace | reading |
| 2 | Legacy vtxos locked out of renewal at cutover | reading |
| 3 | `epoch_anchor` mutable while batches commit to the grid | reading |
| 4 | Coordinator signed the tree against the wrong batch output — **epoch mode could not produce a valid batch** | reading |
| 5 | Epoch batches rescheduled a grace period out instead of at their date, on every restart | reading |
| 6 | `epoch_expiry_enabled` unsettable from the CLI (bool sent as `0`) | e2e |
| 7 | Redis live-store dropped every epoch setting, so the flag never reached a round | e2e |

Plus a missing invariant — `unroll_grace < rollover_window`, without which batches silently stop sharing a date — and a restart-latency regression I introduced and then removed.

Four of these were found by reading the code, three by tests that did not exist when this PR was opened. **Numbers 4, 6 and 7 each independently made the feature non-functional.** Any one of them would have shipped a flag that looked implemented and did nothing, or worse.

I would still not flip `epoch_expiry_enabled` anywhere until the SDKs ship — `arkana-ai-bot`'s finding #4, the only one still open, and unchanged by any of this.

Note for whoever merges: this PR is based on `e2d9ed44` while master is now at `b158b09f`. Worth a rebase before merge, and #1170 should go first regardless — it is the first 11 commits of this diff.

Kukks:
Rebased onto `b158b09f` and both PRs are green. Plus three from CodeRabbit — `8f51b194`.

**Merge-gate item 2 is done.** Both branches now sit on current master. That also cleared #1170's `Build and Scan`: master's `ci(trivy): never serve the runtime stage from cache` (#1172) was the fix, which is why the same image scanned clean on this PR and dirty on that one — a cached runtime stage. Nothing to do with either diff. Rebase introduced exactly master's three files and nothing else; verified by diffing pre- against post-rebase on both branches.

**The grace-check subtests were vacuous.** Good catch, and the more useful kind — both rejection cases narrowed the rollover window below the grace but left the settlement cutoff at twelve hours, so `cutoff >= rollover` and `EpochSchedule.Validate` rejected on *that* rule before `Settings.Validate` ever reached the new grace check. They passed only because the two errors share the phrase "must be shorter than the rollover window". I confirmed it by deleting the grace check outright: both subtests stayed green.

Fixed by bringing the cutoff down with the window and asserting on the tail of the grace message, which the cutoff error does not use. Deleting the check now fails both, which I verified the same way.

Worth saying plainly: that test was written specifically to pin a constraint the e2e had just caught me violating, and it did not pin it. A test whose assertion matches a substring of a *different* error is worse than no test.

**Cleanup now fails the test.** The e2e cleanup restores a setting shared by every other test in the package; logging a failed restore would leave the rest of the suite running under epoch expiry while the test that left it there reported success.

**`.PHONY`** — both new targets declared.

The remaining inline comment on `redis/settings.go:365` has no finding text, only its analysis script. Reading that line myself: `unrollGrace, _ := arklib.ParseRelativeLocktime(...)` discards the second return exactly as the five delays above it already do, so I have left it consistent with the file rather than making one line different.

Open items are unchanged and both cross-repo: SDK support for fields 4 and 5 before the flag is flipped anywhere, and the documented anchor procedure (disable, drain, change, re-enable), which is enforced only while the flag is on.

Kukks:
`@arkana-ai-bot` — the 14:16 pass has regressed to an earlier baseline. It marks findings 1, 2, 3, 5 and 7 as "Still open — not touched", but your own 23:07 pass on 26 Aug tabulated 1, 2, 3, 5, 6, 7 and 9 as **Fixed**, and the code has not moved backwards since. I think "not touched in this push" has been collapsed into "not fixed" — true of the first, not of the second.

Verified against `50695503`, the current head:

| # | Claim | Actual |
|---|-------|--------|
| 1 | grace read from live settings at schedule time | `service.go:4017` — `grace, err := epochUnrollGrace(round.VtxoTree)`, read from the batch's own tree |
| 2 | anchor not frozen while the flag is on | `settings.go:495` — `"epoch_anchor cannot be changed while epoch_expiry_enabled is true"` |
| 3 | `Governs` discriminator not implemented | `epoch.go:115` defines it; `utils.go:804` applies it in `checkEpochAdmission` |
| 5 | `BatchExpiryDate uint32` against an `int64` wire type | `service_event.go:27` — `BatchExpiryDate int64` |
| 7 | `BoundaryAfter` overflow unguarded | `epoch.go:83` — `if n > math.MaxInt64/int64(s.Length)`, returning the zero time |

One genuine nuance on #1, since a bare grep would show a hit: `service.go:3021` does still read `settings.UnrollGrace`. That one is correct and must stay. It runs at **round start**, building `sweepParams` for the batch being created — live settings are exactly right there. The defect was reading live settings at **schedule** time for a batch that had already committed to a different value. Those are different call sites with opposite requirements, which is worth being precise about rather than treating every reference to the setting as the bug.

Your assessment of the two incremental commits is accurate and I have no quarrel with it — the vacuous-subtest analysis in particular matches what I found independently, down to the shared substring.

Still genuinely open, and I am not claiming otherwise:

- **#4** — rust-sdk and go-sdk support for fields 4 and 5. Cross-repo, gates the cutover, not this merge.
- **N1** — anchor changeable via disable → change → re-enable. Documented procedure, enforced only while the flag is on.
- **N2** — an intent with any offchain output puts all its vtxo inputs under the renewal bound. I read that as correct rather than over-constrained: if any output is offchain the operator is funding a new batch output, which is the thing the upper bound exists to discourage. Happy to be argued out of it.
- **#8** — the coexistence and mixed-sweep paths. Partly overtaken: `TestLegacyAndEpochBatchesCoexist` now passes in the `epoch` job, as do the other four. What remains unexercised is a mixed sweep with legacy and epoch batches maturing together.

A re-check against head would be useful. #1170's pass at 13:07 came back clean with all four of its findings closed, so I do not think this is a systematic problem.

Kukks:
`@arkana-ai-bot` N2 — closed as documented-and-intended, but you were right that something was wrong: **my documentation, not the code.**

`docs/settings.md` said the upper bound "applies only to renewals, since collaborative exits and boarding must stay available throughout the epoch." That promises more than the code delivers. `intentHasOffchainOutput` classifies an intent as a renewal if **any** output is offchain, so a *partial* collaborative exit — some sats onchain, the change kept offchain — is held to the rollover window and refused outside it. Only a *full* exit is available all epoch.

The behaviour stays, for two reasons:

1. The change output makes the operator fund a fresh batch output while the old one stays locked until the end of the epoch. That is exactly what the upper bound exists to discourage — the fact that the user is *also* taking some sats onchain does not change the operator's position.
2. The alternative rule you would reach for — treat any onchain output as an exit — makes the bound trivially bypassable: attach a dust onchain output to an ordinary renewal and it is gone. That is worse than the constraint.

So it is a genuine cost, not a free choice: a user wanting part of their balance onchain mid-epoch must exit in full or wait for the window. `7d6f7faa` documents that plainly instead of promising the opposite, and adds the reasoning to the existing discriminator test so the next reader does not "fix" it into a bypass.

That leaves, on my count:

- **#4** — SDK support for fields 4 and 5. Cross-repo, gates the cutover.
- **N1** — anchor changeable via disable → change → re-enable. Documented procedure; enforced only while the flag is on. I decided against a repository existence check for it: the residual path is three deliberate operator actions against a logged setting, and the alternative is a new repo method across two backends plus sqlc codegen for a guard on a rare admin call. Say if you disagree and I will add it.
- **#8** — remaining e2e gap: a mid-flight unroll surviving to `u+Δ` while its siblings sweep at `E`, and a mixed sweep with legacy and epoch batches maturing together. The other five epoch tests now pass in the `epoch` job.

Also worth re-checking: the findings your 14:16 pass listed as "Still open — not touched" (1, 2, 3, 5, 7) are all present at head with file:line evidence in my previous comment. Your own 26 Aug pass had them as Fixed, and nothing has moved backwards since.

Issue #1159, opened 2026-07-24
On-chain Arkade execution: arkd as an on-chain co-signing oracle


bitcoin-coder-bob:
# On-chain Arkade execution: arkd as an on-chain co-signing oracle

> **Revised after the design call.** Key changes: co-signing is limited to
> **confirmed** UTXOs, the on-chain DAG/chaining scope is dropped (the mempool already
> provides it), there is **no sweeping** on-chain, and the validation rules are inherited
> from the existing boarding-UTXO path. See the comment thread for the diff and rationale.

## Summary

Extend arkd's role from a co-signing oracle for **off-chain** spends to one that also
co-signs **on-chain** spends of Arkade-scripted UTXOs, so any contract (closure) you can
execute off-chain becomes executable on-chain. A UTXO can become a VTXO and vice versa,
giving one contract model across both environments.

Assets on-chain are explicitly a follow-up (see Non-goals).

## Motivation

1. **Removes the operator's liquidity burden.** Off-chain, a long-lived position (say 100
   BTC parked in an insurance bond) hits the batch expiry every 30 days and has to be
   renewed before it lapses, which means the operator fronts that liquidity every cycle.
   Held on-chain there is no expiry and no fronting: one Bitcoin transaction and it is
   renewed.
2. **Keeps the money programmable while it sits on-chain.** Funds can be locked to a
   covenant address that only moves if an equivalent asset was atomically minted
   off-chain. You transact the wrapped representation off-chain (cheap, only needs dust),
   while the real value stays on-chain. Redemption is two-step: burn off-chain, take the
   receipt to the on-chain covenant, withdraw to your own address.
3. **On-chain security for high-value flows.** The same Arkade contract expressiveness
   settled with L1 finality, and every Arkade contract can simply be an on-chain address
   that can receive payments directly.

## Background: how off-chain co-signing works today

To spend a VTXO off-chain, the client builds an ark tx plus checkpoint txs, and arkd
co-signs (musig between the user and the operator / arkd-signer). arkd will only ever
co-sign one spend of a given outpoint, which is the source of the off-chain
no-double-spend property. The cooperative spend path is a user+server multisig closure;
the unilateral exit path is CSV-locked. See `SubmitOffchainTx` / `FinalizeOffchainTx`
in `internal/core/application/service.go`.

## Proposed feature (in protocol terms)

Make arkd co-sign spends of UTXOs locked to Arkade vtxo-scripts where the transaction is
[...]

PR #1174, opened 2026-08-28
feat(vtxo): track onchain spends of unrolled vtxos

Kukks:
> **Taken over from @Kukks on 2026-09-02.** The design and the fix commits below are theirs. What changed since `fddaa5e3`: master merged (btcd v0.26 and four other commits, two test-file conflicts resolved by keeping both sides), a badger-only settlement race fixed, and the new tests regrouped to the repo's test layout. See "Changes in the takeover" and "Merge ordering" at the end.

## Where this sits in #1159

This is the unrolled-vtxo subset of Milestone 5 (chain watcher for unilateral spends, with reorg handling) and it resolves open question 3 for that subset by gating the sweeper and fraud readers on `IsOnchainSpent`. It is scoped to `unrolled = true` and derives "onchain spent" from fields rather than a column, since `VtxoKind` (#1161) is not on master yet. Widening the predicates to `Kind = Onchain` is the M5 follow-up and needs #1161 first.

## Problem

When a vtxo is unrolled, arkd marks it `unrolled` and then never updates it again. If the resulting onchain UTXO is later spent, arkd keeps reporting it unspent — indefinitely. That is true from inside the Ark, since nothing spent it internally, but it is wrong as a description of the coin, and it gets worse as onchain support grows.

Live example on mainnet — outpoint `4ba63c204f39841e3a7c98e458586307cf6d33bbed9a9a520c827ab043f32701:0`:

```
arkd indexer:  "isUnrolled": true, "isSpent": false, "spentBy": ""
mempool.space: spent at height 964276 by 0e85af9b…09e1f72
```

The unroll confirmed at height 960277, so it had been misreported for roughly 4000 blocks.

### Why

The chain-watch path is receive-only by construction. On every `newtransaction`, `searchNewUTXOs` reads the group's UTXO set and keeps only the rows created by that transaction. A transaction that *spends* a watched output creates no new UTXO at that script, so nothing is emitted — and the sole consumer of that channel is also the only writer of `Unrolled`. Nothing ever wrote the spend.

## Approach

Detection reads NBXplorer's group transactions endpoint, whose `MatchedInputs` name the outpoint being spent (`TransactionId`/`Index`) while the enclosing transaction is the spender. Two paths share one write path:

- **Push** rides the existing `newtransaction` event, so it adds no HTTP round trip and picks a spend up as soon as it reaches the mempool.
- **A periodic reconciler** backfills, recovers spends that arrived while arkd was down, and retracts spends whose transaction was replaced or evicted.

The reconciler's **first pass after startup is unwindowed**; the periodic passes are windowed to keep each tick cheap. That asymmetry is deliberate: a vtxo spent longer ago than the window is still in the candidate set but could never appear in a windowed query, so it would stay wrongly unspent forever — which is the backlog this change exists to clear.

The UTXO endpoint cannot detect this: it is built from `wallets_utxos`, which holds only unspent outputs, so `Confirmed.SpentOutpoints` is never populated and a confirmed spend appears nowhere. It is still the right source for the *unspent* set, and is used that way to drive retraction.

Both consumers of chain notifications (UTXOs and spends) are fed by a **single** `NotificationStream`. One event carries both halves in the same message, so a stream per consumer would duplicate every dispatch, discard half of each, and allow a spend to be observed before the UTXOs of the same event.

## State model

Reuses `spent`/`spent_by`, so **there is no schema migration on any backend**. An in-Ark spend always sets `ark_txid` (`SpendVtxos`, on an accepted offchain tx) or `settled_by` (`SettleVtxos`, at batch settlement), so their absence on a spent *and* unrolled vtxo identifies the spend as onchain. That is `Vtxo.IsOnchainSpent`, and every write is scoped by it, so an onchain spend can never clobber or undo an in-Ark spend.

`ark_txid` has existed since the initial schema on both backends (`20250702115345_init.up.sql:56`, `20250702160209_init.up.sql:56`) and no later migration adds it, so the sweeper predicate cannot drop historical rows out of the queue.

Both directions run on positive evidence, never on absence:

| Signal | Meaning |
| --- | --- |
| a matched input naming the outpoint | spent, with the spending txid |
| the outpoint present in the unspent set | unspent, and the script is still tracked |

Retracting on *absence* would be unsafe: an outpoint can be missing because its script fell out of tracking, and marking a live vtxo spent blocks its owner from registering it in an intent. A vtxo in neither set produces no action in either direction.

Mempool spends are recorded as well. Waiting for a confirmation would leave arkd building batches around an input whose owner has already broadcast a conflicting spend; the reconciler is what undoes a spend that never confirms.

## Guards

Each is a no-op for every row written before this change.

| Site | Change |
| --- | --- |
| sweeper candidate query | requires `ark_txid`; it resolves `spent_by` as a checkpoint tx, which an onchain spend has no counterpart for |
| fraud reaction | skips onchain spends, which would otherwise log `fraud detected` and then fail broadcasting a checkpoint that never existed |
| intent registration | no longer unwatches an unrolled vtxo's own script, which would have blinded tracking from that moment on |

Unrolled vtxos that are confirmed and still comfortably within the unilateral exit threshold continue to register in intents exactly as before (`TestUnrolledVtxoRejoinBatch` passes).

Two behaviour changes worth calling out explicitly:

- An unrolled, onchain-spent vtxo used as an offchain-tx input now returns `VTXO_ALREADY_SPENT` instead of `VTXO_ALREADY_UNROLLED`, because `Spent` is checked first. Both reject; the new code is the more accurate one. Pinned by test.
- `isSpent` becomes provisional: the indexer has no confirmed/unconfirmed distinction, so a mempool spend renders like a confirmed one and can flip back. This follows from reusing the existing fields and matches how mempool spends read in block explorers.

## Compatibility

Proto changes are additive only. `spends` is a new field on `NotificationStreamResponse`, so an older arkd-wallet simply never populates it. The two new RPCs return `Unimplemented` against an older wallet; arkd logs that once and degrades to not tracking onchain spends rather than erroring every tick. arkd and arkd-wallet ship as separate containers, so the skew is real.

No ark-side API change is needed — `isSpent`/`spentBy` already exist on the indexer vtxo message and simply start being populated.

Both new RPCs sit on the arkd-wallet service, which arkd reaches over `ARKD_WALLET_ADDR` and which, like every other wallet RPC, serves plaintext gRPC with no client auth of its own. They are not part of arkd's public API. Whether spend history for watched scripts is reachable from outside is a property of where the wallet listens in a deployment, and this PR does not change it.

## Config

`ARKD_ONCHAIN_SPEND_RECONCILE_INTERVAL`, in seconds, default `300`, documented in the README. `LoadConfig` rejects a non-positive or overflowing value, since `time.NewTicker` panics on one and would take the process down from inside a goroutine.

## Notes on the fix commits

Beyond the feature commit, this PR carries fixes for problems the reviews, the integration suite, and a live NBXplorer run found:

- **`c9bebce`** — a real regression introduced here. Adding the `errNotFound` sentinel changed a 404 error *message* that `GetTransaction` was parsing for the substring `"404"`. That broke the `ErrTransactionNotFound` translation the gRPC handler turns into "not confirmed", and the sweeper reads a failure there as "cannot schedule this sweep", so it stopped scheduling batch sweeps. Caught by `postgres/redis`, not by the unit suite.
- **`d3730d9`** — `inRetryableTx` batched every outpoint into one badger transaction. `ErrTxnTooBig` is not retried and `Discard` drops the whole buffer, so an oversized batch would lose the entire set; the unwindowed first pass is exactly the large-backlog case that reaches it. Now chunked. Also gives the shared stream's listeners a lifecycle, so an abandoned consumer cannot stall the dispatcher for everyone else.
- **`94b7635`** — closes a gap this PR would otherwise have shipped with. `restoreWatchingVtxos` rebuilt the watch list from sweepable rounds only and returned early when there were none, so an unrolled vtxo whose batch was no longer sweepable was dropped from tracking on the first restart. That is not a delayed spend but an unobservable one: no push arrives for an unwatched script, and NBXplorer records the matched inputs the reconciler reads only for sources it was tracking when it indexed the spending transaction. The vtxo would have stayed reported unspent forever, which is the original bug reached through a plain restart.
- **`4a5c5c7`** — found by running this against a live NBXplorer (below). `parseOutpoint` expected `<txid>-<index>`, which is the rescan *request* encoding; responses serialise an outpoint as 36 bytes of hex, the hash in internal byte order plus a little-endian index. Every spent outpoint would have failed to parse and been skipped, so the unspent set was never reduced — and because a confirmed output whose spend is in the mempool stays listed in `confirmed.utxOs`, the reconciler would have retracted each mempool spend on the tick after recording it.

## Live verification

Run against a live **NBXplorer 2.6.7** on a regtest node: tracked one address, funded it (`19631d10…a4bc:0`), then spent that exact outpoint with a raw transaction so the spend was deterministic.

Confirmed at runtime rather than from source:

- **`MatchedInputs` names the spent outpoint** — `inputs[0].transactionId=19631d10…`, `index=0`, `inputIndex=0`, with the enclosing transaction `8b570b17…` being the spender.
- **A confirmed spend appears nowhere in `/utxos`** — both `utxOs` and `spentOutpoints` empty afterwards, which is why detection reads the transactions endpoint instead.
- **A mempool spend leaves the output in `confirmed.utxOs`** while listing it in `unconfirmed.spentOutpoints`, so the subtraction in `GetUnspentOutpoints` is mandatory rather than defensive.

The third is what exposed the `parseOutpoint` bug: the observed value was `bca470cd…631900000000`, not a dashed pair.

## Changes in the takeover

- **Merged master** (`878312d`). Conflicts only in `internal/test/e2e/e2e_test.go` and the scanner test, both sides had appended functions, both kept. The btcd v0.26 import move merged cleanly into the new code; both modules build and vet.
- **Badger settlement race fixed** (`6678e7a`). A rejoined unrolled vtxo is spent onchain by the commitment tx itself, and the detector can record that before the round is projected. Badger's `settleVtxo` skipped any vtxo already marked spent, so the settlement was lost and the vtxo stayed recorded as an onchain spend with no `settled_by`. The SQL backends have no such guard, so this was badger-only. The settlement now overrides an onchain-spend mark. Covered on sqlite, postgres and badger, and the badger case fails with the guard reverted.
- **Tests regrouped** (`5a31edf`). Regroup only, no logic change: one `Test` per file with `t.Run` subtests, the restore-unrolled cases folded into the existing restore test, and every helper, fake and fixture moved below the tests that use them.

## Merge ordering

**Merge after #1093 and #1161.**

- **#1093** rewrites the same NBXplorer `newtransaction` loop to build UTXOs from the event payload instead of a per-event fetch. This PR still does that fetch and adds a second per-event call, `GetTxSpends`. The event payload also carries the matched inputs, so once #1093 is in, the spend half of this PR should read them from the event too and drop the per-event call. Expect a conflict in `nbxplorer/service.go` and its test when rebasing onto #1093.
- **#1161** adds `VtxoKind`. Nothing here depends on it in code, but the M5 follow-up that widens this tracking from `unrolled = true` to `Kind = Onchain` does, and landing #1161 first keeps that follow-up a pure extension of this PR.
- No code dependency on #1160 or #1162, and no conflicts with either.

## Test plan

Baseline captured on `b158b09f` first, then compared after: **18 packages ok, identical failure set** — `TestRoundSummariesBadger` and `TestBackfillWiringSharedStore` (Windows-only badger `.vlog` `t.TempDir` cleanup, no assertion failures) and `internal/test/e2e` (needs the `nigiri` binary). Same for the arkd-wallet module, including its pre-existing `TestNewSeedRepository` artifact, checked against the base commit rather than assumed. `golangci-lint`: no findings in the changed files.

New coverage:

- A batch settlement overriding an earlier onchain spend mark, on sqlite, postgres and badger. Fails on badger without the `settleVtxo` fix.
- `IsOnchainSpent` across the in-Ark, settled, unrolled and never-unrolled cases.
- The write path: initial mark, RBF re-point, retraction, idempotency, and the no-clobber guard against a concurrent offchain spend — on **sqlite, postgres and badger**.
- `TestOnchainSpendLargeBatch`: a batch spanning several badger transaction chunks, on all three backends.
- The sweeper's candidate set: excludes onchain spends, still includes an in-Ark spend that was later unrolled.
- The reconciler: backfill, retract-only-on-presence, no retraction when the lookup fails, graceful degradation on `Unimplemented`, and `TestReconcileFirstPassIsUnwindowed`.
- `TestNotificationStreamIsShared`: one `NotificationStream` feeding both channels.
- `TestGetTransactionNotFound` / `TestGetTxSpendsNotFound` / `TestGetTxSpendsServerError` / `TestMakeRequestNotFoundMessage`: the 404 seam that caused the sweep regression. Verified to fail when either half of that bug is reintroduced.
- The wallet scanner's spend fan-out, including that a spend-only notification does not wake the UTXO listeners with an empty map.

### Not covered, and a verified limitation

The NBXplorer contract is verified against a live instance (above) and the feature is proven end to end by `TestUnrolledVtxoOnchainSpend`, which unrolls a vtxo, claims the matured output onchain, and asserts the indexer reports it spent by the claiming transaction. Both of those cover a vtxo whose script was tracked throughout.

What is **not** covered is a vtxo that was unrolled and spent while its script was untracked. `94b7635` closes the path into that state going forward, but a vtxo already in it cannot be recovered by arkd. Verified on the same live NBXplorer:

| Action on an already-spent, previously untracked outpoint | Result |
| --- | --- |
| `rescan-utxos` (what `RescanUtxos` calls) | no effect — the group sees nothing |
| `/v1/cryptos/BTC/rescan` with the **spending** txid | recovers it — the spend and its matched input appear |
| `/v1/cryptos/BTC/rescan` with only the **funding** txid | imports the outpoint as *unspent*; the spend stays invisible |

`rescan-utxos` cannot help because `ImportUTXOs` resolves each outpoint with `gettxout`, which returns nothing once the output is spent. It does work for a still-unspent outpoint, and it does reach group-tracked addresses — that part was confirmed separately.

So recovery needs the spending txid, which is exactly what arkd is trying to learn, and an endpoint arkd-wallet does not expose. For an operator with a stranded vtxo the manual procedure is: obtain the spending txid from a block explorer, then `POST /v1/cryptos/BTC/rescan` against NBXplorer directly with that txid. Automating it would need a second chain source and is deliberately out of scope here.

An earlier revision of this description claimed `RescanUtxos` was the repair lever for anything that fell out of tracking. That was wrong for the spent case, which is the only case where repair matters, and is corrected above.


Kukks:
Thanks — all six were real, and all six are fixed in `b8df3ae`. Notes on the two I'd have been most likely to get wrong on my own:

**Unwindowed first reconcile pass.** This was the most consequential one. The rolling window silently defeated the whole point of the backfill: a vtxo spent longer ago than `reconcileLookback` was still in the candidate set but could never appear in the windowed query, so it would have stayed wrongly unspent forever. The startup pass is now unwindowed and the periodic passes stay windowed, since those only have to cover what push notifications could have missed. Pinned by `TestReconcileFirstPassIsUnwindowed`, which asserts the first call passes a nil bound and records a spend far outside the window.

**Atomic badger transitions.** Correct, and it mattered more here than the surrounding code because this specific clobber destroys `ArkTxid` — the field the sweeper and the fraud reaction key off, and the discriminator the whole no-migration design rests on. The read, the state check and the write now share one transaction, and the *whole* function is retried on conflict rather than just the write, since retrying only the write would re-apply a decision made from a stale snapshot.

The other four are straightforward: `LoadConfig` now rejects a non-positive or overflowing interval before `time.NewTicker` can panic; the spend watcher uses the service context and selects on `Done`; `Close` waits on a `WaitGroup` for in-flight fan-out deliveries before closing listener channels; and a failed UTXO lookup no longer returns early past `GetTxSpends`.

Separately, and not a review finding: the `postgres/redis` integration job failed on the previous commit with batch sweeps timing out. `TestSweep/checkpoint` passed, which is the path that uses the query this PR changes, so that guard looks exonerated — I'm treating the batch-sweep failures as unexplained rather than a flake, since master's `postgres/redis` passed in the same window, and I'll chase it on this run.

Kukks:
Found and fixed the integration failure in `c9bebce`. It was a real regression from this PR, not a flake — it reproduced identically on two runs, and master's `postgres/redis` passed in the same window.

**Root cause.** Adding the `errNotFound` sentinel changed `makeRequest`'s 404 message from `"HTTP 404: ..."` to `"not found: ..."`. `GetTransaction` detected a missing transaction by testing that message for the substring `"404"`, so it silently stopped recognising 404s and returned a generic error instead of `ErrTransactionNotFound`.

That sentinel is load-bearing three layers up:

1. `GetTransaction` no longer returns `ErrTransactionNotFound`
2. `walletHandler.IsTransactionConfirmed` special-cases exactly that sentinel to answer `{Confirmed: false}` — the ordinary answer for a transaction the node has not seen yet — so it returned a gRPC error instead
3. the sweeper reads an error from `IsTransactionConfirmed` as "cannot schedule this sweep" rather than "not yet", and stopped scheduling batch sweeps

Which is exactly what arkd logged:

```
failed to create sweep task for batch <id>:
  rpc error: code = Unknown desc = failed to get transaction: not found:
```

It also explains the failure shape: every batch sweep timed out while `TestSweep/checkpoint` passed, since that path did not depend on the not-found translation.

**Fix, on both sides.** `GetTransaction` now matches the sentinel with `errors.Is` instead of parsing a message, and the 404 message carries the status code again so any other caller still matching the old substring keeps its behaviour.

New tests cover the seam that broke — a 404 must surface as `ErrTransactionNotFound`, another status must not, and the message must still contain the status code. I verified they fail when either half of the bug is reintroduced, rather than just asserting they pass.

Worth noting for the record: I changed an error *message* that another function parsed. Nothing in the type system connected the two, and the unit suite was green throughout — only the e2e sweeps caught it.

Kukks:
Thanks — C1 and N1 are fixed in `94eca1b`. Two of your points I'm answering rather than implementing; details below.

**C1 · single NotificationStream — fixed.** Both consumers now register with one shared stream that fans out internally. You were right that the duplicate dispatch was the visible cost and the ordering window the real one. Three details worth flagging since they were judgement calls rather than mechanical:

- The stream starts on first use with its **own** context, not a caller's. Binding it to whichever consumer registered first would silently stop feeding the other once that one went away — a failure mode with no error and no log.
- Listener channels are buffered. With a shared reader, an unbuffered send would let one slow consumer stall the reader and so delay the other; that coupling didn't exist when they had separate streams, and I didn't want the fix to introduce it.
- All listeners are closed when the stream dies, preserving the termination behaviour `listenToScannerNotifications`'s `range` already depended on.

`TestNotificationStreamIsShared` pins the contract: one `NotificationStream` call, both channels fed from it.

**N1 · dispatcher not tracked in `inFlight` — fixed.** Your sequence is right, and my previous fix was incomplete in exactly the way you describe: `Wait()` could return between two batches while the dispatcher was still able to `Add(1)`. The dispatcher is now registered before launch and `Done`s on exit, so `Wait()` cannot return while it might still add work.

**C8 · `GetSpends` with `from = 0` — intentional now, not a gap.** This changed meaning between your two passes. `b8df3ae` made the *first* reconcile pass deliberately unwindowed, because a vtxo spent longer ago than the rolling window could otherwise never be reached — which is the backlog this PR exists to clear. `from = 0` is how that intent crosses the wire, so a server-side floor would reintroduce the bug. The periodic passes still window. Happy to add an explicit `unbounded` flag instead of overloading zero if you'd prefer the intent to be legible at the proto level.

**N2 · `inRetryableTx` batches all outpoints atomically — acknowledged.** Agreed on both the improvement and the trade-off. I'll leave it as is: a persistent read failure blocking a batch is self-correcting on the next reconcile tick, and partial-success semantics would make "what state is this vtxo in" harder to reason about, which is the property the whole `ark_txid` discriminator depends on.

**C7 · explicit `onchain_spent` column — agreed as a follow-up.** Your framing matches the trade-off the author of this PR (me) took deliberately: the implicit encoding is what avoids a migration, at the cost of an invariant that a future write path could violate silently. Worth an explicit column later; I'd rather not add a migration in the change that also introduces the tracking.

Also for the record, since it postdates your first pass: the integration failure was a real regression from this PR, fixed in `c9bebce`. Adding the `errNotFound` sentinel changed a 404 error *message* that `GetTransaction` was parsing for the substring `"404"`, which broke the `ErrTransactionNotFound` translation and made the sweeper stop scheduling batch sweeps. `postgres/redis` is green on `c9bebce`.

Kukks:
All three fixed in `d3730d9`. Each was a consequence of one of the previous fixes, which is worth saying plainly.

**Badger `ErrTxnTooBig`.** Correct, and worse than it first looks: `inRetryableTx` doesn't retry that error and `Discard` drops every buffered write, so an oversized batch loses the *whole* set rather than part of it. It was also directly reachable — the reconciler's first pass is unwindowed by design (that was the fix for the 30-day-window gap earlier in this PR), so the large-backlog case is precisely the one that pass exists to serve. Work is now chunked at 200, with each chunk keeping its reads, guards and writes in one transaction so the atomicity the guards depend on still holds within a chunk. `TestOnchainSpendLargeBatch` covers a batch spanning several chunks on sqlite, postgres and badger.

**Listener contexts discarded.** Also correct, and specifically a coupling *I introduced* by muxing: with a stream per consumer, an abandoned consumer only stalled its own stream; with one shared dispatcher, a full abandoned buffer stalls everybody. Listeners now carry their consumer's context, deliveries select on it, and a listener is deregistered when its context ends. Deregistration deliberately does not close the channel — the dispatcher may be mid-send on a snapshot of the slice, and closing under it would panic. The stream-death teardown is the only closer, and it runs on the same goroutine as the dispatcher, so a send can never race it.

**`notifyStarted` never reset.** Correct. Fixed as part of the teardown, so a consumer registering after the stream died opens a fresh one instead of holding a channel that can never receive and will never close.

Full local suites are at baseline on both modules.

Kukks:
@arkana-ai-bot — your latest review is stamped `94eca1b` but its incremental section analyses `b8df3ae → c9bebce`, so it hadn't seen the commit that resolves C1. Status of the three open items:

**C1 · dual `NotificationStream` — resolved in `94eca1b`, before that review posted.** Both consumers now register with one shared stream that fans out internally. Your point about causal ordering across two physical streams was the part that mattered most; that class of bug is gone rather than mitigated. Three judgement calls in the fix: the stream owns its own context (binding it to whichever consumer registered first would silently starve the other), listener channels are buffered, and all listeners close on stream death to preserve `range` termination. `TestNotificationStreamIsShared` pins it.

Muxing did introduce one coupling that separate streams didn't have — a single dispatcher means an abandoned consumer with a full buffer stalls everybody. CodeRabbit caught that on the next pass; `d3730d9` gives listeners a context, selects on it during delivery, and deregisters them when it ends.

**C6 · `ark_txid` on mainnet — answered, and it's safe.** `ark_txid TEXT` is in the *initial* schema for both backends (`20250702115345_init.up.sql:56` for postgres, `20250702160209_init.up.sql:56` for sqlite), and no later migration adds it — the only other migrations touching it rebuild views. So there has never been a deployed schema version where `vtxo` existed without `ark_txid`, and `COALESCE(ark_txid,'') <> ''` cannot drop historical in-Ark spends out of the sweep queue. Thanks for pressing on it twice; I'd asserted the invariant from the write path rather than the schema history, which is the weaker of the two arguments.

**Stale comment + missing test — both fixed in `6257777d`.** You're right that after the sentinel fix `GetTransaction` is the only caller and it no longer substring-matches, so the comment's premise was gone. Reworded to state what actually makes the status worth keeping: dropping it is what broke `GetTransaction` in the first place. `TestGetTxSpendsServerError` covers the non-404 branch and documents why the two failure modes are deliberately asymmetric.

**C4 / C7 / C8** remain as you characterised them: accepted design, deferred to a follow-up, and not a production path respectively. C8 in particular is now load-bearing rather than an oversight — the unwindowed first reconcile pass is how the backlog gets cleared.

One note for whoever does the human pass: `d3730d9` fixed a `badger.ErrTxnTooBig` risk that `inRetryableTx` introduced. Because that error isn't retried and `Discard` drops the whole buffer, an oversized batch would have lost the entire set — and the unwindowed first pass is exactly the large-backlog case that reaches it. Now chunked, with `TestOnchainSpendLargeBatch` covering a multi-chunk batch on all three backends.

Kukks:
`94b7635` closes the tracking gap I flagged in the PR description as the main thing this change could not guarantee. It is worth a look because it turns a "known limitation" into a fixed one, and because it was more serious than I first characterised it.

`restoreWatchingVtxos` rebuilt the watch list from sweepable rounds only, and returned early when there were none. An unrolled vtxo whose batch is no longer sweepable was therefore dropped from tracking on the first restart after it unrolled.

I had described that as "a spend could be missed". It is stronger than that, in a way that matters:

- no push notification arrives for an unwatched script, and
- the reconciler cannot recover it either, because NBXplorer records the `MatchedInputs` the reconciler reads **only for sources it was tracking when it indexed the spending transaction**

So a spend occurring while the script is untracked is not delayed, it is *unobservable* — and the vtxo stays reported as unspent forever. That is the exact bug this PR exists to fix, reachable through a plain restart.

Both directions are now restored: still-unspent unrolled vtxos so a spend is seen, and already onchain-spent ones so a spend later reorged out can be observed as unspent again and retracted. Keys are deduplicated against the round-derived set, and a lookup failure soft-fails in line with the other restore lookups rather than blocking startup.

`TestRestoreWatchingUnrolledVtxos` covers the no-sweepable-rounds case, the dedup, and the soft-fail; I verified it fails when the early return is reintroduced. `RescanUtxos` is still the repair lever for anything that fell out of tracking before this lands.

This does not change the remaining caveat: none of it has run against a live NBXplorer, so whether an already-untracked historical vtxo can be recovered without an explicit rescan is still unverified on a real deployment.

Kukks:
I ran this against a **live NBXplorer 2.6.7** on a regtest node, which was the one thing the PR description listed as unverified. It confirmed two of the three assumptions the design rests on — and caught a real bug in the third. Fixed in `4a5c5c7`.

### Method

Created a throwaway tracking group, tracked one address, funded it (`19631d10…a4bc:0`), then spent that exact outpoint with a raw transaction so the spend was deterministic. Observed the responses at each stage.

### Confirmed

**`MatchedInputs` names the spent outpoint.** With the spend confirmed, `/groups/{id}/transactions` returned:

```
spendingTxid=8b570b17bd7ed73b conf=1
   input -> transactionId=19631d10d01417d7 index=0 inputIndex=0
```

`transactionId`/`index` are the **spent** outpoint; the enclosing transaction is the spender; `inputIndex` is its position in the spending tx. Matches what the code assumes.

**A confirmed spend appears nowhere in `/utxos`.** After confirmation both `confirmed.utxOs` and `confirmed.spentOutpoints` were empty — which is why detection reads the transactions endpoint rather than the UTXO one.

**A mempool spend leaves the output in `confirmed.utxOs`.** With the spend unconfirmed:

```
confirmed   utxos=[('19631d10d014', 0)]  spentOutpoints=[]
unconfirmed utxos=[]                     spentOutpoints=['bca470cd…631900000000']
```

So the subtraction in `GetUnspentOutpoints` is mandatory, exactly as the PR describes.

### The bug it caught

Look at that `spentOutpoints` value. It is **not** `<txid>-<index>` — it is 36 bytes of hex: the hash in internal (reversed) byte order plus a little-endian index. Reversing `19631d10…a4bc` gives `bca470cd…6319`.

`parseOutpoint` split on `-`. I had generalised from `rescanUTXOs`, which *sends* the dashed form — but that is the request encoding, and responses use NBitcoin's `OutPoint` serialisation.

The failure mode was silent and would have defeated retraction completely: every entry fails to parse, gets logged and skipped, so the unspent set is never reduced. A confirmed output whose spend sits in the mempool stays listed in `confirmed.utxOs`, so it reads as unspent — and the reconciler would have retracted each mempool spend on the tick after recording it. That is precisely the flapping the subtraction was written to prevent, and no unit test would have caught it, because the fixture would have been my own wrong assumption.

`TestParseOutpoint` now pins the response encoding using those exact live bytes, and keeps the dashed request form working.

### Stack cleanup

Untracked the test address; the group itself remains because `DELETE /v1/groups/{id}` returns **405** on 2.6.7. Worth flagging separately: `nbxplorer.Close()` issues that same call, so group cleanup on wallet shutdown is silently a no-op on this version. Not touched here — it predates this PR. Chain advanced 2512 → 2514 (regtest, `AUTOMINE_INTERVAL=0`).

Kukks:
Thanks for the close-out. C8 is now dispositioned in `339f54c` — documented rather than guarded, because the guard you originally suggested would break the field's current behaviour.

`from = 0` is not a missing default. The reconciler's first pass after a restart is deliberately unwindowed: a vtxo spent longer ago than any rolling window would otherwise never be reached and would stay wrongly unspent forever, which is the gap `b8df3ae` fixed. A server-side floor or a rejection of zero would reintroduce it. It is documented on the proto field, where a future caller actually looks, together with the warning that anyone not doing a backfill should always set a bound.

Also took your suggestion to record why the shared stream does not reconnect — same behaviour as the per-consumer streams before the mux, push being the latency path rather than the correctness one, with the reconcile loop covering the gap.

On the two informational notes: I left `notificationBufferSize` alone. Adding back-pressure signalling would be speculative without a real burst to size it against, and the sizing only matters if a consumer stalls, which the context-aware dispatch already prevents from harming the other consumer. Worth revisiting with production evidence rather than now.

Two things your last pass could not have seen, since it analysed `94eca1b → d3730d9`:

- **`94b7635`** closes the tracking gap the PR body listed as a known limitation. `restoreWatchingVtxos` rebuilt the watch list from sweepable rounds only and returned early when there were none, so an unrolled vtxo whose batch was no longer sweepable was dropped from tracking on the first restart. That is not a delayed spend but an unobservable one — no push arrives for an unwatched script, and NBXplorer only records the matched inputs the reconciler reads for sources it was tracking when it indexed the spending transaction.
- **`4a5c5c7`** fixes a bug found by running this against a live NBXplorer 2.6.7. Responses encode an outpoint as 36 bytes of hex, not `<txid>-<index>`; the dashed form is the rescan *request* encoding. `parseOutpoint` split on `-`, so every spent outpoint would have been skipped, the unspent set never reduced, and each mempool spend retracted one tick after being recorded. Details and the observed values are in the comment above.

Both are worth a look on your next pass, since they touch the reconcile path you have been reviewing.

Kukks:
**Correction.** Earlier in this PR I described `RescanUtxos` as the repair lever for a vtxo that fell out of NBXplorer tracking. That is wrong for the case that matters, and the description has been updated.

Tested against the live NBXplorer 2.6.7 by funding an address, spending it, and only then tracking it — reproducing "the script was untracked when NBXplorer indexed the spend":

| Action on an already-spent, previously untracked outpoint | Result |
| --- | --- |
| `rescan-utxos` — what `RescanUtxos` calls | no effect; the group sees nothing |
| `/v1/cryptos/BTC/rescan` with the **spending** txid | recovers it: `bd55db86bc92 inputs=[('05993c9b69b0', 0)]` |
| `/v1/cryptos/BTC/rescan` with only the **funding** txid | imports the outpoint as *unspent*; the spend stays invisible |

`rescan-utxos` cannot help because `ImportUTXOs` resolves each outpoint through `rpc.GetTxOuts` — `gettxout` — which returns nothing once the output is spent, so no record is created and nothing is matched. It *does* work for a still-unspent outpoint, and it *does* reach group-tracked addresses; I verified both separately, so the mechanism is fine, my claim about its reach was not.

Two consequences worth having on the record:

**Recovery needs the spending txid**, which is precisely what arkd is trying to learn, via an endpoint arkd-wallet does not expose. So arkd cannot self-repair a vtxo in this state. The manual procedure for an operator is to get the spending txid from a block explorer and `POST /v1/cryptos/BTC/rescan` against NBXplorer directly. Automating it would need a second chain source — arkd does carry an `EsploraURL` — but that is a design decision beyond this PR.

**The funding-only rescan is a trap worth knowing about.** It makes the outpoint reappear as unspent. Nothing in this PR triggers it: `processBoardingInputs` calls `rescan-utxos`, not `/rescan`, and `rescan-utxos` cannot import a spent outpoint, so there is no path here that could resurrect a spent vtxo as unspent and cause a wrong retraction. But anyone adding a `/rescan` call later should be aware that a partial rescan can produce exactly that.

This does not change any code in the PR. It narrows the honest claim: `94b7635` prevents new vtxos from entering the untracked state, and nothing here recovers ones already in it.

Kukks:
Both points taken, fixed in `fddaa5e`.

**L1 — parity gap.** Correct, and the reasoning is the part that matters: `Unrolled` is half the onchain-spend discriminator, so a regression where the reconciler write path cleared it would have gone unnoticed on exactly the path where it is hardest to spot. Now asserted in both tests.

**L2 — deadline.** Also right, and worth stating why the two tests should differ rather than converge. In the push test the short deadline is *part of the assertion*: push is the only mechanism that can meet it, so passing inside it proves the notification path saw the spend rather than the 300s reconciler picking it up later. In the backfill test push is impossible by construction, so the startup reconcile is the only candidate and a tight bound proves nothing while adding flap risk. Moved to `serverWait` with that reasoning in a comment, so the asymmetry does not look like an oversight to the next reader.

**L3** — noted, no action; the first-run caveat is now moot since both tests have passed in CI (`TestUnrolledVtxoOnchainSpend` 25.97s, `TestUnrolledVtxoOnchainSpendBackfill` 27.86s), which I confirmed from the job log rather than from the green check.

Separately, I have cross-referenced this work on #1159, since it overlaps that epic in three places: it resolves open question 3 (the `reactToFraud` misfire on a legitimate onchain spend, which the design doc predicted and I confirmed), it is a partial Milestone 5 chain watcher including the read-model rollback that doc notes does not exist yet, and it carries live-verified NBXplorer constraints relevant to Design 7 — in particular that a spend indexed while its script is untracked is unobservable rather than merely delayed, and that `rescan-utxos` cannot recover it. I also flagged the scoping gap that matters for that epic: everything here is bounded by `unrolled = true`, so on-chain-*kind* UTXOs are not covered and Milestone 5 should widen those predicates rather than build a second watcher.

PR #1176, opened 2026-09-02
fix: harden client-side validation of counterparty data

Kukks:
Three client-side gaps on data that arrives from the counterparty. All of this is in `pkg/client-lib` / `pkg/ark-lib` validation used by clients; `TxTree.Validate()` is not called by the server's round processing, so arkd's own path is untouched.

**Sighash policy.** `verifyOffchainTx` verified a counterparty signature under the sighash type that same counterparty declared. It now checks the declared type against an allowlist, requires it to match the tx we built ourselves, and computes the verification hash from our own value rather than theirs. `finalizeTx` additionally refuses to sign a returned checkpoint declaring anything but `SIGHASH_DEFAULT` — there is no locally built checkpoint to compare against there, and we never stamp one ourselves. That guard also covers the previously unchecked pending-tx path.

**Tree finality.** `TxTree.Validate()` checked structure, version and amounts but not whether a node could actually be broadcast. It now also requires a zero locktime and a final sequence on every node. The operator-side builder already emits exactly these values.

**Validate before signing.** The vtxo tree was only validated at batch finalization — after nonces and signatures had already been submitted. The checks that need nothing but the tree and the commitment tx are now grouped and run at signing-start, before any nonce goes out. They still re-run at finalization against that event's commitment tx, so no check was removed or weakened.

## Test plan

- New unit tests per fix, including accept cases. Disabling each check individually makes its own tests fail and leaves the accept cases passing; with the `finalizeTx` guard disabled the malicious PSBT reaches the signer with no error at all.
- The finality tests walk the full graph (up to 255 vtxo-tree and 171 connector-tree nodes per vector), so recursion is covered rather than just roots.
- `go build ./...` clean across all modules. `go test ./...` unchanged against baseline: same 8 failing packages and same 11 failing tests before and after, all environmental here (Redis, Postgres, Docker e2e, Windows temp-dir locks). Both modules touched (`pkg/ark-lib`, `pkg/client-lib`) were green at baseline and remain green.

No e2e run — the regtest stack needs Docker and already fails at baseline in this environment. Worth a run against a live stack before merge.

## Summary by CodeRabbit

- **Security**
  - Strengthened transaction signing safeguards by rejecting unsupported or mismatched signature-hash settings.
  - Prevented signing or finalizing transactions that fail required validation checks.

- **Bug Fixes**
  - Added validation to ensure transaction trees are final and can be broadcast promptly after their parent confirms.
  - Transactions with invalid locktimes, sequences, funding, recipient amounts, or checkpoint data are now rejected before signing.

### DIFF
--- a/pkg/ark-lib/tree/tx_tree.go
+++ b/pkg/ark-lib/tree/tx_tree.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/btcsuite/btcd/psbt/v2"
+	"github.com/btcsuite/btcd/wire/v2"
 )
 
 // LeafOutput represents the output of a leaf transaction.
@@ -199,6 +200,7 @@ func (t *TxTree) SerializeNode() (*TxTreeNode, error) {
 // It verifies :
 // - every tx is a valid partial transaction.
 // - every tx has exactly one input.
+// - every tx is final, ie. it can be broadcasted as soon as its parent is confirmed
 // - the child txs spend the right parent's output
 // - the sum of the child txs' output amounts matches the parent tx input amount
 func (t *TxTree) Validate() error {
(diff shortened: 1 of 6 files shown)


Kukks:
`Build and Scan` red here is pre-existing and unrelated to this change — flagging so it isn't read as a regression.

The failing step is the Trivy image scan (`severity: CRITICAL,HIGH`, `exit-code: 1`), on a single finding:

```
golang.org/x/crypto  CVE-2026-56854  CRITICAL  installed v0.53.0  fixed 0.55.0
golang.org/x/crypto/ssh: Authentication bypass due to unenforced source-address restrictions
```

This PR changes no dependencies — the diff is 6 files, none of them `go.mod` or `go.sum`:

```
pkg/ark-lib/tree/tx_tree.go
pkg/ark-lib/tree/tx_tree_finality_test.go
pkg/client-lib/batch-session/handler/default_handler.go
pkg/client-lib/batch-session/handler/default_handler_test.go
pkg/client-lib/offchain-tx/sighash_test.go
pkg/client-lib/offchain-tx/utils.go
```

`Trivy Security Scan` on master already failed on 2026-09-01, before this branch existed. Clearing it needs a `golang.org/x/crypto` bump to ≥0.55.0, which seemed wrong to fold into an unrelated security fix — happy to do it separately if you'd prefer it here.

Issue #1115, opened 2026-06-15
GetVirtualTxs: enforce txid batching to avoid gRPC size limit and SQL variable limit


ghost:
## Summary

`GetVirtualTxs` has two failure modes when called with large numbers of txids:

1. **gRPC message size limit** — responses exceeding the 4 MB default max message size cause a `ResourceExhausted` error.
2. **SQLite variable limit** — queries with too many bound parameters cause a `SQL logic error: too many SQL variables` error.

## Expected behavior

The implementation should batch txid lookups client- and/or server-side to stay within both limits, and return a unified result set transparently to callers.

## Notes

- The SQL variable cap is a known SQLite constraint.
- The gRPC cap is the default `4194304` byte limit.
- Batching should be done carefully to avoid N+1 round-trips harming performance.

## Action items

- [ ] Determine safe batch size (below both limits)
- [ ] Implement batching in the client or server
- [ ] Add a test that exercises a txid count above the current limit

ghost:
## Triage Analysis

**Classification:** Bug — root cause confirmed. Two distinct failure modes with different fix points.

---

### Failure Mode 1 — SQLite variable limit (server-side)

**`internal/infrastructure/db/sqlite/round_repo.go:546`** — `GetTxsWithTxids` passes all txids to `q.SelectTxs` in a single call:

```go
rows, err = q.SelectTxs(ctx, queries.SelectTxsParams{
    Ids1: txids,  // txids × 3 = total SQL variables
    Ids2: txids,
    Ids3: txids,
})
```

Each txid generates 3 SQL bind variables. SQLite's compiled default `SQLITE_MAX_VARIABLE_NUMBER` is **999** (though it can be raised to 32766 at compile time — see the comment at `sqlite/vtxo_repo.go:527`). At the default limit, queries fail at **≥ 334 txids** (`334 × 3 = 1002 > 999`).

`vtxo_repo.go:527` already has this pattern documented: batch into chunks of 5000 (appropriate for single-variable queries). For `SelectTxs` (3 variables per txid), the safe batch size is **≤ 333** at the default limit.

**Fix:** Add batching to `round_repo.go:GetTxsWithTxids` — split txids into chunks of 333 (or make it configurable against `SQLITE_MAX_VARIABLE_NUMBER / 3`), run per-chunk queries, merge results. Postgres (`postgres/round_repo.go:433`) has no variable limit and doesn't need batching.

---

[...]

PR #1177, opened 2026-09-02
sqlite: batch GetTxsWithTxids to stay under SQLITE_MAX_VARIABLE_NUMBER

Tyagiquamar:
## Problem

`GetVirtualTxs` fails on the sqlite backend when called with a large txid list: `roundRepository.GetTxsWithTxids` passes the entire slice to a single `SelectTxs` query, which expands it into **three** `IN` clauses (`tx`, `offchain_tx`, `checkpoint_tx`). With N txids the statement binds 3N variables, so anything beyond ~10,922 txids exceeds `SQLITE_MAX_VARIABLE_NUMBER` (32766 in modernc.org/sqlite) and the request dies with:

```
SQL logic error: too many SQL variables (1)
```

The gRPC response-size concern from the issue is already mitigated by result pagination (`paginate` + client-side default paging from #1011); the unbounded request-side SQL variable count is not.

Refs #1115.

## Root cause

[`internal/infrastructure/db/sqlite/round_repo.go`]: `GetTxsWithTxids` → one `SelectTxs` call with `Ids1/Ids2/Ids3 = txids`, unbounded.

The postgres backend is unaffected (its `SelectTxs` binds a single `pq.Array` parameter) and is intentionally untouched.

## Fix

Mirror the pattern already established in this package by `getVtxoPubKeysByCommitmentTxidsBatched`:

- New `sqliteSelectTxsBatchSize = 5000` const (each query binds at most 3×5000 = 15000 params, generous headroom under 32766).
- `GetTxsWithTxids` delegates to an unexported `getTxsWithTxidsBatched` inner that chunks the txid slice, issues one query per chunk, and merges results in Go.
- Dedup keys on `(txid, data)` so the merged result preserves the single-query `UNION` semantics even when the input contains duplicated txids spanning two batches; first-seen order is preserved (no map-iteration randomness leaking into pagination).

For inputs up to 5000 txids the loop iterates once and behaviour is identical to today.

## Regression test

New `round_repo_batching_test.go`:

- `TestGetTxsWithTxidsBatched` — drives the inner helper across all three tx tables with batch sizes 1, 2, 3, len-1, len, len+1, 0; asserts exact union, no duplicates, duplicated-input-txid dedup across batch boundaries, unknown txids ignored.
- `TestGetTxsWithTxidsOverVariableLimit` — seeds 12,000 rows and calls the public method. On current master this fails with `too many SQL variables`; with this change it returns all 12,000 rows.

Tests use per-test file-backed DBs (`t.TempDir()`) rather than the shared `file::memory:` name, and close the full service handle, so they neither pollute nor get polluted by other package tests.

## Verification

- `go test ./internal/infrastructure/db/sqlite/ -count=1` — full package green, incl. both new tests
- Same over-limit test against unmodified master — fails with `too many SQL variables`
- `go vet ./internal/infrastructure/db/...` — clean; `gofmt` clean on touched files
- `go build ./...` — clean
- `internal/infrastructure/db/postgres` tests require a live postgres and were not run locally; that backend is unchanged

## Scope

Only the sqlite `GetTxsWithTxids` path. `GetRoundsWithCommitmentTxids` (single `IN` clause) and other slice-binding queries are unaffected by this issue's failure mode at realistic sizes and are left as-is; happy to follow up if you want them batched too.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction lookup reliability for large sets of transaction IDs by processing requests in manageable batches.
  - Preserved complete results while avoiding duplicate transaction data across batches.
  - Continued to ignore unknown transaction IDs without failing the overall lookup.

- **Tests**
  - Added coverage for varied batch sizes, duplicate IDs, unknown IDs, and large requests exceeding SQLite parameter limits.

### DIFF
--- a/internal/infrastructure/db/sqlite/export_test.go
+++ b/internal/infrastructure/db/sqlite/export_test.go
@@ -20,3 +20,17 @@ func GetVtxoPubKeysByCommitmentTxidsBatched(
 		ctx, commitmentTxids, withMinimumAmount, batchSize,
 	)
 }
+
+// GetTxsWithTxidsBatched exposes the unexported batching helper to tests in
+// sibling packages. The _test.go suffix keeps this out of the production
+// binary.
+func GetTxsWithTxidsBatched(
+	ctx context.Context,
+	repo domain.RoundRepository,
+	txids []string,
+	batchSize int,
+) ([]string, error) {
+	return repo.(*roundRepository).getTxsWithTxidsBatched(
+		ctx, txids, batchSize,
+	)
+}
--- a/internal/infrastructure/db/sqlite/round_repo.go
+++ b/internal/infrastructure/db/sqlite/round_repo.go
@@ -554,22 +554,77 @@ func (r *roundRepository) GetRoundVtxoTree(
 	return nodes, nil
 }
 
+// sqliteSelectTxsBatchSize bounds the number of txids bound to a single
+// SelectTxs invocation. The query expands the same slice into three IN
+// clauses (tx, offchain_tx, checkpoint_tx), so each call binds 3N
+// parameters. modernc.org/sqlite caps bound parameters at
+// SQLITE_MAX_VARIABLE_NUMBER = 32766; with batchSize = 5000 a single call
+// binds at most 15000 params, leaving generous headroom. The wrapper below
+// splits the input slice into batches of this size and merges the
+// deduplicated results in Go.
+const sqliteSelectTxsBatchSize = 5000
+
 func (r *roundRepository) GetTxsWithTxids(ctx context.Context, txids []string) ([]string, error) {
-	var rows []queries.SelectTxsRow
-	if err := withReadQuerier(ctx, r.db, func(q *queries.Queries) error {
-		var err error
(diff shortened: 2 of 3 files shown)


Tyagiquamar:
Hi @guggero, could you please review when you have a chance? Thanks!

Tyagiquamar:
@altafan @tiero Friendly ping for review when you have a moment. This PR batches GetTxsWithTxids inputs into 5,000-item chunks (max 15,000 bound params per query), staying safely below SQLite's SQLITE_MAX_VARIABLE_NUMBER limit (32,766). Includes unit tests for boundary sizes and cross-batch deduplication.

Tyagiquamar:
@bitcoin-coder-bob when you have a chance, could you take a look at this? It fixes the SQLite variable-limit failure in GetTxsWithTxids by batching large txid inputs while preserving UNION/dedup semantics, with regression coverage above the variable limit. Happy to adjust anything needed.

Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

